PR #17, opened 2014-12-27
fix bashisms in patch-rework.sh script

Ledest:


bhundven:
Could you explain more what this patch is fixing?
Also, I need a Signed-off-by: <you> line.

Thanks,

-Bryan

Ledest:
> I need a Signed-off-by: line.

Ok.

PR #22, opened 2015-01-02
fix bashisms in patch-rework.sh script

Ledest:
Signed-off-by: Led <ledest@gmail.com>

### DIFF
--- a/scripts/patch-rework.sh
+++ b/scripts/patch-rework.sh
@@ -67,10 +67,10 @@ do_check_files_at_depth() {
   exec 6<&0
   exec 7<"${flist}"
 
-  while read -u7 f; do
+  while read f; do
     f="$( echo "${f}" |sed -r -e "s:^([^/]+/){${depth}}::;" )"
     [ -f "${f}" ] || ret=1
-  done
+  done </dev/fd/7
 
   exec 7<&-
   exec <&6
@@ -106,16 +106,18 @@ $1=="+++" && mark==1 { nextfile; }
   >"diffstat.orig"
   printf " done\n"
 
-  pushd "${base}" >/dev/null 2>&1
+  cd "${base}"
 
   # Check all files exist, up to depth 3
   printf "  checking depth:"
-  for((d=0;d<4;d++)); do
+  d=0
+  while [ $d -lt 4 ]; do
     printf " ${d}"
     if do_check_files_at_depth "../diffstat.orig" ${d}; then
       printf " ok, using depth '${d}'\n"
       break
     fi
+    d=$((d + 1))
   done
   if [ ${d} -ge 4 ]; then
     printf "\n"
@@ -133,7 +135,7 @@ $1=="+++" && mark==1 { nextfile; }
   printf "  applying patch..."
   if ! patch -g0 -F1 -f -p${d} <"${p}" >"../patch.out" 2>&1; then
     printf " ERROR\n\n"
-    popd >/dev/null 2>&1
+    cd - >/dev/null
     printf "There was an error while applying:\n  -->  ${p}  <--\n"
     printf "'${base}' was restored to the state it was prior to applying this faulty patch.\n"
     printf "Here's the 'patch' command, and its output:\n"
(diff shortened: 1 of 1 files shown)


bhundven:
@Ledest As I mentioned in: #17 ... is there some error you are getting or a specific reason you have posted the changes. The changes result in effectively the same results. I don't see any problems with your change, I just don't see why it is nessasary. The 'git log' should also reflect that description.

On a secondary note, you don't need to close a pull request and open a new one. The trick is to just force push the updated change to your repository/branch and click the pull request button again. It will update the pull request you already have. :)
That way we also keep the review in tact instead of having to reference the previous pull request (as I did above).

Ledest:
> I just don't see why it is nessasary.

patch-rework.sh is the shell script (with /bin/sh shebang). But it contain bash-specific constructions ("pushd/popd", "for((...))", "read" with '-u' options). These constructions can easily be replaces to standard POSIX-shell's constructions. Or, in other words, "fix bashisms in shell script".

bhundven:
Ok, so with that in mind and considering that 70% (63.51% of all statistics are made up on the spot ;) ) of the scripts on crosstool-NG are bash (with /bin/bash shebang), would it make more sense to just change the shebang and call it good? I ask this because we do have a requirement to have a recent bash installed. Again, I still don't see anything wrong with the patch, and I have no problem keeping this one script posix, but I have to play the devils advocate.

Ledest:
> considering that 70% of the scripts on crosstool-NG are bash (with /bin/bash shebang), would it make more sense to just change the shebang and call it good?

May be - you know better. I try fix one particular shell script only.

bhundven:
On Fri, Jan 2, 2015 at 10:51 AM, Led <notifications@github.com> wrote:

> considering that 70% of the scripts on crosstool-NG are bash (with
> /bin/bash shebang), would it make more sense to just change the shebang and
> call it good?
>
> May be - you know better. I try fix one particular shell script only.
>
Not trying to sound confusing, but it's possible that these scripts are
supposed to be posix shell scripts for a reason, and somehow (I'm sure git
log can tell us) bashism's have slipped in on accident. @yann-morin-1998 Do
you know if that is the case?

If we fix bashism's in patch-rework.sh, should we also either switch the
shabang or update others in scripts/ dir?

For instanace, here is a short list of ```#!/bin/sh``` files in the scripts
directory:

```
addToolVersion.sh
get_in_frags.sh
install.sh (which is ok to be /bin/sh)
patch-renumber.sh
patch-rework.sh (this one)
showSamples.sh
```

If the shebang moves to /bin/bash, then should we rename these files to
${scriptname}.sh.in so that if the user has bash in /usr/local/bin or
/opt/bash/bin (and in their PATH env) the shabang points to the right place:

Such as from: crosstool-NG.sh.in
```
#!@@CT_bash@@
```

yann-morin-1998:
AS discussed with Bryan on IRC: bashisms may have sneaked in by accident.

When I wrote that script years ago, I also did not use it from a confiogure crosstool-NG tree, so I did not use the usual `@@CT_bash@@` like for other scripts, and just used the default shell.

But at that time, I was not as careful about POSIXly correctness as I am (trying to be) now.

So yes, the patch looks sane. Don't bother about my comment about the alternate addition, I was just bragging a bit. ;-)

Regards,
Yann E. MORIN.

bhundven:
With that... I'm happy to accept the patch, but I'd like to see the commit description has more then "fix bashism's". If you can update it in your local repository, and click the pull request button again to update this pull request, I'll accept the patch. (no need to make a new pull request ;) )

bhundven:
Again, All I want before I apply this patch is the commit description updated to just say something along the lines of what the patch does, i.e.: "This script has a '#!/bin/sh' shabang and might be running on a posix shell. Over time some bash features have slipped in and need to be removed."

git clone git@github.com:Ledest/crosstool-ng
cd crosstool-ng
git commit --amend
<update commit message>
git push --force

Then just hit the pull request button on your clone again and you don't need to close this pull request and just update this one instead.

I need to update the docs soon to reflect some of the new workflow.

bhundven:
@Ledest Thanks!

PR #28, opened 2015-01-09
Update of powerpc-e500v2-linux-gnuspe

manni83:
I have updated this tool chain to a newer GCC 4.8.4 version. I had to add some patches from mailing lists so that the tool chain is actually working.

I've tried the tool chain on a p2020 board, but it should work for any e500v2 powerpc.


manni83:
I just realized that with this config the compiler is generating HW floating point instructions, but when I select glibc 2.19 with --without-fp the resulting glibc uses software floating point, which is not what I want. Do you have any idea what I did wrong?

PR #35, opened 2015-01-26
arch/nios2:add nios2

netzimme:
add the nios2 architecture to crosstool-ng.

Signed-off-by: Daniel Zimmermann <netzimme@gmail.com>


bhundven:
I will merge this, but in a later commit, could you add a sample? Thanks!

PR #36, opened 2015-01-26
arch/sample : add sample file for the nios2 architecture

netzimme:
add sample file for the nios2 architecture.


bhundven:
Awesome, Thanks!

PR #39, opened 2015-02-01
Add patch to fix building glibc 2.20 on sparc

stefanct:
This actually comes from upstream:
https://sourceware.org/ml/libc-alpha/2014-09/msg00317.html
It is needed for plain glibc as well as linaro's version.
A symlink is added for the latter's version 2.20-2014.11.


stefanct:
@bhundven there you go: c98155a
I have decided to add the original sign-off too... hope that's ok with you.

bhundven:
@stefanct Thanks!

PR #42, opened 2015-02-03
libc/glibc: Add patch to fix CVE-2015-0235

bwalle:
This patch fixes the so-called "GHOST" buffer overflow in glibc 2.16.
See http://www.openwall.com/lists/oss-security/2015/01/27/9.

Signed-off-by: Bernhard Walle <bernhard@bwalle.de>


bhundven:
@bwalle I'm happy to accept this change, but the patch file name needs to be in the format:

```<3 digit patch series>-<name>.patch```

For example from patches/glibc/2.18/:
```100-4f2bcda-ARM-Fix-clone-code-when-built-for-Thumb.patch```

PR #43, opened 2015-02-03
libc/glibc: Add patch to fix CVE-2015-0235

bwalle:
This patch fixes the so-called "GHOST" buffer overflow in glibc 2.16.
See http://www.openwall.com/lists/oss-security/2015/01/27/9.

Signed-off-by: Bernhard Walle <bernhard@bwalle.de>


bwalle:
I hope closing the old pull request and creating a new one is the right process. I'm quite new to github.

bhundven:
It's ok. I still need to update the docs/ directory with updated workflow.

All you should need to do is ```git push --force``` to the branch you were working on to update the commit, and the pull request will figure it out and update the pull request.

PR #46, opened 2015-02-07
Add glibc 2.21

stefanct:
2.21 was released yesterday, announcement: https://sourceware.org/ml/libc-alpha/2015-02/msg00119.html

I have compile-tested this with a mips-multilib configuration, but apart from that I can not guarantee anything :)
Should/can we encode the minimum requirement for gcc 4.6 mentioned in that mail somewhere?


bhundven:
I'll be holding this off until after 1.20.1 is out.
Some things with glibc are going to change after the release.

bhundven:
Thanks!

PR #49, opened 2015-02-18
Updates for Linaro GCC and Linux Kernel versions

Christopher83:


bhundven:
@Christopher83 Thanks!

PR #53, opened 2015-03-06
fix gccgo build

andrewchambers:
This allows gccgo to build with gcc 4.9.2 , I tested with mips big endian, linux 2.6, and linux 3.1
The errors were missing binary utilities, and a false compiler warning about a maybe uninitialized variable.

### DIFF
--- a/patches/gcc/4.9.2/001_gccgo_unused.patch
+++ b/patches/gcc/4.9.2/001_gccgo_unused.patch
@@ -1,0 +1,13 @@
+diff --git a/libgo/runtime/mprof.goc b/libgo/runtime/mprof.goc
+index 7507dfc..44d5669 100644
+--- a/libgo/runtime/mprof.goc
++++ b/libgo/runtime/mprof.goc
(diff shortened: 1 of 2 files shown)


bhundven:
I'm hesitant to merge this. This is an issue that should be fixed upstream, then backported to older versions of gcc. I would request that you send a patch to gcc https://gcc.gnu.org/contribute.html .
If you need any help, you're welcome to email me off-list and I'll be happy to help out!

bhundven:
Is this issue fixed in gcc-5.1?

andrewchambers:
I can check sometime in the next few days (currently not at my computer). One of the issues was that crosstool uses a custom install script which doesn't install all the needed binutils, so I doubt that part is fixed.

bhundven:
If you still want this patch to be merged, all I need is a signed-off-by amended to the commit message.

andrewchambers:
Hmm, I will close for now, and redo the patch for gcc 5 if it is needed. gccgo has had a few major additions in the 5 series.

PR #57, opened 2015-03-11
bugfix: pass extra build CFLAGS and LDFLAGS to glibc

smoofra:
Glibc actually does create a build executable.  It's under sunrpc and it's
called cross-rpcgen.  It uses gettext, so if that's not available in a standard
place on your system (for example if you're using Mac OS X and Homebrew), then
you are all out of luck.

### DIFF
--- a/scripts/build/libc/glibc.sh
+++ b/scripts/build/libc/glibc.sh
@@ -371,6 +371,10 @@ do_libc_backend_once() {
             ;;
     esac
 
+    CT_CFLAGS_FOR_BUILD+=" ${CT_EXTRA_CFLAGS_FOR_BUILD}"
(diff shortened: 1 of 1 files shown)


bhundven:
Again, just need a 'signed-off-by' added, and I'll start reviewing the change.

smoofra:
I just updated both branches.

Thanks!

bhundven:
Sorry, but you'll have to update it again.

scripts/build/libc/glibc-eglibc.sh-common moved into glibc.sh, as eglibc is no longer supported with crosstool-ng. cd47c091ba6f7d6d9a98c85fc5729a434c99d4ea

Thanks for adding the sign-off!

smoofra:
OK, I just rebased it to master and tested it out.

bhundven:
Thanks!

PR #60, opened 2015-03-15
Do not use strlen to check that a string is empty

quarthex:
Here, the exact size of a not empty string is useless.

### DIFF
--- a/patches/binutils/2.25/300-012_check_ldrunpath_length.patch
+++ b/patches/binutils/2.25/300-012_check_ldrunpath_length.patch
@@ -15,7 +15,7 @@ index 137446f..bb8391a 100644
    rpath = command_line.rpath;
    if (rpath == NULL)
      rpath = (const char *) getenv ("LD_RUN_PATH");
-+  if ((rpath) && (strlen (rpath) == 0))
++  if ((rpath) && (*rpath == '\0'))
 +  	rpath = NULL;
  
    for (abfd = link_info.input_bfds; abfd; abfd = abfd->link.next)


bhundven:
I might be wrong, but I think you are thinking of the GNU `secure_getenv`:

```
The GNU-specific secure_getenv() function is just like getenv() except
that it returns NULL in cases where "secure  execution"  is  required.
```
from [getenv(3)](http://man7.org/linux/man-pages/man3/getenv.3.html) and the [info page here](http://man7.org/linux/man-pages/man3/getenv.3.html)

The line originally says: if `rpath` is not a null pointer and the length of `rpath` is zero length, set `rpath` to NULL. But that is only if `command_line.rpath` set `rpath` to NULL.

I think the original patch is doing it right.

quarthex:
It does. But my suggestion enhences the patch bypassing the length calculation.
A string is empty if its first byte is nul.
`strlen(x) == 0` has complexity O(n) whereas `*x == '\0'` has complexity O(1).

bpringlemeir:
It has nothing to do with 'rpath'.

  int foo(char *p) { if (p && strlen(p)) return 1; else return 0; }
  int foo(char *p) { if (p && p[0] != '\0')  return 1; else return 0; }

Both are the same.

I don't think the commit is wrong.  However, it is better pushed upstream.  Email the binutils people.  It is not a functional defect, but an optimization.

bhundven:
@bpringlemeir Ah, I see now.

I also think that we should try to push back on adding more patches to crosstool-ng. @yann-morin-1998 and I spoke about this a lot.

bhundven:
Thanks!

PR #63, opened 2015-03-24
Updated script to support mingw versions above major 2

neftedal:
Tested and verified with mingw version 2.0.7, 3.3.0 and 4.0-rc3

### DIFF
--- a/scripts/build/libc/mingw.sh
+++ b/scripts/build/libc/mingw.sh
@@ -89,6 +89,7 @@ do_libc() {
     do_set_mingw_install_prefix
     CT_DoExecLog CFG                                                        \
     "${CT_SRC_DIR}/mingw-w64-v${CT_WINAPI_VERSION}/mingw-w64-crt/configure" \
+        --with-sysroot=${CT_SYSROOT_DIR}                                    \
         --prefix=${MINGW_INSTALL_PREFIX}                                    \
         --build=${CT_BUILD}                                                 \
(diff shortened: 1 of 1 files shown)


jpf91:
I can confirm this is working fine, but I've got one suggestion:
If we don't specifiy `--with-sysroot=` when configuring mingw, mingw assumes sysroot==prefix. Because of this it includes `-I/usr/i686-w64-mingw32/include` when compiling the mingw libraries. If this directory exists on the build machine it could cause all kinds of issues.

neftedal:
Thanks for the input, Johannes. I agree that leaving out --with-prefix might indeed cause problems if the prefix dir is also located on the build machine. I've changed my pull request to include "--with-sysroot=${CT_SYSROOT_DIR}" when configuring mingw and verified that mingw v4.0-rc3 still builds fine.

bhundven:
All looks good to me, but I'd like to see a better description in the commit and a signed-off-by, which is required by gpl. http://elinux.org/Legal_Issues#Signed-off-by_lines_and_the_DCO

Simple as doing an interactive rebase, editing the two commits in your branch, push back to your repo and this pull-request should be updated.

neftedal:
Bryan, the commits are now updated with improved descriptions including signed-off-by

neftedal:
Let me know if you want me to change something else or if it has a go for merging as is.

bhundven:
Nope, looks good to me. Thanks!

PR #64, opened 2015-04-03
Only create ${CT_TARGET}-cc${ext} symlink if ${CT_TARGET}-gcc exists

jpf91:
Without this canadion cross builds create invalid symlinks: When `do_cc_core_backend` is called there is no `${CT_TARGET}-gcc` in the install directory. Therefore `ext` is empty and we create a link to `${CT_TARGET}-gcc`. The final compiler step then installs `${CT_TARGET}-gcc.exe` and creates a working `${CT_TARGET}-cc.exe` symlink but we still keep the invalid link as well.

Tested with various cross/canadian cross compilers with windows/linux hosts.


bhundven:
Just add the signed-off-by to the commit and I'll review this for merge.

Thanks!

jpf91:
@bhundven sorry, totally forgot to add the sign-off line. Updated.

bhundven:
Thanks!

PR #65, opened 2015-04-08
build compat fix for mac os: define offsetof if it's missing

smoofra:
When building on Mac OS, we don't seem to have offsetof when we need it for
gperf-generated code.   This patch solves that issue.

Signed-off-by: Lawrence D'Anna <larry@elder-gods.org>


bhundven:
Thanks!

PR #68, opened 2015-04-22
cc/gcc: add experimental 5.1.0

mostynb:
GCC 5.1 was released today :)
https://gcc.gnu.org/gcc-5/changes.html

### DIFF
--- a/config/cc/gcc.in
+++ b/config/cc/gcc.in
@@ -36,6 +36,12 @@ choice
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
+config CC_V_5_1_0
+    bool
+    prompt "5.1.0"
+    depends on EXPERIMENTAL
+    select CC_GCC_5_1
+
 config CC_V_linaro_4_9
     bool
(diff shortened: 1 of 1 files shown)


bhundven:
Well, I wish it were that easy. There are some other things that need to change here with the way we build the toolchain to make this "proper". I'm not saying no to gcc-5.1 (still confused where gcc-5.0 went...), I'm saying that just adding gcc-5.1 is not enough. Gcc 5.1 no longer needs cloog, only isl. Also, I don't feel a rush to get this out as no sane toolchain user will ever use a .0 in production. I'd like to wait for a 5.1.1.

I'm going to also mention the lack of the "Signed-off-by: <you>" in your commit. I need that to be in your commit message.

Also, I'm trying to finalize 1.20.1 (which all that is left is two bugs and my documentation over-haul). Lets wait on this version.

I'm going to close this request, with no foul by you, but with all I said above there are some dependencies before we get here. (really! I want gcc 5.1 in NOW!!! But it needs to wait.)

PR #69, opened 2015-04-23
Update case-sensitivity information for Cygwin

pwaring:
Cygwin no longer supports managed mounts, enabling case-sensitivity in Cygwin requires changing the registry.

FAQ number has also changed.


bhundven:
Thanks!

PR #73, opened 2015-04-26
Display 'Other languages' for bare metal as well

jpf91:
There are other languages which work with bare metal compilers. As an example crosstool-ng is recommended to build D/GDC bare metal compilers.

Signed-off-by: Johannes Pfau <johannespfau@gmail.com>

See http://wiki.dlang.org/GDC/Cross_Compiler/crosstool-NG


bhundven:
Thanks!

PR #76, opened 2015-04-29
debug/ltrace : avoid a gcc 5.1.0 error

netzimme:
to avoid a gcc 5.1.0 error add a patch for ltrace 0.7.3

Signed-off-by: Daniel Zimmermann <netzimme@gmail.com>


bhundven:
Thanks!

PR #81, opened 2015-05-02
complibs: Update requirements for gcc-5.1

bhundven:
As per: https://gcc.gnu.org/gcc-5/changes.html

"The Graphite framework for loop optimizations no longer requires the
CLooG library, only ISL version 0.14 (recommended) or 0.12.2. The
installation manual contains more information about requirements to
build GCC."

This change helps to avoid version badness.


bhundven:
I don't like the way this is done. I'm going to change this, but I have to add gcc-5.1 first.
Expect brokenness.

PR #82, opened 2015-05-02
config: Add gcc-5.1.0

bhundven:
This change needs another change from pull request #81, but it's kind of
a chicken/egg scenario. The 'select's in CC_GCC_5_1 need to be
refactored a bit, and would be easier to test if gcc-5.1.0 was commited.

Most of the refactoring will happen with CC_GCC_HAS_GRAPHITE.

### DIFF
--- a/config/cc/gcc.in
+++ b/config/cc/gcc.in
@@ -36,6 +36,11 @@ choice
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
+config CC_V_5_1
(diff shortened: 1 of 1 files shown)


bhundven:
I'm merging this, but until #81 is updated and merged, the build will be broken.

#81 is currently wrong, and the real change needs to happen in CC_GCC_HAS_GRAPHITE to choose the right dependencies based on gcc version.

PR #89, opened 2015-05-12
Windows target

mingwandroid:
Update to latest releases and add "DEVEL" version for fetching from git. Contains an example of using https://github.com/crosstool-ng/crosstool-ng/commit/f95b34d52fb2ff58cddbc426f8f008f83734f96f

### DIFF
--- a/config/libc/mingw.in
+++ b/config/libc/mingw.in
@@ -16,6 +16,31 @@ choice
 
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
+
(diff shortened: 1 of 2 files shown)


bhundven:
I can't test this, so :four_leaf_clover:

mingwandroid:
Thanks. These patches have existed for a good while now, I last built it a few weeks ago so some of the most recent changes in the mingw build script I haven't tested yet, but these version updates are orthogonal to those. I'll make sure everything's in a good state ASAP.

Issue #100, opened 2015-05-29
populate script generates error about missing libJIS.so


RevChas:
After building a new toolchain, using the populate script to get a working copy of the sysroot prints an error about libJIS.so not being found, even though it exists in both the source sysroot as well as the new copy of the sysroot.

```
% x86_64-pc-linux-gnu-populate -s ~/dev/x-tools/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sysroot -d ./rootimage
x86_64-pc-linux-gnu-populate: library 'libJIS.so' not found!
% find rootimage -name libJIS.so
rootimage/usr/lib/gconv/libJIS.so
% find ~/dev/x-tools/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sysroot -name libJIS.so
/home/tc/dev/x-tools/x86_64-pc-linux-gnu/x86_64-pc-linux-gnu/sysroot/usr/lib/gconv/libJIS.so
% 
```

bhundven:
This issue was fixed in #102

PR #102, opened 2015-05-31
populate: also search for gconv modules

yann-morin-1998:
The gconv modules are present in the (e)glibc toolchains, and some
applications directly link with one or more of those modules (even
though the classic way of using them is by dlopen()ing them).

So, also look in /usr/lib/gconv when searching for libraries.

Signed-off-by: "Yann E. MORIN" <yann.morin.1998@free.fr>

### DIFF
--- a/scripts/populate.in
+++ b/scripts/populate.in
@@ -184,13 +184,13 @@ do_add_lib() {
     local dir
     local mode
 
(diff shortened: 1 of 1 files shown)


bhundven:
This change closes #100

PR #110, opened 2015-06-03
gcc: Remove --enable-c99

bhundven:
This option is old. GCC 4.3.x old. It isn't supported anymore, and just
confuses me. I'm not planning to support 4.3.x, or really anything older
then 4.7. So this option is gone to the wind.


bhundven:
Added Acked-by

PR #111, opened 2015-06-03
autoconf: Update version numbering

bhundven:
This change removes the static .version file, and moves to using the git
tag as the version number, the change number if the current commit is
newer then the latest tag, and '-dirty' if there are changes to the git
repository since the last commit that are uncommitted.

This helps us in the troubleshooting process to identify if the user is:
* using a released version (i.e.: 1.21.0)
* using a clone from git (i.e.: 1.21.0-29-13e14f)
* using a clone from git with local uncommitted changes (i.e.: 1.21.0-29-13e14f-dirty)


bhundven:
Added Acked-by

PR #119, opened 2015-06-12
32-bit multilib support for gcc and glibc

trisk:
Quoting the email I'm drafting to the ct-ng list:
I've encountered a series of problems after enabling multilib support with CT_MULTILIB=y using gcc and glibc on amd64. These may also affect non-x86 64-bit targets with a 32-bit multilib target.

When multilib support is enabled, --host=${CT_TARGET} is still blindly passed to glibc's configure, although ${extra_cflags} is set. This causes multilib builds to fail when the primary target is amd64 and ${extra_cflags} is -m32. To solve this, we need to derive a target tuple for i586 or i686 (glibc 2.21 does not support a i386 target) and pass it with --host when ${extra_cflags} contains -m32. I introduced a new CT_TARGET_32 in CT_DoBuildTargetTuple, used in conjunction with -m32 by scripts/build/libc/glibc.sh (a similar entry for -m64 may be useful).

The second problem encountered is that pass 2 gcc's -m32 libgcc_s requires a stub libc.so installed with libc_startfiles, but all copies of libc.so are built identically without -m32. This is because the multilib ${extra_cflags} aren't being passed for their compilation. I fixed the parameters to be consistent with those used in CC when configuring glibc.

The third problem is that script/build/libc/glibc.sh's do_libc_backend installs a symlink to the location of the stub and startfiles to usr/lib/${multi_dir} on the assumption that "gcc expects them in {,usr/}lib/${extra_dir}", while the pass 2 gcc build on Linux/amd64 actually searches usr/lib32 and not /usr/lib/32 when -m32 is used. This is because for Linux/amd64, gcc/config/i386/t-linux64 specifies "m32=../lib32", etc. in MULTILIB_OSDIRNAMES for the available multilib options.

In ct-ng, usr/lib32 and usr/lib64 are both symlinks to usr/lib created by scripts/crosstool-NG.sh with this explanation: "we're *not* multilib on the target side". This assumption appears to be broken by enabling CT_MULTILIB. Either MULTILIB_OSDIRNAMES or the usr/lib32 symlink has to be altered to satisify libgcc_s linking. MULTILIB_OSDIRNAMES has variety of platform-specific configurations, so there are potentially bad side effects from overriding it globally globally, and extra work if patching each affected platform configuration file. Pointing the usr/lib{32,64} symlink corresponding to the multilib target to usr/lib/{32,64}, respectively, may be preferable for this case. For a 64-bit primary target I leave usr/lib64 symlinked to lib, as glibc still */lib64 install and runtime paths.

The final problem is the missing gnu/stubs-32.h and gnu/lib-names-32.h headers while building the final compiler, despite a multilib install of glibc. The initial gnu/stubs.h is an empty file created by scripts/build/libc/glibc.sh, but it is populated by the final glibc install, which also installs gnu/stubs-64.h. The multilib build of glibc installs the headers under ${CT_SYSROOT_DIR}/${multi_dir} which is subsequently deleted. The solution is to copy ${CT_SYSROOT_DIR}/${multi_dir}/usr/include/gnu/*-{multi_dir}.h to ${CT_HEADERS_DIR} before the deletion.

These changes produce a gcc/glibc toolchain capable of successfully compiling and linking programs with -m32.

-Albert

### DIFF
--- a/scripts/build/arch/x86.sh
+++ b/scripts/build/arch/x86.sh
@@ -8,6 +8,8 @@ CT_DoArchTupleValues() {
         CT_TARGET_ARCH=x86_64
     else
         arch="${CT_ARCH_ARCH}"
+	# Check if this is a multilib target variant
+        [ -z "${arch}" -a "${CT_TARGET_ARCH}" = "x86_64" ] && arch=i686
         [ -z "${arch}" ] && arch="${CT_ARCH_TUNE}"
         case "${arch}" in
             "")                           CT_TARGET_ARCH=i386;;
--- a/scripts/build/libc/glibc.sh
+++ b/scripts/build/libc/glibc.sh
@@ -172,6 +172,11 @@ do_libc_backend() {
                         fi
                     done
                 done
+                # Copy the gnu/{lib-names,stubs}-${multi_dir} headers needed by gnu/{lib-names,stubs}.h
+                CT_DoExecLog ALL cp -v "${CT_SYSROOT_DIR}/${multi_dir}/usr/include/gnu/lib-names-${multi_dir}.h" \
+                                       "${CT_HEADERS_DIR}/gnu/lib-names-${multi_dir}.h"
+                CT_DoExecLog ALL cp -v "${CT_SYSROOT_DIR}/${multi_dir}/usr/include/gnu/stubs-${multi_dir}.h" \
+                                       "${CT_HEADERS_DIR}/gnu/stubs-${multi_dir}.h"
                 # Remove the multi_dir now it is no longer useful
                 CT_DoExecLog DEBUG rm -rf "${CT_SYSROOT_DIR}/${multi_dir}"
             fi # libc_mode == final
--- a/scripts/build/libc/glibc.sh
+++ b/scripts/build/libc/glibc.sh
@@ -355,7 +355,7 @@ do_libc_backend_once() {
     CT_DoExecLog CFG                                                \
     BUILD_CC="${CT_BUILD}-gcc"                                      \
     CFLAGS="${glibc_cflags}"                                        \
-    CC="${CT_TARGET}-gcc ${CT_LIBC_EXTRA_CC_ARGS} ${extra_cc_args}" \
+    CC="${cross_cc} ${CT_LIBC_EXTRA_CC_ARGS} ${extra_cc_args}" \
     AR=${CT_TARGET}-ar                                              \
     RANLIB=${CT_TARGET}-ranlib                                      \
     "${CONFIG_SHELL}"                                               \
@@ -450,7 +450,9 @@ do_libc_backend_once() {
             # However, since we will never actually execute its code,
             # it doesn't matter what it contains.  So, treating '/dev/null'
             # as a C source file, we produce a dummy 'libc.so' in one step
-            CT_DoExecLog ALL "${cross_cc}" -nostdlib        \
+            CT_DoExecLog ALL "${cross_cc}" ${CT_LIBC_EXTRA_CC_ARGS} \
+                                           ${extra_cc_args} \
+                                           -nostdlib        \
                                            -nostartfiles    \
                                            -shared          \
                                            -x c /dev/null   \
--- a/scripts/build/libc/glibc.sh
+++ b/scripts/build/libc/glibc.sh
@@ -296,6 +296,15 @@ do_libc_backend_once() {
             ;;
     esac
 
+    glibc_target="${CT_TARGET}"
+    case "${extra_flags}" in
+        *-m32*)
+            if [ -n "${CT_TARGET_32}" ]; then
+                glibc_target="${CT_TARGET_32}"
+            fi
+            ;;
+    esac
+
     touch config.cache
     if [ "${CT_LIBC_GLIBC_FORCE_UNWIND}" = "y" ]; then
         echo "libc_cv_forced_unwind=yes" >>config.cache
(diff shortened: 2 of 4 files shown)


bhundven:
@trisk ... @mingwandroid and myself have been, I'll be it... slowly, working on multilib support here:

https://github.com/diorcety/crosstool-ng/commits/multilib

Take a look at those changes, as I was about to merge those (they need to be rebased).
I haven't looked at your changes, yet. But I just wanted to mention there was already work in progress.

With that said, I'm not saying no to your changes. Maybe you caught things we weren't checking for. Lets sync up our efforts.

trisk:
Thanks! I accidentally submitted this PR early (it was meant our private fork), but I'll look into what your branch covers!

trisk:
I'll comment on the individual commits there based on my findings.

trisk:
I haven't yet located equivalents of https://github.com/mistifyio/crosstool-ng/commit/136cb758165d296c5b04438a95a2765babb754ea and https://github.com/mistifyio/crosstool-ng/commit/4273782e6c144fb79ab4310be9bbfa843775ffee in your branch, although I'll check again.

Is "lib32" is created as an actual directory after the removal of the "lib32" and "lib64" symlinks by https://github.com/diorcety/crosstool-ng/commit/5096a74937b0b34af89e339c45b4caac5b3fe246 ? "../lib32" is present in gcc's MULTILIB_OSDIRNAMES for most platforms and"lib/32" won't be searched automatically, so creating "lib32" (whether as a symlink or real directory) appeared to be necessary.

Looking forward to a rebase so I can directly compare your branch with master. :)

mingwandroid:
Thanks for helping out with this!

(AFAIR) lib32 vs lib64 vs lib is a matter for the normal GCC sysroot libdir determination procedures that vary from backend to backend. The stage1 GCC is asked for the facts on the ground as that way there's no chance of incorrectly second-guessing it.

I'll recheck all these details in the patches though as we wrote them a good while ago. Once I've found my feet again I'll give detailed comments on your comments.

mingwandroid:
https://github.com/crosstool-ng/crosstool-ng/pull/120

mingwandroid:
> I haven't yet located equivalents of mistifyio@136cb75 

https://github.com/diorcety-ctng/crosstool-ng/commit/42d30e7e90f73225643d81e63d9715da98663afb

    CT_DoExecLog ALL "${cross_cc}" ${extra_flags}   \

This part:

    -    CC="${CT_TARGET}-gcc ${CT_LIBC_EXTRA_CC_ARGS} ${extra_cc_args}" \
    +    CC="${cross_cc} ${CT_LIBC_EXTRA_CC_ARGS} ${extra_cc_args}" \

.. isn't *very* important since earlier we have:

    cross_cc=$(CT_Which "${CT_TARGET}-gcc")

.. but yes it is cleaner, so I'm happy to re-submit with that fixed, or later you can submit a pull request for that part if you prefer.

and mistifyio@4273782 in your branch, although I'll check again.

I don't *think* we have an equivalent of this, but we didn't need one either. Can you verify that it's definitely important? Which version of glibc did you try with? My guess is that the symlink 'stuff' caused the problem that this is working around (headers getting found in the wrong place because paths exist that shouldn't.

Glibc is actually engineered to not work if headers are found in places they are not expected - crazy I know. I had to change the behaviour of include path lookup in GCC-on-Windows due to this:
https://github.com/diorcety/crosstool-ng/commit/643562f8457f10811a86987739afa9d9366b57b9

trisk:
> This part:
> 
> -    CC="${CT_TARGET}-gcc ${CT_LIBC_EXTRA_CC_ARGS} ${extra_cc_args}" \
> +    CC="${cross_cc} ${CT_LIBC_EXTRA_CC_ARGS} ${extra_cc_args}" \
> .. isn't very important since earlier we have:
> 
> cross_cc=$(CT_Which "${CT_TARGET}-gcc")
> .. but yes it is cleaner, so I'm happy to re-submit with that fixed, or later you can submit a pull request for that part if you prefer.
> 
Yeah, that was just incidental cleanup I did while I was there. I see the important part of the change is present in your branch. I see you're not passing `${glibc_cflags}` to the stub, which makes sense since it's only used locally.

> and mistifyio@4273782 in your branch, although I'll check again.
> 
> I don't think we have an equivalent of this, but we didn't need one either. Can you verify that it's definitely important? Which version of glibc did you try with? My guess is that the symlink 'stuff' caused the problem that this is working around (headers getting found in the wrong place - instead of in the right place - because paths exist that shouldn't exist).
> 
> Glibc is actually engineered to not work if headers are found in places they are not expected - crazy I know. I had to change the behaviour of include path lookup in GCC-on-Windows due to this:
> diorcety/crosstool-ng@643562f

This was glibc 2.21. There was exactly one copy of the gnu/{lib-names,stubs}-32.h headers, which was installed under ${CT_SYSROOT_DIR}/32/usr/include, since glibc's Makefile generates only the version for the target abi.

${CT_SYSROOT_DIR}/${multi_dir} was subsequently deleted so building the final compiler would fail since it picks up ${CT_HEADER_DIR}/usr/include/gnu/stubs.h which tries to include gnu/stubs-32.h. I assumed only ${CT_HEADER_DIR} was supposed to survive so I copied the *-32.h headers there, but I see you've removed the deletion of ${CT_SYSROOT_DIR}/${multi_dir} from the end of do_libc_backend, thus preserving the entire ${CT_SYSROOT_DIR}/${multi_dir}/usr/include for each multilib target. That may also resolve this.

trisk:
Can you confirm if my comment about ${CT_SYSROOT_DIR}/${multi_dir}/usr/include is correct?

mingwandroid:
You'd really have to verify this for yourself but your logic seems sound.

My motivation for these changes was that it's not the business of crosstool-ng to create symlinks up front or to remove folders that glibc created afterwards. Perhaps these things worked around old bugs that have since been fixed. It's also a lot cleaner both in terms of crosstool-ng's build scripts and the final folder structure.

We'd be quite interested to know how well the patch series worked out for you.

bhundven:
This PR has been superseded by #383. Closing.

PR #124, opened 2015-06-22
avr: add support for AVR avr-libc toolchains

### LINKED ISSUE #66: avr+avr-libc 8-bit toolchain not supported
I believe this would fall in to the enhancement category, but there isn't a flow for building a plain old, 8-bit AVR toolchain with avr-libc. Only avr32 with newlib seems to be supported.

enunes:
This patch series adds support for AVR avr-libc toolchains.
It is broken in a patch to add the AVR architecture and first 8-bit options,
followed by adding avr-libc as a C library option, then an avr target sample
and finally some other minor issues observed during testing.

Initially this was tested by adding a lot of other configuration options by
taking the Arch Linux recipes as a reference. They can be found in:
https://projects.archlinux.org/svntogit/community.git/plain/trunk/PKGBUILD?h=packages/avr-binutils
https://projects.archlinux.org/svntogit/community.git/plain/trunk/PKGBUILD?h=packages/avr-gcc
https://projects.archlinux.org/svntogit/community.git/plain/trunk/PKGBUILD?h=packages/avr-libc
Trying to make crosstool-ng equivalent to these options would require a lot of
changes to the crosstool-ng scripts and configuration options.
Because of this, I ended up just using them for initial tests and to make it
work for the first time, but I decided to keep the ammount of changes for this
initial inclusion of the AVR architecture to the minimum possible which
produces a working AVR toolchain with avr-libc.

The avr-libc manual and most distributions build the AVR 8-bit gcc toolchain
with the "avr" (non-canonical) target.
Some experimentation also led to the conclusion that other (canonical) targets
are not very well supported, so I'm forcing the "avr" target for crosstool-ng
as well. avr-libc seems to have an "avr/include" path hardcoded in its
Makefiles which causes the standard include paths to miss the avr-libc headers
when the target is different than "avr".

The manual also recommends building avr-libc after the final gcc build.
To accomplish this with crosstool-ng, a new do_libc_post_cc step is added, in
which currently only avr-libc performs its build, and is a no-op for the other
libc options.

It seems that it is also possible to build the AVR toolchain with newlib
instead of avr-libc, but this is untested yet with crosstool-ng. This may be
tested and supported after the initial AVR inclusion in crosstool-ng has been
concluded.
For the newlib case, it may be possible to build it with a canonical target.
There is at least this reference:
http://antoni.buszta.info/2013/07/18/avr-toolchain-with-gcc-4-8-1-and-c-stl-support

Another detail which took some time to figure out is the fact that AVR
toolchains require multilib support. Building it with "--disable-multilib"
made link crt* with a wrong path (crtm328p.o instead of avr/lib/avr5/crtm328p.o
or equivalent, depending on the mcu) and to generate an error during linking.
I have decided then to force multilib for the AVR architecture.

After putting it to work and having cleaned all of the experimental code, I
performed the following tests:
- Built my existing AVR/avr-libc code at https://github.com/enunes/avr with
this toolchain (it finds all of the avr-libc headers and mcu specific libraries
as well as it did with the distribution toolchain). Also built some C++ code
with avr-g++.
- Programmed an AVR binary to an ATmega328p microcontroller and observed it
working properly.

To test whether it would have affected other architectures:
- Rebuilt a few of my existing toolchains, for arm-unknown-linux-gnueabi,
arm-none-eabi (newlib) and i686-unknown-elf targets, and rebuilt code that I
used to build them with.
- Built a linux filesystem with buildroot using the arm-unknown-linux-gnueabi
toolchain (qemu_arm_vexpress_defconfig) and booted it in qemu.
- Programmed an arm binary built with arm-none-eabi to an arm cortex-m4
microcontroller and observed it working properly.

This series also closes #66.

### DIFF
--- a/config/arch/avr.in
+++ b/config/arch/avr.in
@@ -1,0 +1,8 @@
+# AVR specific config options
+
+## select ARCH_SUPPORTS_8
+## select ARCH_DEFAULT_8
+## select ARCH_REQUIRES_MULTILIB
+##
+## help The 8-bit AVR architecture, as defined by:
+## help     http://www.atmel.com/products/microcontrollers/avr
--- a/config/kernel/linux.in
+++ b/config/kernel/linux.in
@@ -1,5 +1,6 @@
 # Linux kernel options
 
+## depends on ! ARCH_avr
 ## select KERNEL_SUPPORTS_SHARED_LIBS
 ##
 ## help Build a toolchain targeting systems running Linux as a kernel.
--- a/config/libc/avr-libc.in
+++ b/config/libc/avr-libc.in
@@ -1,0 +1,51 @@
+# avr-libc options
+
+## depends on ARCH_avr
+## depends on ! LINUX && ! WINDOWS && BARE_METAL
+##
+## select LIBC_SUPPORT_THREADS_NONE
+##
+## help The AVR Libc package provides a subset of the standard C library for
+## help Atmel AVR 8-bit RISC microcontrollers. In addition, the library
+## help provides the basic startup code needed by most applications.
+
+choice
+    bool
+    prompt "avr-libc version"
+# Don't remove next line
+# CT_INSERT_VERSION_BELOW
+
+config LIBC_AVR_LIBC_V_1_8_1
+    bool
+    prompt "1.8.1"
+
+config LIBC_AVR_LIBC_V_1_8_0
+    bool
+    prompt "1.8.0"
+
+config LIBC_AVR_LIBC_CUSTOM
+    bool
+    prompt "Custom avr-libc"
+    depends on EXPERIMENTAL
+
+endchoice
+
+if LIBC_AVR_LIBC_CUSTOM
+
+config LIBC_AVR_LIBC_CUSTOM_LOCATION
+    string
+    prompt "Full path to custom avr-libc source"
+    default ""
+    help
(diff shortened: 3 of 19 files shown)


grant-h:
Compiles fine for me, but it appears that the default avr sample doesn't build `avr-g++`, which is needed for any Arduino based project. Still doing more testing.

bhundven:
@grant-h You could just ```ct-ng menuconfig``` and enable c++ support. It should be that simple. Please report if enabling c++ worked.

enunes:
Hey @grant-h , thanks for testing.
g++ is disabled in the current sample but as I mentioned in the log I did perform simple tests with avr-g++. I did not test using this toolchain as a backend for the Arduino tool, although I believe this could be a nice test.
I agree that we can have C++ enabled in the sample, I have checked and many of the samples do have it enabled, I'll submit a further patch later which enables it by default. Please let me know if you have any other feedback.

bhundven:
@grant-h Could you test again with the new sample added? If everything looks good after your testing, I'll merge.

Thanks in advance!

grant-h:
Going to test right now on a non-trivial C++ project and upload.

grant-h:
@bhundven Tested and working out-of-the-box on my end. Thanks for your contribution @enunes!

bhundven:
@grant-h Thank you for testing!
@enunes Merging! Thank you for your contribution!

PR #125, opened 2015-06-23
Add support for native and cross-native toolchains

jpf91:
This patch series adds support for native and cross-native compilers. We're using these patches to build the toolchains at http://gdcproject.org/downloads so for now they only include the features we needed:

Native compilers:
* Builds for the current build machine
* Uses locally installed glibc, linux headers. No way to specifiy other glibc versions for now
* Builds companion libs as usual. Allows static linking
* Doesn't build binutils for now

cross-native compilers:
* Needs a existing build=>target compiler (just like canadian builds)
* Uses glibc, linux headers from build=>target compiler. No way to specifiy other glibc versions for now
* Builds companion libs as usual. Allows static linking
* Optionally builds binutils (required for mingw toolchains)

Native compilers don't use sysroots and can use headers & libraries installed on the system they'll be running on. (Exception: MinGW native compilers still use sysroots).

Tested with:

build: x86_64-linux-gnu
* native
* cross-native
    * i686-w64-mingw32
    * x86_64-w64-mingw32
    * i686-linux-gnu
    * arm-linux-gnueabihf
    * arm-linux-gnueabi
* canadian
    * i686-w64-mingw32=>arm-linux-gnueabi
    * i686-w64-mingw32=>arm-linux-gnueabihf
    * x86_64-w64-mingw32=>arm-linux-gnueabi
    * x86_64-w64-mingw32=>arm-linux-gnueabihf
    * i686-linux-gnu=>arm-linux-gnueabi
    * i686-linux-gnu=>arm-linux-gnueabihf
* cross
    * i686-linux-gnu
    * arm-linux-gnueabihf
    * arm-linux-gnueabi
    * x86_64-w64-mingw32
    * i686-w64-mingw32

Remaining questions: I currently have the install binutils & disable bootstrap options in toolchain.in. Should these be moved to binutils.in / gcc?


jpf91:
@bhundven It'd be great if you could provide some early feedback. Do we want support for native/cross-native toolchains? Is this approach basically OK? Should the C library, kernel etc options be hidden in menuconfig for native/cross-native builds?

bhundven:
@jpf91 I'll test this change set this weekend. Sorry for the delay!

bhundven:
@jpf91 I thought I'd get time, and that for me has been the elusive part of my life... finding time.

There is someone in IRC that needs this support. If you're around, maybe you can pop in to #crosstool-ng on freenode?

jpf91:
> I thought I'd get time, and that for me has been the elusive part of my life... finding time.

I guess every OSS developer has experienced that problem ;-) Don't worry, this is not urgent. I was just wondering if there's any interest in this functionality at all.

> There is someone in IRC that needs this support. If you're around, maybe you can pop in to #crosstool-ng on freenode?

Sorry, I wasn't around a computer for the last few days. Was there any kind of problem with this code? AFAIK it should just work, at least we're using it actively. Maybe these configuration files could be useful as examples:

Native compiler: https://github.com/D-Programming-GDC/build-gdc-config/blob/master/configs/x86_64-linux-gnu/gcc-5/x86_64-linux-gnu/.config
Cross-native compiler: https://github.com/D-Programming-GDC/build-gdc-config/blob/master/configs/arm-linux-gnueabi/gcc-5/arm-linux-gnueabi/.config

The `CT_CC_LANG_OTHERS="d"` option should be removed for normal builds (You need a patched GCC for D support)

bhundven:
@jpf91 Could you fix any conflicts? I'd be happy to take another look at this!

jpf91:
@bhundven Rebased, but not yet tested. Testing will take some more hours.

bhundven:
@jpf91 Thanks! I'll do some tests on the branch too.

jpf91:
OK, my `x86_64-linux-gnu` native and `arm-linux-gnueabi` and `x86_64-w64-mingw32` cross-native tests seem to work.

bhundven:
@jpf91 Sorry to do this again, but I have since merged some other changes and this needs to be rebased again.

Another request from me is to also add your samples to the PR? (don't add them to .travis.yml yet, just the samples/...)

jpf91:
>Another request from me is to also add your samples to the PR? (don't add them to .travis.yml yet, just the samples/...)

How do you generate the sample configuration files? `ct-ng savedefconfig`?

bhundven:
On Oct 15, 2015 3:20 AM, "Johannes Pfau" <notifications@github.com> wrote:
>>
>> Another request from me is to also add your samples to the PR? (don't
add them to .travis.yml yet, just the samples/...)
>
> How do you generate the sample configuration files? ct-ng savedefconfig?

Yup.

That should be it.

jpf91:
OK, I've added and tested a `native` and some `cross-native` samples.

bhundven:
Rad! I know a lot of people that will love this support! (including me!)

bhundven:
Another thought about this feature...

Do you think native and cross-native should depend on static-toolchain support?

Say I am building a cross-native for RPI, but my distro does not have development libraries or has different versions of libraries the toolchain depends on. All sorts of weird un-explainable issues will incur.

jpf91:
> Do you think native and cross-native should depend on static-toolchain (CT_STATIC_TOOLCHAIN) support?

Maybe. I'd only ever ship static toolchain builds for sure. But OTOH I know people who successfully shipped dynamically linked toolchains, so I don't really want to force static linking. Canadian cross compilers should also be affected by this. How is this handled for these right now?

> Say I am building a cross-native for RPI, but my distro does not have development libraries or has different versions of libraries the toolchain depends on. All sorts of weird un-explainable issues will incur.

I theory yes, in practice usually things just work. The GCC dependencies are properly versioned and care about ABI compatibility. So in my experience this is not a big problem. However, I'd still ship statically built toolchains whenever possible ;-)

BTW: The libc is often a much bigger problem. If you link against newer libc versions, the resulting programs will usually not run on systems with older libc versions. I therefore always use old glibc versions in cross compilers (2.14.1). The cross-native toolchains will depend on the libc version provided by the cross compiler used to compile the toolchain.

Native compilers even depend on the locally installed libc version on the build machine. This might be one thing to fix in the future.

bhundven:
@mingwandroid, @diorcety and @yann-morin-1998 can I get a second opinion on some of these changes. I have some comments I will add tomorrow to some of the changes. I would like to see this support go in, but there are a few moving targets going on as well:

* 2 pass toolchain build is being worked on
* multilib support

that might throw a wrench in this set of changes. I would almost like to see 2 pass toolchain build support go in before this.

bhundven:
@jpf91 As time goes by, I'm feeling less interested in native. I do like the cross-native support. I feel that if you want a native compiler, just use your distribution's toolchain? Just curious what the motivation for native is?

jpf91:
The main motivation is native toolchains which are shipped to other machines. For example, we build all our compilers on a `x86_64-linux-gnu` machine. But we need to ship a `x86_64-linux-gnu` native toolchain as well, so we need native support*. Shipping many different toolchains is easier if all are built with the same build tool. And ct-ng makes static linking of the toolchain and companion libs and changing companion libs versions easy, automates downloads, .... At least easier than if I have to write a new build script for that ;-)


(*) I'm not sure if ctng can actually configure a cross-native or a canadian toolchain with host=build and get the expected result.

mingwandroid:
Re: static, with GCC and now binutils moving to being compiled as C++ code that's an extra level of ABI incompatibility - and a much more frequently triggered one if they use libstdc++ - to worry about. I'd also try hard to avoid shipping dynamically linked toolchains.

Re: native, I quite like having it. Just so you can use one tool for all your toolchain requirements. It would also help to test that the binaries produced by the built cross compilers don't contain different code just because they were cross compiled. I know it goes against the name and original purpose though, but if the changes aren't too intrusive then I am in favour.

bhundven:
@jpf91 I was talking with @yann-morin-1998 on irc:

> 13:35 <@y_morin> bhundven: Back. About the 'native' type: there are two options on how to interpret it:
> 13:35 <@y_morin> bhundven: 1) generate code the native _system_
> 13:36 <@y_morin> bhundven: 2) generate code for the native _cpu_
> 13:36 <@y_morin> bhundven: The second is just a standard cross-compiler.
> 13:36 <@y_morin> bhundven: The first is much more complex. It is probably not achievable easily.
> 13:37 <@bhundven> y_morin: I just commented on the pull request.
> 13:38 <@bhundven> I'm starting to feel less comfortable supporting native.
> 13:38 <@bhundven> as you said, we almost already do the second form
> 13:38 <@y_morin> bhundven: Exactly.
> 13:38 <@bhundven> and I don't want to support the first form, apt-get source gcc
> 13:46 <@bhundven> y_morin: Johannes responded to the PR
> 13:46 <@bhundven> sounds like he is talking about the second form
> 13:48 <@y_morin> bhundven: If we keep 'native' and it means the second form, I'm afraid users will want to compile programs with that toolchain and expect them to work on the build machine.
> 13:50 <@bhundven> true

So with that, we mean, we don't want to support a native toolchain for the system.
For a native toolchain for the cpu, we currently do that by making a cross-compiler, say: build:x86_64-pc-linux-gnu host:x86_64-pc-linux-gnu target: x86_64-unknown-linux-gnu
(note the vendor difference)

bhundven:
It's less of a technically feasible issue, and more of a support/logistical issue.

jpf91:
The way I've implemented `native` compilers here as actually your `type 1`. The compiler generates code for the native system as in:

* It uses the system wide installed binutils linker
* It uses system-wide installed libc
* It can use system wide installed headers and link system wide installed libraries. 

Note that all these points apply for `cross-native` compilers as well. The only difference between `cross-native` and `native` is that for `native` compilers, `BUILD==HOST==TARGET` whereas for `cross-native` compilers `BUILD=/=HOST==TARGET`. When shipping built toolchains to other machines, there's no difference between `native` and `cross-native` compilers.

An alternative would be to explicitly support `cross-native` compilers with `TARGET` manually set to the `BUILD` tuple. I'm not sure if that works right now with this pull request.

PR #128, opened 2015-07-10
config: Add new gcc stable versions

bhundven:
This commit adds gcc 4.8.5 and 4.9.3.
Release notes can be found in these urls:
https://gcc.gnu.org/gcc-4.8/
https://gcc.gnu.org/gcc-4.9/


bhundven:
This commit is pretty trivial, so I'm just pushing it through.

PR #129, opened 2015-07-14
Patch glibc 2.11's configure script to support Make 4

kkaefer:
Similar patch in http://colocsbar.blogspot.fr/2014/01/crosstool-ng-1190-et-make-40.html


kkaefer:
Note that crosstool also has an option to build a custom `make` (3.82) in the "companion tools" section.

PR #140, opened 2015-08-24
mingw-w64: Rename targets to {i686,x86_64}-w64-mingw32

mingwandroid:
.. and set CT_TARGET_VENDOR to w64. Otherwise config.gcc doesn't
pickup the right files for include_next (and probably many other
things go wrong too).

w64 has been the correct vendor for absolutely ages now on all
distributions that provide MinGW-w64 cross compilers.


bhundven:
Thanks!

Issue #142, opened 2015-09-01
Extra config for core GCC no working


jasmin-j:
When extra configuration parameters for the core GCC are configured ("Core gcc extra config"), they
are not used in the configure command.
I will add a branch to my repo on GitHub.

jasmin-j:
Please find the fix for this issue on branch
https://github.com/jasmin-j/crosstool-ng/tree/fix_142

PR #144, opened 2015-09-01
Fix for issue #142:

jasmin-j:
Configure for core GCC did not use Core gcc extra config.
Use now config variable CT_CC_GCC_CORE_EXTRA_CONFIG_ARRAY.


bhundven:
I'm surprised this wasn't caught during the review process when CC_CORE was replaced by CC_GCC_CORE for the multi_cc work. Good catch!

Issue #145, opened 2015-09-01
GCC version "Custom gcc" will force GCC_latest


jasmin-j:
When you want to use a custom GCC version, this will not always be the latest GCC version.
Forcing this selection is very limiting and makes this option more or less useless.
The right behaviour would be to open an additional selection for the custom GCC version.
I will add a branch to my repo on GitHub to implement the new feature.

PR #146, opened 2015-09-01
Fix for issue #145:

jasmin-j:
Added gcc version selection if CC_GCC_CUSTOM is selected.

### DIFF
--- a/config/cc/gcc.in
+++ b/config/cc/gcc.in
@@ -304,11 +304,46 @@ config CC_GCC_CUSTOM
     bool
     prompt "Custom gcc"
     depends on EXPERIMENTAL
-    select CC_GCC_latest
 
 endchoice
 
 if CC_GCC_CUSTOM
+
+choice
+    bool
+    prompt "Custom gcc version"
+
+config CC_GCC_V_5_2_X
+    bool
+    prompt "5.2.x"
+    select CC_GCC_5_2
+
+config CC_GCC_V_5_1_X
+    bool
+    prompt "5.1.x"
+    select CC_GCC_5_1
+
+config CC_GCC_V_4_9_X
+    bool
+    prompt "4.9.x"
+    select CC_GCC_4_9
+
+config CC_GCC_V_4_8_X
+    bool
+    prompt "4.8.x"
+    select CC_GCC_4_8
+
+config CC_GCC_V_4_7_X
+    bool
+    prompt "4.7.x"
+    select CC_GCC_4_7
+
+config CC_GCC_V_latest
+    bool
+    prompt "latest"
+    select CC_GCC_latest
+
+endchoice
 
 config CC_GCC_CUSTOM_LOCATION
     string


bhundven:
I'm going to hold on this change. I think there might be another way to handle this.
I'm in -08:00 (PST), and I'll look at this in the morning.

jasmin-j:
> I think there might be another way to handle this.
I tried to reuse the already existing GCC configuration options. CT-NG supports a lot of versions and reusing this mechanism is the best I can think of.

Which other way?
What shall I try?

yann-morin-1998:
What @bhundven and I discussed on IRC: the CC_GCC_CUSTOM option should get out of the choice.

So, you still select the version from the existing choice, and the CC_GCC_CUSTOM is now an option of its own. So, no need to duplicate the choice just to get the correct _or_later.

jasmin-j:
OK, I will update my patches (gcc, newlib) according to this suggestion. Moreover, I will do this for Binutils and GDB too, when I have time (I need it anyway).

bhundven:
Just out of curiosity, How are you using these changes?
I have planned work to add generic git/svn/bzr support for all packages. These seem like similar changes?

jasmin-j:
> Just out of curiosity, How are you using these changes?

I want to build an arm-none-eabi cross compiler from this sources
   svn://gcc.gnu.org/svn/gcc/branches/ARM/embedded-4_9-branch
Later maybe newlib from here
   git://sourceware.org/git/newlib-cygwin.git

There is a script available to build this compiler, but it is not that easy for non experienced people.
And the script is only available inside of the tarball:
  https://launchpad.net/gcc-arm-embedded/4.9/4.9-2015-q2-update

It is work to trim crosstool-ng to build with the exactly same  configuration parameters as the script uses, but I am sure I will manage it ;)
In the long run, I plan to use crosstool-ng to build a dedicated toolchain for a new embedded OS called distortos (arm-distortos-eabi):
   https://github.com/DISTORTEC/distortos
So it will require some patches to newlib and maybe gcc. The rest of the tools should be build out of the box. But it should be much easier to build this version with crosstool-ng, than with the script. And crosstool-ng will improve over time (see your idea with git/sv/bzr support). This is a great synergy!

> I have planned work to add generic git/svn/bzr support for all packages.

Perfekt!

> These seem like similar changes?

Not really. This would use the changes ;)
Even if you use git, svn or whatever, crosstool-ng needs to know the gcc, binutil, newlib version of the checked out sources.
I guess your checkout feature would be very useful as a pre step of the build (instead of getting the tarballs). And then you would still require a directory to checkout, which could be the custom location.
Moreover, you will need some patches for checked out versions, too. Look to the patches, they are really old and still not in the main line. Some of them might be never go there, but are required for a cross build.
At the end you will have two directories. One with the original upstream version and one with the patches applied. Your feature could be so clever to use a symbolic link, if there are no patches.

jasmin-j:
I implemented @yann-morin-1998 suggestion on my branch "gcc_custom_version".
Closing this pull request, because it is now obsolete.

Issue #147, opened 2015-09-01
GCC configuration "Use __cxa_atexit" is disabled with bare-metal configuration


jasmin-j:
When the target OS is selected to bare-metal, the GCC configuration option "Use __cxa_atexit" is disabled. The mentioned GCC configuration should be enabled by the used libc, if it provides such a function.
I will add a branch to my repo on GitHub to implement the new feature.

bhundven:
This change is merged. Closing. :)

PR #148, opened 2015-09-01
Fix for issue #147:

jasmin-j:
Enable definition of "Use __cxa_atexit" for bare-metal systems, if the used
libc does provide such a function. The libc configuration have to select
LIBC_PROVIDES_CXA_ATEXIT.


bhundven:
Thanks!

PR #151, opened 2015-09-01
Support binutils V 2.25.1

jasmin-j:
Adding patches for binutils version 2.25.1 as copy of 2.25.
Support binutils 2.25.1 in configuration.
Note: The patches do apply, but I didn't check the resulting tools.

### DIFF
--- a/config/binutils/binutils.in
+++ b/config/binutils/binutils.in
@@ -25,6 +25,11 @@ choice
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
+config BINUTILS_V_2_25_1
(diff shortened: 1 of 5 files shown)


bhundven:
@enunes Is this the correct version of the AVR patch?
Considering you're the author of that patch, I just want to check this is the version of the patch you want committed?

enunes:
@bhundven @jasmin-j yes the avr patch is right and working, I just performed a test with an avr toolchain from this branch with binutils 2.25.1.

bhundven:
Thanks to you both!

PR #163, opened 2015-09-03
Remove CC_GCC_CUSTOM from the version choice

jasmin-j:
CC_GCC_CUSTOM is no longer part of the gcc version choice, but an independent
configuration to enable CC_GCC_CUSTOM_LOCATION.


bhundven:
Thanks!

PR #164, opened 2015-09-03
Remove LIBC_NEWLIB_CUSTOM from the version choice

jasmin-j:
LIBC_NEWLIB_CUSTOM is no longer part of the newlib version choice, but an
independent configuration to enable LIBC_NEWLIB_CUSTOM.
All newlib versions >=2.0.0 does provide __cxa_atexit. To enable this function
in GCC, all versions >=2.0.0 does now select LIBC_PROVIDES_CXA_ATEXIT.
Note: The patch for issue #147 needs to be applied for a working
"ct-ng menuconfig" execution!
    

bhundven:
Thanks!

PR #165, opened 2015-09-04
Raspberry Pi 2 sample

Frederick888:
There has been a sample for Raspberry Pi Model B+ but Pi 2 is different. It features a Cortex-A7 (ARMv7 neon vfpv3) CPU. I have successfully built the toolchain with the configurations and compiled some programs for Pi 2.

I think Pi 2 now is a quite popular platform so could you please add this sample to crosstool-NG?


bhundven:
I totally agree to adding this to ct-ng. Why not squash these two commits into one?

Frederick888:
Yeah, it's better now.

bhundven:
Thanks!

PR #166, opened 2015-09-05
Remove BINUTILS_CUSTOM from the version choice

jasmin-j:
BINUTILS_CUSTOM is no longer part of the binutils version choice, but an
independent configuration to enable BINUTILS_CUSTOM.
    

bhundven:
Thanks!

PR #167, opened 2015-09-05
Adding CLooG/isl constraint

jasmin-j:
Adding CLooG 0.18.4.
Adding isl 0.15.
Added following constrains:
   isl 0.14 and 0.15 require  CLooG 0.18.4 or later.
   isl 0.11.1 and 0.12.2 require CLoog older than 0.18.4.


jasmin-j:
Sorry, branch was not based on master, but on my internal development version.
--> Closed

PR #168, opened 2015-09-05
Adding CLooG/isl constraint

jasmin-j:
Adding isl 0.15.
Added following constrains:
   isl 0.14 and 0.15 require  CLooG 0.18.4 or later.
   isl 0.11.1 and 0.12.2 require CLoog older than 0.18.4.

### DIFF
--- a/config/companion_libs/cloog.in
+++ b/config/companion_libs/cloog.in
@@ -9,6 +9,11 @@ if ISL
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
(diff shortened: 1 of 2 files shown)


jasmin-j:
Vania (from the crossgcc ML) and I discovered a compile problem in CLooG 0.18.1 with isl 0.14.
This CLooG version did require the older isl version  0.12.2.
I tested today different isl/CLooG combinations and this patch is the result.

bhundven:
Thanks!

PR #170, opened 2015-09-06
Newlib specific target cflags

jasmin-j:
Added additional newlib specific target cflags with new option
LIBC_NEWLIB_TARGET_CFLAGS.
    

jasmin-j:
The already defined option TARGET_CFLAGS will be used for all target libraries. It was not possible to define different target flags for libgcc and libc.

The new option LIBC_NEWLIB_TARGET_CFLAGS allows now additional target CFLAGS for newlib.
It might be worth to consider moving this out of libc/newlib.in to libc.in, so that all libc variants may use this new option.

PR #182, opened 2015-09-22
Adding missing if/else blocks in do_gcc_core_backend.

jasmin-j:
Adding missing if/else blocks in do_gcc_core_backend for CT_CC_GCC_LIBSSP,
CT_CC_GCC_HAS_LIBQUADMATH and CT_CC_GCC_LIBQUADMATH (--en/disable-libssp,
--en/disable-libquadmath, --en/disable-libquadmath-support) from function
do_gcc_backend.


jasmin-j:
Because do_gcc_core_backend is used for the final compiler in case of bare-metal, this options need to be set according to the configuration, too.

PR #183, opened 2015-09-22
Synchronize CC_GCC_USE_LTO parameter setting.

jasmin-j:
Synchronize CT_CC_GCC_USE_LTO parameter setting in do_gcc_core_backend with the
one from do_gcc_backend, by removing "--enable-lto"/"--disable-lto".

### DIFF
--- a/scripts/build/cc/100-gcc.sh
+++ b/scripts/build/cc/100-gcc.sh
@@ -307,10 +307,8 @@ do_gcc_core_backend() {
     fi
     if [ "${CT_CC_GCC_USE_LTO}" = "y" ]; then
         extra_config+=("--with-libelf=${complibs}")
-        extra_config+=("--enable-lto")
     elif [ "${CT_CC_GCC_HAS_LTO}" = "y" ]; then
         extra_config+=("--with-libelf=no")
-        extra_config+=("--disable-lto")
     fi
 
     if [ ${#host_libstdcxx_flags[@]} -ne 0 ]; then


jasmin-j:
Because do_gcc_core_backend is used for the final compiler in case of bare-metal, this options need to be set similar to the one in do_gcc_backend.

jasmin-j:
@bhundven
Additional info for you. I have a patch in the queue to add a parameter "build_step" to do_gcc_core_backend, so that this function "knows" for which step it gets executed.
The result of the patch series can be seen on my branch
   https://github.com/jasmin-j/crosstool-ng/tree/arm-none-eabi
Files scripts/build/cc/100-gcc.sh and config/cc/gcc.in.2 .

If you have concerns about my recent pull requests (#182, #183) to have side effects on the core
gcc, I can guard them with the new build step parameter. I doesn't like this, because for the missing
parameters from #182 the gcc defaults are used, if they aren't defined.
Moreover, I would like to build the core-gcc as similar as possible to the original arm-none-eabi
build script.
But if you have doubts ... .

jasmin-j:
I just discovered, that gcc 4.9.3 has no "--with-libelf" configure option and LTO is enabled by default.
So the correct solution would be __adding__ "--en/disable-lto" to "do_gcc_backend" to be able to
disable LTO, if this is configured.
We could keep the "--with-libelf" argument which seems to be ignored or we guard it with a gcc version check?

@bhundven what do you think?

jasmin-j:
I will prepare a patch to add "--en/disable-lto" to "do_gcc_backend".
-> Close this pull request, because it is obsolete then.

Issue #179, opened 2015-09-22
Fail to build with isl-0.15


jserv:
While GCC graphite optimization is enabled, crosstool-ng fails to build with isl version `0.15`. Here are the messages:
```
[EXTRA]    Building gcc
[ERROR]    /tmp/ctng/.build/src/gcc-5.2.0/gcc/graphite-poly.h:398:43: error: 'isl_constraint' has not been declared
[ERROR]    /tmp/ctng/.build/src/gcc-5.2.0/gcc/graphite-poly.h:402:35: error: variable or field 'debug_isl_constraint' declared void
[ERROR]    /tmp/ctng/.build/src/gcc-5.2.0/gcc/graphite-poly.h:402:35: error: 'isl_constraint' was not declared in this scope
[ERROR]    /tmp/ctng/.build/src/gcc-5.2.0/gcc/graphite-poly.h:402:51: error: expected primary-expression before ')' token
[ERROR]    make[2]: *** [graphite.o] Error 1
...
[ERROR]    make[2]: *** [graphite-dependences.o] Error 1
[ERROR]    make[1]: *** [all-gcc] Error 2
```

jasmin-j:
> While GCC graphite optimization is enabled, crosstool-ng fails to build with isl version 0.15.
What was your target?
What additional options did you use?

jserv:
Target: arm-unknown-linux-gnueabi
* binutils: 2.25.1
* gcc: 5.2.0
    - Graphite loop optimization enabled
    - LTO + gold linker enabled
    - plugins enabled

jasmin-j:
I checked this now and I found gcc 5.2.0 will not compile with isl 15.0.
There is a patch available to make this working:
  http://ftp.clfs.org/pub/clfs/conglomeration/gcc/gcc-5.2.0-isl-1.patch
But I guess it is not worth to add this patch to CT-NG, as gcc 5.2.0 will compile with
isl 14.0.

I am just compiling an arm-unknown-linux-gnueabi compiler with isl 14.0 and then I will provide a patch to remove isl 15.0 from CT-NG.

Sorry for the inconvenience, I didn't expect such troubles when I added isl 15.0 as a new
configuration option with commit 218bea6f24fe3b49a54d14e26f7f0922aee2c0c . Moreover, I was
new to CT-NG and didn't realize, that always the newest tool versions are used as a default.
This shows again, how important the regression tests would be.

jasmin-j:
[...]

PR #185, opened 2015-09-22
Remove isl 0.15

jasmin-j:
Remove isl 0.15, because it will not compile with gcc 5.2.0 without patching gcc.


jasmin-j:
This will fix #179.

bhundven:
:)

We'll revisit this with gcc6.

Issue #115, opened 2015-06-09
ltrace fails to compile for armv5te


voltagex:
```
[INFO ]  =================================================================
[INFO ]  Installing gdbserver
[INFO ]  Installing gdbserver: done in 20.69s (at 25:26)
[INFO ]  =================================================================
[INFO ]  Installing ltrace
[ERROR]    ltrace-elf.c:221:42: error: logical not is only applied to the left hand side of comparison [-Werror=logical-not-parentheses]
[ERROR]    ltrace-elf.c:221:42: error: logical not is only applied to the left hand side of comparison [-Werror=logical-not-parentheses]
[ERROR]    ltrace-elf.c:221:42: error: logical not is only applied to the left hand side of comparison [-Werror=logical-not-parentheses]
[ERROR]    make[3]: *** [ltrace-elf.lo] Error 1
[ERROR]    make[2]: *** [all-recursive] Error 1
[ERROR]    make[1]: *** [all] Error 2

```
Config https://gist.github.com/voltagex/4ef4a0b8df4dcd2f5b48, crosstool-NG git+bf5d513

bhundven:
Thanks for reporting! I'm pretty sure this is a gcc-5 issue. I'll look into this later today!

bhundven:
I'm running a build with CT_ARCH_ARCH set to armv5te with some update patches from ubuntu.

voltagex:
@bhundven ah, should I be using an older GCC?

bhundven:
No, you're using 5.1.0

bhundven:
Ok, so this isn't just a gcc-5 issue, this is more of an arm armv5te issue. I'm still looking into this, but I'm trying to get 1.22.0 out, so I'm going to push this out to 1.23.0.

voltagex:
No worries, I suspect there aren't too many people affected by this.

Thanks for your help.

bhundven:
CI was broken due to ltrace not being able to build. I'm currently disabling it in the samples, and just revert #188 when we fix ltrace so that it is also tested.

bhundven:
@stilor has found that ltrace is broken because gcc is installing libstdc++.la, which is wrong.
He is working on a patch to not install that library so that ltrace can be re-enabled.

bhundven:
This has been fixed in #224

PR #188, opened 2015-09-29
Remove ltrace from samples

bhundven:
As of right now, ltrace fails to build. To get CI happy, lets just
disable it!

When we fix ltrace, just revert this change.

This is also in reference to bug #115


bhundven:
Build passes. Merging.

PR #189, opened 2015-09-29
configure: Add automated build option

bhundven:
This commit introduces a configure time option to let the build know
that this is going to be an automated build.

This forces the build to disable the progress bar, log tool warnings,
and force the log level to debug.


bhundven:
This change doesn't enable the new configure option ```--enable-automated```, it merely verifies that the original behavior is the same. If you check the build log in travis-ci, you'll still notice the same ```\ | / - \ | / - \ | /``` progress bar eating the log file.

PR #190, opened 2015-09-30
Fix bfin-unknown-linux-uclibc.

stilor:
Pin GCC to 4.5.3 and binutils to 2.22, the last versions working.
See:
* https://sourceware.org/bugzilla/show_bug.cgi?id=17334 (binutils bug)
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=47779 (gcc bug #1)
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55144 (gcc bug #2)


stilor:
Botched my branches... will resubmit.

PR #191, opened 2015-09-30
travis-ci: Enable new automated build option

bhundven:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ addons:
 # Building crosstool-NG core
 install:
     - ./bootstrap
-    - ./configure --enable-local
+    - ./configure --enable-local --enable-automated
     - make
 
 # Here is the list of all the standard samples tracked
@@ -43,7 +43,7 @@ env:
 # Building the standard samples
 script:
     - ./ct-ng $CT_SAMPLE
-    - CT_LOG_DEBUG=y CT_LOG_LEVEL_MAX="DEBUG" ./ct-ng build.2
+    - ./ct-ng build.2
 
 # On failure displaying the last lines of the log file
 after_failure:


bhundven:
Hrm. gcc does sometimes take > 10 minutes to compile. Is there a way to extend the timeout?

bhundven:
I also note: "All checks have failed", where 4 tests were ran, and 2 failed. I'd like a fix for that as well.

jasmin-j:
> Is there a way to extend the timeout?

Theoretically you could use "travis_wait" (see: http://docs.travis-ci.com/user/build-timeouts/#Build-times-out-because-no-output-was-received), but this will work for 20 minutes only.
I guess the approach from JML #193 is a better solution.

jmlemetayer:
@jasmin-j The #193 is actually inspired by [`travis_wait`](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/templates/header.sh#L108).

@bhundven I think the whole matrix build (the 4 builds) count for one test only.

jasmin-j:
> I think the whole matrix build (the 4 builds) count for one test only. See this blog post.

I don't believe this. When you look to [this] (https://travis-ci.org/crosstool-ng/crosstool-ng/builds/83047634) build result, the sum of all the builds does exceed 50 minutes (ran for 2 hrs 35 min 16 sec).
Or do they count CPU time instead of real elapsed minutes?

jmlemetayer:
I think, the total time restriction applies to a single virtual machine. So in our case for each sample. Moreover, regarding [my first build which I cancelled manually after 1h15](https://travis-ci.org/jmlemetayer/crosstool-ng/builds/80435130), the total time has probably been extended.

bhundven:
A better solution was found in #202

PR #193, opened 2015-09-30
Manage travis-ci restrictions

jmlemetayer:
I've been working on a script to manage travis-ci timming restrictions and other stuff.

I've seen that you recently push another solution with these two patches:

 * 4f266e1a8c78cec2c425b17784576db56125bbbd
 * d6413fe29fbdcd72a04b3d5820ba2e134dbaf558

My idea was also to manage the `CC=` stuff in this script to be able to test both `gcc` and `clang` compilers.

Note that there is something wrong because I have this error on some builds:

    g++: internal compiler error: Killed (program cc1plus)

which means that we are out of memory ... so It's still under development.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,43 @@
+# Add current directory to PATH
+export PATH="$(pwd):$PATH"
+
+# Manage the travis build
+ct-ng_travis_build()
+{
+	# Override the log behaviour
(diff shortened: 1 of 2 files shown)


jmlemetayer:
I've add a fix for the out of memory issue. I've tested on my server but not on travis-ci yet. Waiting for the result.

jmlemetayer:
Replaced by #202.

PR #200, opened 2015-10-01
Fix link error in arm/uclibc with GCC 5.x

stilor:


stilor:
Also fixes other arm*-*-uclibc* samples with GCC 5.2.0.

PR #202, opened 2015-10-03
Manage Travis-CI build

jmlemetayer:
Here is a proper implementation of #193.

I have also added the revert of your commits which - if you choose this option - are no longer needed.


bhundven:
I also think this is a better way to do this.

PR #220, opened 2015-10-21
Enable all working samples in Travis.

stilor:


bhundven:
Muhahahaha >;)

stilor:
You asked for it :) ~30 min per sample, 46 samples, 5 workers => about 5 hour test cycle.

bhundven:
Yikes! Ok, I need to dig into their docs to learn more now, I guess. :D

stilor:
Since Travis is not going to give us more workers, this is unfeasible (~7 hours test cycle). Perhaps, we should use 5-10 samples, but we probably need a more representative set (e.g., covering arm/m68k/mips/powerpc/s390/x86_64 arch, glibc/uclibc/newlib/mingw as libc, etc.)

On the bright side, `ct-ng build-all` now works and can be used locally to run the tests fully. I plan to enhance it a bit further to allow "selectors", e.g. "run all `*-*-linux-uclibc`.

bhundven:
I appreciate you doing the full build all samples and fixing build-all!
While travis-ci does help us test the build, it really doesn't tell us if the toolchain is usable or not.
We'll get there.

PR #221, opened 2015-10-21
Add support for uClibc-ng

bhundven:
This set of commits updates support for uClibc-ng.
More samples need to be updated and added.


bhundven:
@stilor and @yann-morin-1998 go ahead and review again.

I updated the global CT_Kconfig* functions and used a sed operation in the rewrite of the manage_uClibc_config instead of using CT_KconfigSetOption incorrectly.

Issue #222, opened 2015-10-21
Only support latest gcc releases


bhundven:
Starting with crosstool-NG-1.23.0, we will only support the latest supported releases of gcc.
For example:

* 4.8.5
* 4.9.3
* 5.2.0

Just as you would find by going to https://gcc.gnu.org

If you really need an older version of gcc, I have nothing against supporting older releases of crosstool-ng that have your compiler version in it. When you open a pull request, just make sure you change the target branch to the correct version of crosstool-ng, and not use master. When that branch functions the way you expect, I will push a branch release (say: 1.20.1).

In the mean time, in 1.22.0 milestone timeline, I will be removing older releases of gcc (I.E. 4.9.0, 4.8.1) and only support the latest of each minor version (I.E. 4.6.4, 4.7.4, 4.8.5, etc..).

This issue is here to track the progress of these changes.

bhundven:
#232 has been merged, which now limits the number of gcc releases in crosstool-ng to the target for 1.22.0.

pfalcon:
From #241:

> Also, as per #222, I am soon to drop older gcc releases (as well as other components, like uclibc, glibc, gdb, binutils, etc...).

My concern here is dropping support for libc versions. The problem is that different versions of libc's are not compatible. This in particularly true for uClibc, which manages to break binary compatibility between micro releases (each of which is spaced few years apart, yeah). Supposed I want to build a cross-toolchain for uClibc 0.9.28, what are my options?

bhundven:
Get the 1.21 branch, and use it. If you have problems with the 1.21 branch, open a bug and we'll fix the problems, but no new features will come to 1.21. Only stability fixes. Same with older branches of crosstool-ng.

At some point I have to stop supporting older ancient releases. It is making the code unmanageable. So there aren't many options, but that is your option.

pfalcon:
Well, that doesn't sound too bad, thanks for reply.

PR #225, opened 2015-10-28
Unbreak nios2-elf-mingw32 sample

stilor:
Do not use non-CT environment variables in the config (by default, the sample tries to write into build system's root). Disable with-system-zlib and cross-gdb (for now); see the commit messages for details.


bhundven:
I would suggest doing the promotion in a separate PR, and we'll get back to this one after we know that the promotion of expat and ncurses didn't break anything.

Also, I noticed (I'm not hip with the times as much as I should be) that ncurses 6 is out. Maybe we should update that as well. I'm still unsure if we need to stick with the ncurses 5 ABI, or if gdb will work with the new ncurses 6 ABI? That might be a point of experimentation (with some prior versions of gdb). But also keep in mind issues like #222 where I am going to start trimming older versions of gdb as well. (I still need to open an issue on that)

stilor:
I agree with a separate PR for expat/ncurses changes, but I think this change should not be held up by that task. See, this sample had other breakages in addition to expat/ncurses issue, so I don't want this sample to get broken further while the "promotion" is being worked on. Once expat/ncurses are done, we can always re-enable cross-gdb in this sample, just as we did with ltrace.

bhundven:
Fair enough. Merged. Thanks! I'm guessing you are going to work on the promotion? Or should I do that?

stilor:
I can work on that.

bhundven:
@stilor cool. Thanks! Let me know if you need help!

Issue #219, opened 2015-10-21
Add support for building components from version control


bhundven:
This issue is opened to track adding support for building various components from version control.
Whether or not that is the upstream version control or a local repository is not the goal.

Currently I am looking at supporting the following version control systems:
- git
- svn
- mercurial
- bzr
- perforce
- cvs

Please update this ticket if your VCS is not listed.

stilor:
Added. Supports CVS/SVN/Git/Mercurial; we don't have other types used by any of the packages we use - if one wants to add, say, perforce for use in private repo - PRs welcome.

PR #232, opened 2015-10-29
gcc: Support only the latest branch releases of gcc

bhundven:
gcc: Support only the latest branch releases of gcc

This change, as per #222, reduces the number of supported releases of
gcc to the latest branch releases.

I noticed while doing this work that gcc-4.5.4 was never added, so I
moved patches for gcc-4.5.3 to 4.5.4 and updated the
bfin-unknown-linux-uclibc example. Also, 120-siginfo.patch was fixed
upstream in the 4.5.4 release, so this patch is omitted.

I also bumped the avr sample to 4.9.3 from 4.9.2.

With the addition of gcc-5.x, the gcc release team now releases the
major.minor.0 versions, while updates to the branch are available in
svn/git. We'll address that when we get to issue #219. This change just
removes CC_GCC_5_1 and moves CC_GCC_5_2 to CC_GCC_5, and removes
CC_GCC_5_1_or_later and moves CC_GCC_5_2_or_later to CC_GCC_5_or_later.

This is the first of two part changes, as mentioned in #222.
This change is slated for release in 1.22.0. The next change will be
slated for 1.23.0, and will limit gcc versions to what is on
https://gcc.gnu.org under "Release Series and Status", which is
currently 4.9.3 and 5.2.0, although I will also support the previous
supported version. In this example that would be 4.8.5.

Last, but not least, this change also retires AVR32 support.


bhundven:
@tpetazzoni This gets us closer to less versions. I need to get patches for avr ported forward to a newer gcc version (preferably 5.2.0). I see these patches in buildroot, and I may just take those.
I still need to add tracking issues for other ct-ng components with the same goal.

tpetazzoni:
@bhundven Are you talking about AVR or AVR32 ? Buildroot never supported AVR since it's a 8 bits architecture that doesn't run Linux. We did support AVR32 a while ago, but removed its support because only gcc 4.2 with a massive amount of patches would work, and AVR32 is a deprecated architecture.

bhundven:
@tpetazzoni Ah! Ok. I will plan removal of avr32 with this series then!

bhundven:
@yann-morin-1998, @tpetazzoni: Ok, as promised. This change now also removes AVR32 support.

PR #236, opened 2015-11-01
[RFC] Do not simply overwrite previously saved states

stefanct:
This made me rather angry last night...
Patch is completely untested and I don't really know
what I am doing there but it seems easy enough to
propose a change by showing what I have in mind.

Even better than a simple abort would be to check if
the configuration or prerequisites actually changed and
resume automatically etc. like actual makefiles do.
However, anything that does less harm to the user is a
good thing IMHO, especially if I am the user ;) so this
would be a nice step forward.

I am open for suggestions but would rather see the ct-ng
team do something with the idea than me implementing
any details :)


bhundven:
@stefanct I'm going to personally track this pull request.
I understand what you're trying to do, but I think this needs to be planned/designed out.
When we can come up with a better solution, the pull request will mention this one in reference.

PR #238, opened 2015-11-03
Fix some cross_compile issues

bhundven:
While debugging new support for uClibc-ng, I noticed issues with not yet supported arc, and blackfin architectures.

* Linux needs CROSS_COMPILE set during the install/check headers step, as some architectures (notably arc and blackfin) set a default CROSS_COMPILE if it is not set.

* uClibc build script was using a "compat" command line argument, but we should use CROSS_COMPILE instead of CROSS to define our cross compiler.

### DIFF
--- a/scripts/build/kernel/linux.sh
+++ b/scripts/build/kernel/linux.sh
@@ -119,6 +119,7 @@ do_kernel_install() {
     CT_DoLog EXTRA "Installing kernel headers"
     CT_DoExecLog ALL                                    \
     make -C "${kernel_path}"                            \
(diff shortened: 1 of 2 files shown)


bhundven:
@jpf91 @mingwandroid One question for you two on this change is if this affects cross-canadian/cross-native/native?

mingwandroid:
I think it will be ok. I really need to get some complicated canadian cross CI samples going ASAP.

jpf91:
I don't think this will cause any problems. But we'll see ;-)

PR #239, opened 2015-11-06
Canadian cross build = x86_64 Cygwin host = x86_64 MinGW_W64 target = x86_64 GNU/Linux

mingwandroid:
The sample config needs work. It's got lots of cruft in there. I attempted to minimise it but when I tried to use my minified version, it didn't detect it as canadian for some reason. Weird!


mingwandroid:
@bhundven For some reason, on Cygwin, building libtool requires help2man so I added it in 898a8ea. This doesn't seem to be the case on GNU/Linux. I've no idea why. I'll have to leave this until late next week then unless you can figure it out? If anyone else is able to test in the meantime it'd be much appreciated.

mingwandroid:
I've investigated this and I think we need to add help2man as a dependency to be honest.

The reason we're running into it is because I added a patch for libtool from Cygwin - 2.4.5-pass-ldflags.patch - that changes ltmain.in that re-creates build-aux/ltmain.sh which is a dependency of doc/libtool.1 which requires the help2man program.

Alternatively, I could hack `do_companion_tools_libtool_build()` as follows:

    make || touch "${CT_SRC_DIR}/libtool-${CT_LIBTOOL_VERSION}/doc/libtool.1"

.. but then you get this ugliness in the log:

    [EXTRA]    Installing libtool
    [01:53] - /home/ray/cml/.build/src/libtool-2.4.6/build-aux/missing: line 81: help2man: c
    WARNING: 'help2man' is missing on your system.
             You should only need it if you modified a dependency of a man page.
             You may want to install the GNU Help2man package:
             <http://www.gnu.org/software/help2man/>
    make[3]: *** [/home/ray/cml/.build/src/libtool-2.4.6/doc/libtool.1] Error 127
    make[2]: *** [all-recursive] Error 1
    make[1]: *** [all] Error 2
    [EXTRA]    Installing libtool: done in 65.01s (at 02:17)
    [INFO ]  Extracting, patching and installing companion tools: done in 123.44s (at 02:17)

.. which I think is worse. Fact is I'm adding a dependency that's actually needed.

yuxuanchiadm:
Can not building GCC with --enable-plugin. Take a look at this: http://zachsaw.blogspot.jp/2012/10/building-cygwin-gcc-453-with-enable.html

yuxuanchiadm:
In menuconfig (Binary utilities->Enable support for plugins) can switch BINUTILS_PLUGINS on or off at ./config/binutils/binutils.in:295. And CC_GCC_ENABLE_PLUGINS at ./config/cc/gcc.in:335 will depends on BINUTILS_PLUGINS. And ./scripts/build/cc/100-gcc.sh:875 will add --enable-plugin when make GCC.

bhundven:
@yuxuanchiadm we don't have gcc-4.5.3. We have gcc-4.5.4. Also, in 1.23.0, the oldest gcc that will be available will be 4.7.4. Does 4.5.4 also have issues with ```--enable-plugin```?

bhundven:
@mingwandroid It seems to me that the build is failing because help2man needs to be added to:
https://github.com/crosstool-ng/crosstool-ng/blob/master/.travis.yml#L22

yuxuanchiadm:
I`m using x86_64 Cygwin GCC 4.9.3 to build. Still have issues with --enable-plugin.

mingwandroid:
@bhundven done, but please help me with the sample config. I've not used defconfig before ..

bhundven:
@mingwandroid ```ct-ng saveconfig```

mingwandroid:
@bhundven @stilor please review.

stilor:
Looks good to me, just a couple of notes:
- if my expat/ncurses PR gets merged first, you'll need to add empty do_x_for_target in the companion lib scripts (if yours goes in first, I'll have to do it)
- I am still not sure why we can't have just 'cross-gdb' and 'gdbserver' (and require native-gdb), but let's revisit that after both mine and your PRs get merged. Was it failing in this configuration, or did you change anything to make it fail?

mshal68:
na, graue haare hab ich noch keine, aber ne zeitlang gedauert hats, wa ???? :)

bhundven:
@mshal68 Please use english. Thank you.

mingwandroid:
Should be ready to merge now.

bhundven:
@mingwandroid Great! Gonna wait for the travis work to finish, as well as my uclibc-ng work. Will merge after both are done. Could be tomorrow.

PR #242, opened 2015-11-09
Restore uClibc-based samples.

stilor:
Commit 1a25115a1851d3defdf4d37825d8a291be078e53 deleted non-GCC related
files, including the patch for uClibc to compile with Linux kernels after
3.4.

uClibc 0.9.30 patches are not restored by this change (0.9.30 is broken
with recent kernels for multiple other breakages in addition to that; if
not retired, it needs to be fixed properly).


bhundven:
Very quickly, I am almost done testing new uClibc-ng update, which also removes older uClibc versions (except for 0.9.33.2 and custom). I'm going to close this pull request as it won't be relative when that is merged. Thanks! Luckily that was the only mistake I made with that gcc commit. :)

bhundven:
Crud. I also didn't notice this was for 0.9.33.2. Restoring and merging.

PR #245, opened 2015-11-10
Use install-strip target for gcc optionally

villytiger:
Starting from GCC 4.6 there is an install-strip target in Makefile which is provided for stripping GCC executables and libraries. The most important for me is that this option allows to strip libstdc++.a so that I can use static libcstdc++ without debugging symbols.

### DIFF
--- a/config/global/paths.in
+++ b/config/global/paths.in
@@ -117,12 +117,22 @@ config INSTALL_DIR_RO
       
       Useful for toolchains destined for production.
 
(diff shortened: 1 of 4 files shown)


bhundven:
Otherwise, this commit looks good to me.
I think in later commits I'll look at using the new ```STRIP_TARGET_TOOLCHAIN_EXECUTABLES``` to strip other tools and libraries if that option is enabled.

PR #247, opened 2015-11-11
dmalloc: Symlink SO Version numbers

bhundven:
This commit adds the version numbers to dmalloc libraries.
Some programs/autoconf look for the library with version numbers.


bhundven:
I'm dropping this patch for now, as I don't see it's value.

PR #249, opened 2015-11-11
Promote expat/ncurses to 'companion libs'

stilor:


bhundven:
You'll have to rebase this because of the #248

Issue #243, opened 2015-11-09
Add support for the Xtensa architecture


bhundven:
This issue has been opened to track support for Xtensa, and discussion on that process.

bhundven:
> jcmvbkbc said: (from https://github.com/crosstool-ng/crosstool-ng/issues/241)
> So far we have the following patches for xtensa:
>
> patches to the crosstool-NG itself. Besides adding architecture bits they add hooks to do_{binutils,cc,debug_gdb}_extract functions that allow us to overwrite selected parts of binutils, gcc and gdb with processor-specific files. This is done by unpacking corresponding subtree of the configuration overlay archive. An example of such archive: https://github.com/jcmvbkbc/crosstool-NG/blob/lx106/overlays/xtensa_lx106.tar
> xtensa-specific patches backported from development trees of binutils/gcc/gdb: fixes or improvements that old toolchain components need.
> xtensa-specific patches that haven't got to any toolchain component repository: e.g. call0 ABI support patches for gcc-4.8.2 or xtensa support for newlib-2.0.0.
>> What I would see as acceptable is if patches that are applied to components of crosstool-ng (I.E.: gcc, gdb, binutils, etc...) can be applied without affecting other targets.
>
> All changes that we do to binutils, gcc and gdb in extract hook apply to xtensa-specific files only.
Here's the list of files from configuration overlay:
>
>binutils/bfd/xtensa-modules.c
>binutils/include/xtensa-config.h
>gcc/include/xtensa-config.h
>gdb/gdb/gdbserver/xtensa-xtregs.c
>gdb/gdb/regformats/reg-xtensa.dat
>gdb/gdb/xtensa-config.c
>gdb/bfd/xtensa-modules.c
>gdb/include/xtensa-config.h
>
>So where do I start?

bhundven:
1. First, I've set the milestone for this support to be in 1.23.0. With that, the patches will need to be rebased to recent binutils, gcc, gdb, newlib.
2. The patches should go where they belong, in patches/{binutils,gcc,gdb,newlib}
3. Add architecture specific options in config/arch/xtensa.in{,.2}
4. Add (if applicable) scripts/build/arch/xtensa.sh

bhundven:
I thought this was going to take longer, so I marked the milestone for 1.23... seeming that it is in, I'm going to reset and close this issue :)

Thanks again @jcmvbkbc!

PR #253, opened 2015-11-13
Add support for Xtensa architecture

jcmvbkbc:
Hi Bryan,

this series adds support for Xtensa architecture to crosstool-NG (issue #243).
There are architecture-specific patches for the latest binutils, gcc and gdb.
Patch for the newlib is not ready for mainline and needs more work.
Please let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ env:
     - CT_SAMPLE=mips64el-n64-linux-uclibc
     - CT_SAMPLE=powerpc-e500v2-linux-gnuspe
     - CT_SAMPLE=x86_64-unknown-linux-uclibc
+    - CT_SAMPLE=xtensa-unknown-linux-uclibc
 
 # Building the standard samples
 script:
--- a/config/arch/xtensa.in
+++ b/config/arch/xtensa.in
@@ -1,0 +1,20 @@
+# xtensa specific configuration file
+
+## select ARCH_SUPPORTS_32
+## select ARCH_SUPPORTS_BOTH_MMU
+## select ARCH_DEFAULT_HAS_MMU
+##
+## help The xtensa architecture
+## help
+## help    Xtensa is a configurable and extensible processor architecture.
+## help    Supporting a specific configuration typically requires minor
+## help    modifications to a small set of configuration files in various
+## help    development tools. This process is automated and only requires
+## help    a configuration specific 'overlay' file.
+## help
+## help    For a custom configuration, select the XTENSA_CUSTOM option and
+## help    provide the name of the overlay file through the
+## help    CT_ARCH_XTENSA_CUSTOM_NAME option.
+## help
+## help    The default option (ARCH_xtensa_fsf) uses a built-in configuration,
+## help    which may or may not work for a particular Xtensa processor.
--- a/config/arch/xtensa.in.2
+++ b/config/arch/xtensa.in.2
@@ -1,0 +1,30 @@
+choice
+	prompt "Target Architecture Variant"
+	default ARCH_xtensa_fsf
+config XTENSA_CUSTOM
+	bool "Custom Xtensa processor configuration"
+
+config ARCH_xtensa_fsf
+	bool "fsf - Default configuration"
+
+endchoice
+
+config ARCH_XTENSA_CUSTOM_NAME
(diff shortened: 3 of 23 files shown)


bhundven:
@jcmvbkbc Great! I'm very excited to have this support added!

On question. I just added uClibc-ng 1.0.8, and reworked architecture specific changes in scripts/build/libc/uClibc.sh (manage_uClibc_config()). Is there anything uClibc/uClibc-ng needs for xtensa support?

jcmvbkbc:
> Is there anything uClibc/uClibc-ng needs for xtensa support?

No, currently they don't need anything special.

bhundven:
> No, currently they don't need anything special.

@jcmvbkbc Great!

One last request, two part.

1. Please add a sample, do a build with ct-ng, and when the build completes and works properly, do ```ct-ng saveconfig```, and it should put something like: xtensa-unknown-linux-... in the current directory. Copy that into the samples/ directory and add that to the commit.
    * Bonus points for adding the sample to the .travis.yml file so that we also check this sample when we modify anything in crosstool-ng to make sure we don't break xtensa support! ;)

jcmvbkbc:
Ok, will do both.

jcmvbkbc:
Done.

bhundven:
Thank you!

pfalcon:
@jcmvbkbc : Thanks for handling this! Just to clarify: what about gcc 4.x branch?

jcmvbkbc:
@pfalcon : I see two big options here:
- merge it to the mainline if it's acceptable (a new xtensa-specific feature addition to gcc-4.8)
- merge mainline back to my branches to fix issues of the old baseline.

Even if merging it to mainline is possible it still wouldn't be complete until newlib is done.

pfalcon:
@jcmvbkbc : I'd consider both options good, but if it can be merged into mainline, and that's easy enough (as this PR shows) - why not? As @bhundven told, he's going to keep maintaining only the last micro version of each 4.7.x, 4.8.x, 4.9.x, but I'm sure there's no talk about dropping those gcc branches any time soon (and I assume any patches would be easily portable across micro versions (though may be worth asking @bhundven what maintenance involvement he expects from contributors)).

And choosing gcc 4.8 for xtensa support is good too - indeed, aggressive undefined behavior optimizations started in 4.9.

P.S. Sorry I can't directly test your patches now, as going from sickness into travel. I'll do that as soon as I can. I also pinged folks like @projectgus and @foogod to have a look too.

bhundven:
@pfalcon I'll maintain any branch as long as someone is willing to do the work.
I'm focused on the master branch and moving ct-ng forward.
I'd prefer that patches to make architectures work properly to be pushed upstream (gcc, glibc, binutils, newlib, etc...), as my job is not to maintain patches, but ct-ng itself.

jcmvbkbc:
@pfalcon : ok, second option for now: 1.22.0-based branch with support for esp8266 and esp32 (configurations called xtensa-lx106-elf and xtensa-esp108-elf with binutils-2.25.1, gcc-4.8.5, gdb-7.10) is here: https://github.com/jcmvbkbc/crosstool-NG/tree/xtensa-1.22.x

bhundven:
4.8.5 will still be available after #222 is complete. The supported versions will be: 4.8.5, 4.9.3 (I see work happening for 4.9.4 in https://gcc.gnu.org/bugzilla/), and 5.x.

https://github.com/crosstool-ng/crosstool-ng/tree/master/patches/gcc/4.8.5  << there are xtensa patches from buildroot in the 4.8.5 patch tree, so it is possible to support 4.8.x in crosstool-ng mainline.

jcmvbkbc:
@bhundven unfortunately there were no support for call0 Xtensa ABI in gcc prior to 5.x, and its support is crucial for esp8266.
With gcc-5.x the only issue is the newlib. The newlib port is rather dirty: e.g. it breaks build for other architectures. I hope to clean it up in late December and try to upstream it.

bhundven:
Is it too difficult to backport call0 support from 5.x to 4.9.x/4.8.x?

jcmvbkbc:
It's not. In fact it was initially developed for 4.8.2.

bhundven:
As you can see in #282, 4.8.5 will be around for a while if you want to backport this support.

PR #262, opened 2015-11-15
configure: Correctly search for tinfo for kconfig

bhundven:
I previously thought that tinfo could only be a link to libncurses in
modern ncurses installations. I was wrong.

Now we check for ncurses, then check for tinfo. If tinfo is a link to
ncurses, it is then not required to link against. But if libtinfo is
providing tgetent, and not libncurses, then we must add it to LIBS as
well.

Also, kconfig shouldn't hardcode -lncurses, it should use what is in
$(LIBS).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -339,15 +339,6 @@ LIBS=$SAVE_LIBS
 AC_SUBST([INTL_LIBS])
 
 #----------------------------------------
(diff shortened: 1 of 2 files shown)


bhundven:
I'm merging this as the build is currently broken on systems without tinfo or a symlink from libtinfo.so to libncurses.so.

PR #266, opened 2015-11-16
Look for tigetnum after {n,}curses{w,}

mingwandroid:
Often it's built into {n,}curses{w,} so this
arrangement fixes a build failure for systems
that don't have libtinfo.so.

Signed-off-by: Ray Donnelly <mingw.android@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -338,16 +338,6 @@ AC_SEARCH_LIBS(bindtextdomain, intl, [test "$ac_res" = "none required" || INTL_L
 LIBS=$SAVE_LIBS
 AC_SUBST([INTL_LIBS])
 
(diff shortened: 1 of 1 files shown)


bhundven:
I pushed #262 because the build was broken, as you mentioned.
I'll close this PR out.

mingwandroid:
My re-ordering also falls through as "none required" since on ArchLinux it already has {n,}curses{w,} in LIBS at that stage and LIBS are searched in subsequent AC_SEARCH_LIBS calls so tigetnum is found in there hence my re-ordering. Not to worry though .. as I said, I did need the other bit of your change and all is good again.

Issue #218, opened 2015-10-21
Multilib support in Crosstool-NG


bhundven:
This issue is open to track the progress of multilib support in Crosstool-NG.
The option for multilib support is already in config/..., but it is not supported.
#383 is the PR that will add support for multilib.

bhundven:
@mingwandroid I've assigned this to you, but this is a group effort.
I would like to see the work going into #217 to be completed first so that the multilib support is simpler.

I would also like to see any work move to an official topic branch in the crosstool-ng tree so that we can all work on it.

bhundven:
This ticket is now tracking #383 to completion.

stilor:
#383 has been completed.

Issue #217, opened 2015-10-21
Two pass gcc build


bhundven:
To simplify the crosstool-ng build process, move from building gcc three times to two times.
This issue is opened to track the progress of this change.

bhundven:
This is going to be delayed until 1.24

bhundven:
@stilor Are you still planning to work on two pass builds? If not, lets close this.

stilor:
Maybe eventually, so let's keep it open as a wishlist item. Not the highest priority though.

patrakov:
Documentation (http://crosstool-ng.github.io/docs/toolchain-construction/) also needs to be changed. In fact, it gives misleading motivation for the three-pass build, so it would be nice to fix it even if the two-pass build is not implemented in the actual script.

cpackham:
#808 duplicates this

PR #275, opened 2015-11-18
Create symlinks for mingw

stilor:
Mingw GCC is configured to look into mingw/include and mingw/lib. Mingw
is itself a symlink to usr/$(CT_TARGET), so the headers/libs end up in
$sysroot/usr/$(CT_TARGET)/include and $sysroot/usr/$(CT_TARGET)/lib. By
default, configure uses /usr prefix (not /usr/$(CT_TARGET)), for install
prefix of the companion libs/tools - so create symlinks from usr/include
and usr/lib to the respective mingw locations.

The right approach, perhaps, would be to make that /usr prefix
overridable (and override it in mingw.sh). However, this will require
changing all the *_for_target() methods, and it is not very testable
right now as some companion tools/libs fail to build in mingw.

With this change, mingw config with native-gdb enabled builds ok - but
fails later (at the `finish' stage), while stripping the gdbserver
(apparently, it doesn't expect .exe suffix on it).

As a side note, mingw builds seem kind of fragile still. Perhaps, mark
them EXPERIMENTAL in 1.22.0?

Signed-off-by: Alexey Neyman stilor@att.net

### DIFF
--- a/scripts/build/libc/mingw.sh
+++ b/scripts/build/libc/mingw.sh
@@ -65,7 +65,16 @@ do_libc_start_files() {
     # It seems mingw is strangely set up to look into /mingw instead of
     # /usr (notably when looking for the headers). This symlink is
     # here to workaround this, and seems to be here to last... :-/
-    CT_DoExecLog ALL ln -sv "usr/${CT_TARGET}" "${CT_SYSROOT_DIR}/mingw"
+    CT_DoExecLog ALL ln -sv "${MINGW_INSTALL_PREFIX#/}" "${CT_SYSROOT_DIR}/mingw"
+
+    # Also create symlinks for companion libraries that want to install
+    # into /usr/lib and /usr/include
+    if [ "${MINGW_INSTALL_PREFIX}" = "/usr/${CT_TARGET}" ]; then
+        CT_DoExecLog ALL rmdir "${CT_SYSROOT_DIR}/usr/include"
+        CT_DoExecLog ALL ln -sv "${CT_TARGET}/include" "${CT_SYSROOT_DIR}/usr/include"
+        CT_DoExecLog ALL rmdir "${CT_SYSROOT_DIR}/usr/lib"
+        CT_DoExecLog ALL ln -sv "${CT_TARGET}/lib" "${CT_SYSROOT_DIR}/usr/lib"
+    fi
 
     CT_EndStep
 }


bhundven:
I think what I'm saying is I want to keep the build consistent.
So rather then clobbering what we've already setup, maybe move this directory modification down the stack to crosstool-NG.sh.in, so that the links and directories that we setup are done right the first time.

bhundven:
@mingwandroid Fair enough. I'm guessing that change will be in with the multilib changes?

When the multilib changes are done, and since this is done early enough (do_libc_start_files), I'd like to still see this change move to crosstool-NG.sh.in.

After seeing that change, I do remember the conversations we had about removing the symlink madness. I was mostly looking for conflicting work in the toolchain setup in this review.

bhundven:
I guess the part I don't understand about this is why:

https://github.com/crosstool-ng/crosstool-ng/blob/master/scripts/build/libc/mingw.sh#L30

Then it's basically reverted with this change, and the official docs say:

```In the example in these
instructions, the sysroot is '/mypath'.  To use a sysroot, pass
--with-sysroot=/mypath and --prefix=/mypath as configure flags, i.e.
	../path/to/configure --with-sysroot=/mypath --prefix=/mypath```

http://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-doc/howto-build/mingw-w64-howto-build.txt#l202

Why not just keep the install prefix as: ```${CT_SYSROOT}/usr```?

mingwandroid:
I need to check into this all a bit more I think @bhundven .. I'll get back to you ASAP, but I'm not finding much spare time for hacking this week unfortunately.

bhundven:
I'm pushing out these mingw pull-requests and issues to 1.23. Mingw shouldn't hold back the release on Friday.

stilor:
@bhundven, I had my doubts about this approach too. I didn't go with what I think is the right approach - selecting --prefix for all target libs/tools to be either /usr (default) or /mingw (mingw override) - because of its pervasiveness. Didn't want to disturb a lot of build scripts before the release, you know :)

But, if we know that mingw builds are easily broken by selecting additional tools/libs (as we currently know, native-gdb/gdbserver/dmalloc/ltrace all cause build failures) - perhaps, mark *-w64-mingw samples as experimental in 1.22?

bhundven:
I will be marking mingw as experimental for the 1.22 release.

I really feel like this is more of an education problem for me, because I don't use windows (desktop since 98/xp, and have done some work with windows server, but for compatibility with ldap/kerberos/freeipa).

On the other hand, every where else that I look at mingw-64 being built, they don't use the --prefix that we use, and I feel confused why ours is different then every other deployment I see?!?

I agree with @mingwandroid about the symlink madness in the sysroot, and I really want to remove that and let gcc do the right thing, with regards to making the paths it expects to need instead of us forcing gcc to use the paths we want, but that is exactly what I see us doing here with mingw.

bhundven:
...unless I'm totally missing something...

mingwandroid:
Overall, I think we shouldn't let mingw-w64 hold up (or worse, mess up) our release plans, and instead I (and you guys too I hope) will focus on mutlilib and mingw-w64 immediately after the 1.22 release.

bhundven:
The first thing I'd like to see done after the release, to ease multilib and other changes is to move to a two-pass toolchain... #217

After that, then #218, then lets come back to mingw-64, as multilib will definitely make mingw-64 a bit more complex(?).

PR #277, opened 2015-11-19
scripts: Be less sensitive to case sensitivity

mingwandroid:
It only matters for The Linux Kernel and glibc so let's not
fail on other setups. Although OS X HFS+ can be formatted as
case sensitive, and the Windows Registry has a flag to enable
it too, they are not the default and many things go wrong when
they are enabled due to years of neglect and ignorance.

Signed-off-by: Ray Donnelly <mingw.android@gmail.com>


mingwandroid:
@bhundven , @stilor , let me know what you think.

bhundven:
I'm closing this PR for now. It can be re-opened later, if still relevant.

PR #283, opened 2015-11-23
Fixes binutils and gold (raspi2 sample builds ok if you disable the static toolchain options)

mingwandroid:
One of the patches came from Google's Android NDK and I kept it as it was for credit and origin tracing purposes.


bhundven:
But, I'll merge this anyways, because I'll probably also merge it to the 1.22 branch for 1.22.1.

PR #287, opened 2015-11-25
binutils/gdb: Fix configure options when building from binutils-gdb repo

bhundven:
When building from:
{git,http}://sourceware.org/git/binutils-gdb.git

and setting BINUTILS_CUSTOM or GDB_CUSTOM, they could be the same
source.

These config options should not affect normal released versions.


jasmin-j:
THX, now I can remove some extra options!
Will test it when I have time again.

PR #288, opened 2015-11-26
Custom locations rewrite

bhundven:
This patch series cleans up and rewrites CT_GetCustom and friends.

We also drop the usage of:

* CT_CUSTOM_LOCATION_ROOT_DIR
* CT_KERNEL_LINUX_USE_CUSTOM_HEADERS

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -4,42 +4,42 @@ export PATH="$(pwd):$PATH"
 # Manage the travis build
 ct-ng_travis_build()
 {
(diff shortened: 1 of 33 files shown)


bhundven:
@stilor I have switch the custom_location_type to a string (being either 'file' or 'dir'), and added a check if it was neither and report an error (for future VCS based location types)

One last thing I'd like to complete before I merge this is to update the ```help``` sections for the custom options.

PR #291, opened 2015-12-01
Cygwin more fixes

mingwandroid:
Fixes https://sourceware.org/ml/crossgcc/2015-11/msg00084.html (expat-for-host library nature (static/shared) to match with the other host libraries; it's linked to gettext and they aren't allowed to differ), then some follow on bugs I encountered trying to build that sample (patches were needed so that GCC plugins can be built on Cygwin). I've yet to verify that it builds completely yet, so this is more for discussion and so I can get testing feedback from Thies Peter, so don't merge yet, even if everything  looks good (which I doubt).

Cheers!

### DIFF
--- a/patches/gcc/5.3.0/370-gcc-plugin-Win-Dont-need-undefined-extern-var-refs-nor-fpic.patch
+++ b/patches/gcc/5.3.0/370-gcc-plugin-Win-Dont-need-undefined-extern-var-refs-nor-fpic.patch
@@ -1,0 +1,160 @@
+diff -urN gcc-5.3.0.orig/config/gcc-plugin.m4 gcc-5.3.0/config/gcc-plugin.m4
+--- gcc-5.3.0.orig/config/gcc-plugin.m4	2015-12-19 14:39:04.120734900 +0000
++++ gcc-5.3.0/config/gcc-plugin.m4	2015-12-20 01:28:45.381965300 +0000
+@@ -20,6 +20,9 @@
+ 
+    pluginlibs=
+ 
++   PICFLAG="-fPIC"
++   UNDEFINEDPREAMBLE="extern int X;"
++   UNDEFINEDCODE="return X == 0;"
+    case "${host}" in
+      *-*-darwin*)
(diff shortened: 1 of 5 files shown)


mingwandroid:
Urgh, I see an immediate problem, I wanted the MinGW-w64 fix to either appear and then perhaps not be included at all finally. Will rebase -i then force push shortly.

bhundven:
@mingwandroid the dlfcn-win32 stuff is GPLv2.1, I don't see why that code is not either:

1. a patch to gcc to fix gcc's support for win32... and at that point...
2. a message to gcc-patches@ with the said patches to actually upstream the work to gcc.

mingwandroid:
@bhundven I want to investigate if GCC 6.0 has already done some work for libdl-type support on Windows or not, if not then I may consider it. Anyway, it's outside the scope of this pull request (cygwin fixes) pull request and shouldn't have been included. I will remove it when I get a chance.

mingwandroid:
Removed the non-cygwin part. Testing incoming. I need to check if AppVeyor have support for Cygwin, or just get MSYS2 support done as I know AppVeyor support that.

mingwandroid:
@bhundven I guess I should port the gcc 5.2 changes to 5.3 now? (as well as 5.2 or instead of 5.2?) Also, what's happening with 6? Should I do the same for that? And also for git?

bhundven:
Don't worry about 5.3 or 6 yet.

bhundven:
Beyond the two patches possibly being backported to older, supported, gcc versions... Looks good to me!

mingwandroid:
Err .. I forward ported them to 5.3 instead!

The configury layout changed a lot unfortunately between 5.x and 4.x (e.g. config/gcc-plugin.m4 doesn't exist in 4.9.3 or 4.8.5, nor does the libcc1 folder) so porting the patches isn't trivial.

mingwandroid:
GCC 5.2.0 patches removed now.

bhundven:
@mingwandroid Thanks!

PR #295, opened 2015-12-07
gcc: add GCC 5.3.0

FreddieChopin:
Signed-off-by: Freddie Chopin <freddie.chopin@gmail.com>

### DIFF
--- a/config/cc/gcc.in
+++ b/config/cc/gcc.in
@@ -37,6 +37,11 @@ choice
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
+config CC_GCC_V_5_3_0
+    bool
+    prompt "5.3.0"
+    select CC_GCC_5
+
 config CC_GCC_V_5_2_0
     bool
     prompt "5.2.0"
@@ -367,6 +372,7 @@ config CC_GCC_VERSION
     string
 # Don't remove next line
 # CT_INSERT_VERSION_STRING_BELOW
+    default "5.3.0" if CC_GCC_V_5_3_0
     default "5.2.0" if CC_GCC_V_5_2_0
     default "linaro-4.9-2015.06" if CC_GCC_V_linaro_4_9
     default "4.9.3" if CC_GCC_V_4_9_3


FreddieChopin:
The build failure seems to be musl's fault. It seems this library is not ready for GCC 5.3. Should I just force GCC version in musl's configuration to be 5.2.0 or maybe some other approach would be better?

bhundven:
Technically it's not musl's fault.
The patches for gcc need to be ported from 5.2.0 to 5.3.0.

I know that new versions are available for gcc and other components, but there is some work to lessen the number of supported gcc versions as well as some clean up activities that are going on.

WIth that, you wouldn't just be adding gcc-5.3.0, but also removing gcc-5.2.0 since gcc-5.3.0 obsoletes gcc-5.2.0 in the gcc-5-branch. It's because of the gcc release process changed from an X.Y release number to just X, so instead of seeing updates released for gcc-5.2.1 - which does happen in svn, but does not happen as an actual release - we only add new versions if gcc-6.0 were to release.

So I would update this PR with those thoughts in mind, as well as realize that #288 and #282 need to happen before I update any versions. So you'll have to rebase this PR after those commits are merged.

FreddieChopin:
OK, I'm going to wait with further work for the patches you mentioned.

bhundven:
Ok, so #282 and #288 have been merged.

If you want to replace 5.2.0 with 5.3.0 (as that's how we roll), and `git mv` the patches from `patches/gcc/5.2.0` to `patches/gcc/5.3.0` and update the PR, I'll be happy to review.

kozyilmaz:
Compared buildroot and crosstools-ng gcc 5 patches:
- 201-libgcc-remove-unistd-header.patch is upstream now
- Buildroot has additional 120-gcc-config.gcc-fix-typo-for-powerpc-e6500-cpu_is_64b.patch for 5.3
- Should keep 110-xtensa-implement-trap-pattern.patch (only difference between buildroot)

bhundven:
Preferring #306 to this PR. Thanks for reporting!

PR #296, opened 2015-12-07
gmp: add version 6.1.0

FreddieChopin:
Signed-off-by: Freddie Chopin <freddie.chopin@gmail.com>

### DIFF
--- a/config/companion_libs/gmp.in
+++ b/config/companion_libs/gmp.in
@@ -6,6 +6,11 @@ choice
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
+config GMP_V_6_1_0
+    bool
+    prompt "6.1.0"
+    select GMP_5_0_2_or_later
+
 config GMP_V_6_0_0
     bool
     prompt "6.0.0a"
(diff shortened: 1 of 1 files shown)


bhundven:
Is this tested with older binutils and gcc combinations?
I would say for binutils, only test 2.23.2, 2.24, and 2.25.1, as soon that's all that will be available.

FreddieChopin:
IS there a script (or a way) to test that automatically? Or should I just create required configs and build them? Generally GCC's docs always say "gmc version x.y.z or later" and it usually worked fine with the most recent version (;

bhundven:
I realize this might work with a recent version, but it needs to be tested with all versions crosstool-ng supports. There really isn't a way to automatically do that at this time, and there are only 3 gcc versions and 3 binutils versions you need to worry about, as I mentioned in my previous response. :)

bhundven:
If it doesn't work, then you need to make sure the newer version depends only on the newer versions of gcc and/or binutils.

bhundven:
I went back and tested the worst case scenario binutils-2.23.2, and gcc-4.8.5, and the build worked fine. Merging.

PR #305, opened 2016-01-02
binutils: Fix installing libiberty for target

### LINKED ISSUE #302: binutils for target did not get libiberty installed
Hi,

I was trying to crossbuild a toolchain with "binutils target install" (libiberty and libbfd) and found that to be not installed (but checking log they were built) after the whole build process is finished.

Upon checking the configure script for binutils (and libiberty inside of it)  libiberty needs --enable-install-libiberty flag to enable the installation.

Thanks.

bhundven:
If CT_BINUTILS_FOR_TARGET_IBERTY is set, then it seems that we also must
set `--enable-install-libiberty` for configure to pickup that it needs
to be installed.

This closes #302

### DIFF
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -351,6 +351,10 @@ do_binutils_for_target() {
             extra_config+=("--disable-multilib")
         fi
 
+        if [ "${CT_BINUTILS_FOR_TARGET_IBERTY}" = "y" ]; then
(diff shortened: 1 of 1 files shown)


tofutim:
I had an iberty error today. [ALL  ]    ld: framework not found ../libiberty/pic/libiberty.a
```
.config:CT_BINUTILS_FOR_TARGET_IBERTY=y
```

PR #273, opened 2015-11-17
native-gdb: Fix search location for expat

modbw:
This fixes building native-gdb when creating a cross toolchain

PR #317, opened 2016-01-09
Fixes 'configure: error: expat is missing or unusable' when native gd…

kongen:
…b is enabled


bhundven:
See pull request #273

kongen:
I see, almost the same but not quite. IMHO it's a bit cleaner to add the prefix of a flag when we add the flag - but in this case the product is the same.

PR #320, opened 2016-01-13
Newlib updates

bhundven:


jasmin-j:
Hello Bryan!

I used your patch for my ARM compiler and it did use the newest newlib package.
By the way, in the meantime a newer package is available: newlib-2.3.0.20160226.tar.gz

I looked to your changes and they are perfectly OK. What exactly did you expect to check by me?

There are a lot more options in newlib, which could be added as CT-NG configuration options,
but using CT_LIBC_NEWLIB_EXTRA_CONFIG_ARRAY is a good method to add them.
From my point of view it is good as it is currently. So why did you add newlib_elix_level,
when the default is 0 anyway?
Where is the difference to all other options which are available in newlib, but have no CT-NG
configuration options?

BR
Jasmin

bhundven:
https://sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=newlib/README;h=bcabfba25892bcd69b895f0b0f281e86558bb7e7;hb=HEAD#l357

bhundven:
I was waiting for anyone to review this change. Anyone. But alas, no one has, so it has sat here.

jasmin-j:
What are your plans for the other missing newlib configure options (e.g. --enable-newlib-register-fini)?
I don't think it is a good idea to add all the currently available configure switches as CT-NG options.
There might be added/removed switches to newlib and CT-NG will get easily out of sync.
On the other hand, it might be useful to have the most important ones in CT-NGs menuconfig with a proper help text.
If this is the direction you like to move, I will have a look on them and add it to CT-NG.
If not, then I ask you again, why you added  the other new CT-NG options (newlib-nano-malloc, newlib-elix-level), if CT_LIBC_NEWLIB_EXTRA_CONFIG_ARRAY would have been sufficient?

bhundven:
@jasmin-j So I get pulled in two different directions. You want me to just add the extra config array, and not worry about adding extra config options.

@yann-morin-1998 suggested that I removed the extra config array, because it adds support burden (if someone misconfigures their newlib, it becomes a crosstool-ng support issue). Instead if the "feature" you are looking to add becomes a configuration option, then it is "supported", and if it breaks later on, we can track it.

The nano options were requested by another user. The elix options are older embedded linux options I was messing around with, but they are no longer used and I plan to remove them from this pull request.

No one has requested --enable-newlib-register-fini, so I don't have that option here.

Does that help you understand where I am coming from?

bhundven:
Updated pull request.

* Removed ELIX options, as mentioned in the previous comment.
* Updated to 2.3.0-20160226

jasmin-j:
Ok, lets put all together:

1. @yann-morin-1998 likes to have (all) possible options as CT-NG configurable items.
2. Moreover he likes to remove the "extra config array" to reduce any possible maintenance.
3. Other users (e.g.: jasmin-j) require currently not available newlib options and need to request them by @bhundven (or the CT-NG developer team), or need to implement them, because the "extra config array" is gone.
4. When a new version (every month) is available, someone needs to check if all the old options are still usable. If not, they can be simply disabled version dependent by the config script, rather to change the CT-NG build scripts.

I can perfectly live with that approach, but I would suggest to allow the "extra config array" with the experimental option, so that people like me can start easier with new samples or new versions.

For my arm-none-eabi (embedded) compiler I need this additional newlib configurations, which I currently
supply via the "extra config array". If it is removed, I would like to have them as CT-NG configuration:
--enable-newlib-register-fini
--enable-newlib-reent-small
--disable-newlib-atexit-dynamic-alloc
--disable-newlib-fvwrite-in-streamio
--disable-newlib-fseek-optimization
--disable-newlib-wide-orient
--disable-newlib-unbuf-stream-opt
--enable-newlib-global-atexit
You don't need to add them. I can do the work, based on this branch.

> Removed ELIX options, as mentioned in the previous comment.

I am not familiar with this. Are you sure nobody uses them?
I hope nobody needs them in the future, when the "extra config array" is gone.

> Updated to 2.3.0-20160226

How do you plan to maintain this? It seems they make a new snapshot every month. It will make a lot of work to update this so frequently.
Moreover, you removed the previous version. This is OK for this development branch, but how do you plan to handle this in the future?
I would suggest to add them less frequently (every 3 months or on request) and to keep at least 3 older versions. You will not have the time to look in every detailed newlib change and it might happen, that a change is bad. Keeping an older version in CT-NG could help us here.

bhundven:
I made a mistake. It's been a long couple of weeks, I apologize!

I confused CC_GCC_EXTRA_ENV_ARRAY with LIBC_NEWLIB_EXTRA_CONFIG_ARRAY.
LIBC_NEWLIB_EXTRA_CONFIG_ARRAY is not going away. The nano options were added for convenience.

CC_GCC_EXTRA_ENV_ARRAY is another topic, and that one is what needs configuration options added.

Although adding supported configure options would also be nice!

Again, I am very sorry for my mistake!

jasmin-j:
> CC_GCC_EXTRA_ENV_ARRAY is another topic, and that one is what needs configuration options added.

We will discuss this in #323 .

> Although adding supported configure options would also be nice!

Done in pull request #371. I did also a compile test with all the new options activated.

> Again, I am very sorry for my mistake!

Hey, that's OK! We are all humans and mistakes can happen ;)

BR
Jasmin

PR #322, opened 2016-01-18
Remove GCC dependency to libelf

modbw:
As crosstools-ng only support GCC >= 4.8 we do not need libelf for gcc. GCC dropped this dependency with 4.6.

Signed-off-by: Matthias Weisser <m.weisser.m@gmail.com>


bhundven:
After this change, what depends on libelf?

modbw:
A quick look around showed me that libelf is build for the target and needed by (at least) ltrace.

PR #326, opened 2016-01-24
patches/binutils: wcsncasecmp is provided by OSX >= 10.7

hanazuki:
OSX SDK has a declaration for `wcsncasecmp` since 10.7, which conflicts with
the definition provided by the patch for binutils.

In `/usr/include/wchar.h` on ElCapitan we have:
```c
/* Additional functionality provided by:
 * POSIX.1-2008
 */

#if __DARWIN_C_LEVEL >= 200809L
__BEGIN_DECLS
// [snip]
int     wcsncasecmp(const wchar_t *, const wchar_t *, size_t n) __OSX_AVAILABLE_STARTING(__MAC_10_7, __IPHONE_4_3);
// [snip]
__END_DECLS
#endif /* __DARWIN_C_LEVEL >= 200809L */
```


mingwandroid:
Thanks!

PR #342, opened 2016-02-17
binutils: fix .init/.fini literals moving in xtensa gas

jcmvbkbc:
Hi Bryan,

this is a backport of a fix for xtensa gas recently committed to binutils.
It fixes build errors seen with projects that have .init/.fini sections and use text-section-literals.


bhundven:
Thanks, Max!

PR #354, opened 2016-02-24
gdb: Add support for v7.11, port patches from v7.10.1 that apply

troglobit:

### DIFF
--- a/config/debug/gdb.in
+++ b/config/debug/gdb.in
@@ -68,6 +68,11 @@ choice
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
+config GDB_V_7_11
+    bool
+    prompt "7.11"
+    select GDB_7_2_or_later
+
 config GDB_V_7_10_1
     bool
     prompt "7.10.1"
@@ -233,6 +238,7 @@ config GDB_VERSION
     string
 # Don't remove next line
 # CT_INSERT_VERSION_STRING_BELOW
+    default "7.11" if GDB_V_7_11
     default "7.10.1" if GDB_V_7_10_1
     default "7.10" if GDB_V_7_10
     default "7.9.1" if GDB_V_7_9_1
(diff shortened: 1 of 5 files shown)


bhundven:
Hrm, broke on x86_64-unknown-linux-uclibc. Might want to check that build and figure out what is going on?

troglobit:
I'll have a look at it, see what I can dig up!

troglobit:
Problem seems to be caused in the gdb-native build, the gnulib included in gdb detects problems with the uClibc-ng `gettimeofday()` and decides to use its own `rpl_gettimeofday()`. The build later fails to link `libincproc.so` due to missing `rpl_gettimeofday()`.   Noticed a similar bug report from 2013 on gettext in Buildroot. But unsure atm what to do, I'll continue looking:

```
configure:13515: checking whether gettimeofday clobbers localtime buffer
configure:13564: result: guessing yes
configure:13580: checking for gettimeofday with POSIX signature
configure:13608: x86_64-unknown-linux-uclibc-gcc -c   conftest.c >&5
conftest.c:136:19: error: conflicting types for 'gettimeofday'
               int gettimeofday (struct timeval *restrict, void *restrict);
                   ^
In file included from conftest.c:134:0:
/home/jocke/x-tools/x86_64-unknown-linux-uclibc/x86_64-unknown-linux-uclibc/sysroot/usr/include/sys/time.h:72:12: not
e: previous declaration of 'gettimeofday' was here
 extern int gettimeofday (struct timeval *__restrict __tv,
            ^
```

From this I can only conclude that gnulib's configure wrongfully guesses the uClibc `gettimeofday()` to be bad.  Two approaches: 1) fix the configure script in the GDB version of gnulib, 2) fix the GDB makefile that forgets to include gnulib when linking the above mentioned library.

troglobit:
Commit f9a704b adds a workaround to the gdb/gnulib/configure script which works on my 12.04 built server.  Hopefully Travis likes it as well.

bhundven:
@troglobit is that sample using uClibc-ng? If not, maybe we need to just move that sample over to use uClibc-ng.

troglobit:
@bhundven The sample is cleverly written to select the latest uClibc, which resolves to uClibc-ng, so I think we're fine.  Also, the patch applies to both uClibc and uClibc-ng.

troglobit:
Alright, all tests in Travis-CI have passed! \o/

bhundven:
Looks good to me!

PR #357, opened 2016-02-26
kernel: Updated kernel revisions

dajamminyogesh:
Kernel revisions updated to match kernel.org.

Signed-off-by: Yogesh Sharma <YogeshASharma@gmail.com>

### DIFF
--- a/config/kernel/linux.in
+++ b/config/kernel/linux.in
@@ -46,7 +46,7 @@ choice
 
 config KERNEL_V_4_4
     bool
-    prompt "4.4.2 (stable)"
+    prompt "4.4.3 (stable)"
 
 config KERNEL_V_4_3
     bool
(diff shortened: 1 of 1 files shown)


bhundven:
Didn't we just update kernel versions 3 days ago?
Another tip. You don't need two commits that do the same thing, you can squash them.
```git rebase -i HEAD~2```
In the editor go to the second commit in the list and change the first word from 'pick' to 'squash', save and quit. Now they are one commit: ```git push --force origin master```, and this pull request is updated.

bhundven:
In the future, when you need to add more to a commit, do the rebase trick that I do, and do the same rebase command, and change the commit from 'pick' to 'edit'. This will drop you to the working directory of the commit before it is committed, and you can edit your files more, then add the files to the index (git add file), and ```git rebase --continue```.

bhundven:
Thanks for the update!

PR #362, opened 2016-03-03
samples: add aarch64-rpi3-linux-gnueabi

bhundven:
This is a aarch64 raspberry pi 3 sample.

Please test!


bhundven:
This looks good to me. Tested the build on Mac and Debian Sid.
Granted there may not be a 64-bit OS to test binaries on yet (and I don't have an RPI3 yet), but this should work. I guess there will be updates if it doesn't ;)

PR #371, opened 2016-03-12
Newlib updates

jasmin-j:
Adding missing newlib configure options.
Based on @bhundven's branch for newlib snapshot releases.

This closes #320

BR Jasmin

PR #372, opened 2016-03-13
Adding new sample for arm-none-eabi

jasmin-j:
This configuration will build an arm-none-eabi cross compiler to be
used for embedded development for this ARM cores:
  armv6-m armv7-m armv7e-m armv7-r

Processors:
  cortex-m0 cortex-m0plus cortex-m1 cortex-m3 cortex-m4 cortex-m7
  marvell-pj4 cortex-r4 cortex-r4f cortex-r5 cortex-r7 cortex-a5
  cortex-a7 cortex-a8 cortex-a9 cortex-a12 cortex-a15 cortex-a17
  cortex-a15.cortex-a7 cortex-a17.cortex-a7 cortex-a53 cortex-a57
  cortex-a72 exynos-m1 xgene1 cortex-a57.cortex-a53
  cortex-a72.cortex-a53

The cross compiler is build as multilib variant for the respective
core and FPU version.

For further details refer to samples/arm-none-eabi/reported.by .

Signed-off-by: Jasmin Jessich jasmin@anw.at


jasmin-j:
Requires pull request #371 (Newlib updates) to be applied before!

jasmin-j:
> "any build"

Sorry, I don't understand. What do you mean?

samples/arm-none-eabi/reported.by
```diff
@@ -0,0 +1,310 @@
...
+# 
+# I tried to mimic the original configure options and environment settings with
+# crosstool-ng (CT_NG) as close as possible. I really analysed every option
+# which was not provided by CT-NG or which was added. The description in this
+# file explains each diversion from the original.
+# 
+# NOTE:   newlib-nano is currently not built!
+# 
+# The original uses the git/SVN versions of binutils-gdb, gcc and newlib. CT-NG
+# doesn't currently support using a git or SVN repository for the source
```
stilor: HEAD of ct-ng does support Git (see, for example, mingw.sh or binutils.sh). You'll need to modify the corresponding build script for gcc, though.

Also curious, why not update the existing arm-unknown-none-eabi? I think having both arm-unknown-none-eabi and arm-none-eabi would be confusing.

samples/arm-none-eabi/reported.by
```diff
@@ -0,0 +1,310 @@
...
+# location. In fact this is not really necessary, because the official packages
+# contain most of the required code already.
+# Only armv8-m is not available in the official packages. This core is very new
+# and even not available on GCC trunk, so I guess we can wait for the next GCC
+# release to add it.
+#
+# If someone really want to experiment with the mentioned core variant, feel
+# free to use them as custom location in CT-NG.
```
stilor: As you may have noticed, I have another pull request that tries to restore most samples to a passing build. As far as I understand, this sample will not build without setting up the sources checked out from SVN/Git as custom locations. I just tried your branch and it indeed failed at CC core pass-2 with:
````
[CFG  ]    Error: --with-multilib-list=armv6-m,armv7-m,armv7e-m,armv7-r not supported.
[CFG  ]    Makefile:4046: recipe for target 'configure-gcc' failed
````

Therefore, it will get reported by `ct-ng build-all` as a failure. I'd suggest you implement `CT_GetSVN` in `scripts/functions` (`CT_GetGit` is already available); add the required support in `config/*` (see `mingw.in` for an example) and use the required Git/SVN locations in crosstool.config, rather than requiring some manual steps.

Please, don't add more broken samples than we already have. When pushing larger changes that are not constrained to a single architecture or triplet, @bhundven asks to regression test on all available targets and it is quite annoying to deal with such breakages.

samples/arm-none-eabi/reported.by
jasmin-j: > HEAD of ct-ng does support Git (see, for example, mingw.sh or binutils.sh). You'll need to
> modify the corresponding build script for gcc, though.

For the gcc 4.9.3 version I used git, but for 5.3.0 all required things are already in the upstream packages. I decided to skip the armv8-m support (this would require the latest changes in the git repos) and to stick with the official packages for now.

> Also curious, why not update the existing arm-unknown-none-eabi? I think having both
> arm-unknown-none-eabi and arm-none-eabi would be confusing.

When you look to the CT-NG options, you will see the difference. My goal was to provide a configuration which is as close as possible to the ARM embedded gcc version (arm-none-eabi). Maybe it would be better to rename my sample to arm-none-eabi-embedded, so it is clear what it does.

stilor: On 03/21/2016 05:53 PM, Jasmin Jessich wrote:
>
> In samples/arm-none-eabi/reported.by 
> <https://github.com/crosstool-ng/crosstool-ng/pull/372#discussion_r56923001>:
>
> > +#
> > +# This configuration will build an arm-none-eabi compiler to be used for
> > +# embedded development. The original build script can be found here:
> > +#    https://launchpad.net/gcc-arm-embedded
> > +# To see the original build script, you have to download the installer:
> > +#    https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major
> > +#
> > +# I tried to mimic the original configure options and environment settings with
> > +# crosstool-ng (CT_NG) as close as possible. I really analysed every option
> > +# which was not provided by CT-NG or which was added. The description in this
> > +# file explains each diversion from the original.
> > +#
> > +# NOTE:   newlib-nano is currently not built!
> > +#
> > +# The original uses the git/SVN versions of binutils-gdb, gcc and newlib. CT-NG
> > +# doesn't currently support using a git or SVN repository for the source
>
>     HEAD of ct-ng does support Git (see, for example, mingw.sh or
>     binutils.sh). You'll need to
>     modify the corresponding build script for gcc, though.
>
> For the gcc 4.9.3 version I used git, but for 5.3.0 all required 
> things are already in the upstream packages. I decided to skip the 
> armv8-m support (this would require the latest changes in the git 
> repos) and to stick with the official packages for now.
>
By upstream, you mean *mainline* GCC 5.3.0? GCC failed to configure for 
me with the multilib list you have in this sample (I merged your branch 
locally to try it out, including #371) with an error message that this 
multilib list is not supported.

Can you check if you have something in your environment that you forgot 
to add to this branch? Perhaps, add this new sample to .travis.yml and 
see if Travis CI can build it.
>
>     Also curious, why not update the existing arm-unknown-none-eabi? I
>     think having both
>     arm-unknown-none-eabi and arm-none-eabi would be confusing.
>
> When you look to the CT-NG options, you will see the difference. My 
> goal was to provide a configuration which is as close as possible to 
> the ARM embedded gcc version (arm-none-eabi). Maybe it would be better 
> to rename my sample to arm-none-eabi-embedded, so it is clear what it 
> does.
>
Of course I see that these samples are currently different; the question 
really was, is there anything in the *old* configuration that the *new* 
configuration does not provide?

And, when choosing the name for the sample, make sure that 'ct-ng 
saveconfig' agrees with your naming :)

Regards,
Alexey.

samples/arm-none-eabi/reported.by
jasmin-j: > As far as I understand, this sample will not build without setting up the sources checked out
> from SVN/Git as custom locations. I just tried your branch and it indeed failed at CC core
> pass-2 with ...

The sample doesn't need any SVN/Git version from any package. But I forgot to add the patches to enable the multilib support for armv6-m, armv7-m, armv7e-m and armv7-r.
I just added the missing parts.

> Please, don't add more broken samples than we already have.

This was ABSOLUTELY NOT my intention! I simply forgot the required patches :(
THX for your review!

stilor: > This was ABSOLUTELY NOT my intention! I simply forgot the required 
> patches :(
> THX for your review!
>
Ah, that explains the failure :) Thanks for fixing it!
Will try it out again tonight when I get home.

Regards,
Alexey.

samples/arm-none-eabi/reported.by
jasmin-j: > the question really was, is there anything in the *old* configuration that the *new* configuration
> does not provide?

It provides a compiler as close as possible to the original from ARM for embedded development (arm-none-eabi). It provides special settings for newlib and it builds for the embedded processor types (armv?-m).
Let me ask the question in the other direction. What was the intention for the "arm-unknown-none-eabi" ?
There is no further explanation within the reported.by file to explain it. So it is really hard to guess from the sample parameters itself.

> And, when choosing the name for the sample, make sure that 'ct-ng saveconfig' agrees with
> your naming :)

I tried this already, but I had no luck. I would like to name the tools exactly like ARM did it for their compiler (arm-none-eabi). But I would like to have the name of the sample "arm-none-eabi-embedded". I don't see how I need to set the options to convince 'ct-ng saveconfig' to do what I like. That's the reason, why I called it simple "arm-none-eabi".
Any hints?

And I plan another sample for mingw, where the compiler shall be named also "arm-none-eabi", but the sample arm-none-eabi-embedded-w64-mingw32 (or similar).

jasmin-j: > Can you check if you have something in your environment that you forgot to add to this
> branch?

Your guess was really good ;)

> Perhaps, add this new sample to .travis.yml and see if Travis CI can build it.

I didn't this on purpose. The sample is new and Travis CI runs very long with the current samples already. But I agree with you, then I would have seen my mistake.
Let's wait until it is merged to Trunk with that.

stilor: On 03/21/2016 06:34 PM, Jasmin Jessich wrote:
>
> In samples/arm-none-eabi/reported.by 
> <https://github.com/crosstool-ng/crosstool-ng/pull/372#discussion_r56925719>:
>
> > +#
> > +# This configuration will build an arm-none-eabi compiler to be used for
> > +# embedded development. The original build script can be found here:
> > +#    https://launchpad.net/gcc-arm-embedded
> > +# To see the original build script, you have to download the installer:
> > +#    https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major
> > +#
> > +# I tried to mimic the original configure options and environment settings with
> > +# crosstool-ng (CT_NG) as close as possible. I really analysed every option
> > +# which was not provided by CT-NG or which was added. The description in this
> > +# file explains each diversion from the original.
> > +#
> > +# NOTE:   newlib-nano is currently not built!
> > +#
> > +# The original uses the git/SVN versions of binutils-gdb, gcc and newlib. CT-NG
> > +# doesn't currently support using a git or SVN repository for the source
>
>     the question really was, is there anything in the /old/
>     configuration that the /new/ configuration
>     does not provide?
>
> It provides a compiler as close as possible to the original from ARM 
> for embedded development (arm-none-eabi). It provides special settings 
> for newlib and it builds for the embedded processor types (armv?-m).
> Let me ask the question in the other direction. What was the intention 
> for the "arm-unknown-none-eabi" ?
>
I have no idea, which is why I think it would be good to replace the 
existing sample with a configuration that has a known purpose :)
We can ask the original submitter (Yann Morin), he's available sometimes 
at the IRC. I'll flag him about this pull request next time I see him 
online.

> There is no further explanation within the reported.by file to explain 
> it. So it is really hard to guess from the sample parameters itself.
>
>     And, when choosing the name for the sample, make sure that 'ct-ng
>     saveconfig' agrees with
>     your naming :)
>
> I tried this already, but I had no luck. I would like to name the 
> tools exactly like ARM did it for their compiler (arm-none-eabi). But 
> I would like to have the name of the sample "arm-none-eabi-embedded". 
> I don't see how I need to set the options to convince 'ct-ng 
> saveconfig' to do what I like. That's the reason, why I called it 
> simple "arm-none-eabi".
> Any hints?
>
'ct-ng saveconfig' follows pretty strict rules as to naming - 
"[host-tuple,]target-tuple". So, it all boils down to how your 
configuration gets named by CT_DoBuildTargetTuple in scripts/functions. 
Perhaps we can add support for a configurable suffix to sample name.

But I think it is redundant in this case, given that EABI already stands 
for "Embedded ABI" [see https://wiki.debian.org/ArmEabiPort].

All that, I guess, is one more reason to replace existing 
arm-unknown-none-eabi sample with this one.
>
> And I plan another sample for mingw, where the compiler shall be named 
> also "arm-none-eabi", but the sample 
> arm-none-eabi-embedded-w64-mingw32 (or similar).
>
That would be named as something 
'i686-w64-mingw32,arm-unknown-none-eabi' by saveconfig.

Regards,
Alexey.

stilor: On a second thought, I think you should just hijack the vendor string. It is currently used for that purpose in many samples already; just set it to the string of your liking.

But still, I think two arm-*-none-eabi samples is one sample too many :)

samples/arm-none-eabi/reported.by
stilor: Confirmed: built fine for me, too. Thanks again!
As a side note, I wonder why ARM GCC does not allow similar precise multilib specification for t-aprofile (which is a fixed set of arm7-a/arm7ve/arm8-a architectures).

jasmin-j: > Confirmed: built fine for me, too. Thanks again!

Fine!

> As a side note, I wonder why ARM GCC does not allow similar precise multilib specification
> for t-aprofile ...

His words:
"Note that arm/t-aprofile is a stand-alone make file fragment to be used only with itself.  We
do not specifically use the TM_MULTILIB_OPTION framework because this shorthand is
more pragmatic."
I guess the developer (Ramana Radhakrishnan) didn't saw it as a requirement, because you normally build a compiler not very frequently. Then it doesn't matter if you compile multilib for only one or more targets, because the build time will be not increased very much.
Or he simply was to lazy ;)

stilor:
@bhundven , @yann-morin-1998 : 
Should this sample:
(1) replace the existing "arm-unknown-none-eabi"?
(2) be added as a new sample?
I favor (1) - I don't see it as being much different, and we have a plethora of arm samples already. But if (2) is preferred, what should be the name for it? "arm-multilib-none-eabi"?

jasmin-j:
As I already wrote several times in this thread, this is intended to build most exactly the already existing arm-none-eabi gcc. All the ARM embedded developers are using exactly this name and the currently the package from ARM. That's why I choose exactly this name. Another (new) name for the arm embedded gcc would confuse a lot of developers and they even wouldn't it consider as a replacement for the ARM variant.

stilor:
Well, the questions to @bhundven and @yann-morin-1998 still stand, whether this sample should replace arm-unknown-eabi (sorry, made a typo in the previous comment) or placed alongside it. I think, replace.

I am ok with any name for the new sample, so long as 'ct-ng saveconfig' updates the sample in the proper directory (other ct-ng maintenance targets depend on this assumption, so breaking it will result in undue maintenance burden).

I've posted a review on your other PR (newlib updates) - let's pick that one up first. Unless Bryan or Yann respond otherwise, I'd suggest to update this PR to remove arm-unknown-eabi and replace it with the your new sample (since this new sample, as you claim, now follows the expected naming).

stilor:
Aside from the sample name... I am reluctant to add a huge patch for GCC until it is merged upstream, it is going to be painful to maintain.

Is this patch integrated upstream in GCC5/GCC6/Upcoming GCC7?

jasmin-j:
> Aside from the sample name...

As I often wrote already, this is the name given from ARM for the embedded variant of the compiler.

> I am reluctant to add a huge patch for GCC until it is merged upstream, it is going to be painful to
> maintain.

Well, it was necessary to add the multilib build option for ARM. To get it into GCC took nearly a year, even by ARM itself.

> Is this patch integrated upstream in GCC5/GCC6/Upcoming GCC7?

For GCC5 I don`t know, but for GCC6 I guess it is already in. I have seen it in GCC master last week.
Thomas (ARM) made a different variant of his original patch which does what my version did.
So yes, it is upstream available at least for GCC7.

I am currently on vacation and back in Vienna end of February. Then I can check CT-NG and how to build the arm-none-eabi variant with GCC6. My goal is still to have a CT-NG configuration which builds exactly the ARM embedded GCC, like the build script from ARM. Please have a look to #375, which is required for my approach.
If the ARM patch is not back ported to GCC6 now, I will provide a new one.

GCC5 we can forget. I will not put effort in older variants, which even don`t support all the CPUs.

I would suggest to close this pull request now and I will make a new one end of February.

PR #375, opened 2016-03-14
Added new gcc config option CC_GCC_CONFIG_TLS

jasmin-j:
Adding new tristate configuration for TLS (Thread Local Storage) to
add "--enable-tls" (y), "--disable-tls" (n) or nothing (m).

Signed-off-by: Jasmin Jessich jasmin@anw.at

---

This change is [Reviewable](https://reviewable.io/reviews/crosstool-ng/crosstool-ng/375)

### DIFF
--- a/config/cc/gcc.in.2
+++ b/config/cc/gcc.in.2
@@ -90,6 +90,25 @@ config CC_GCC_SYSTEM_ZLIB
       
       If unsure, say 'n'.
 
+config CC_GCC_CONFIG_TLS
+    tristate
+    prompt "Configure TLS (Thread Local Storage)"
+    default m
+    help
+      Specify that the target supports TLS (Thread Local Storage). Usually
+      configure can correctly determine if TLS is supported. In cases where
+      it guesses incorrectly, TLS can be explicitly enabled or disabled.
+      This can happen if the assembler supports TLS but the C library does
+      not, or if the assumptions made by the configure test are incorrect.
+
+       Option  | TLS use            | Associated ./configure switch
+      ---------+--------------------+--------------------------------
+         Y     | forcibly used      | --enable-tls
+         M     | auto               | (none, ./configure decides)
+         N     | forcibly not used  | --disable-tls
+      
+      If unsure, say 'M'.
+
 #-----------------------------------------------------------------------------
 # Optimisation features
 
--- a/scripts/build/cc/100-gcc.sh
+++ b/scripts/build/cc/100-gcc.sh
@@ -555,6 +555,12 @@ do_gcc_core_backend() {
         extra_config+=("--with-system-zlib")
     fi
 
+    case "${CT_CC_GCC_CONFIG_TLS}" in
+        y)  extra_config+=("--enable-tls");;
+        m)  ;;
+        "") extra_config+=("--disable-tls");;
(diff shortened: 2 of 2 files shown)


bhundven:
@jasmin-j is nothing(m) different then "--disable-tls"(n)?

jasmin-j:
> is nothing(m) different then "--disable-tls"(n)?

This is a copy of the help text:
Specify that the target supports TLS (Thread Local Storage). Usually configure can correctly determine if TLS is supported. In cases where it guesses incorrectly, TLS can be explicitly enabled or disabled.
This can happen if the assembler supports TLS but the C library does not, or if the assumptions made by the configure test are incorrect.

Yes, there is a difference. At least how I have understood the TLS option description in gcc/doc/install.texi. "--disable-tls"(n) would unconditionally disable it, "--enable-tls"(y) would unconditionally enable it and no TLS option (m) lets configure determine if it is supported or not.

josator:
Facing the same promblem.

jasmin-j:
It seems I need to rebase this. Hope I can do it on the weekend.

josator:
I am trying to disable TLS on gcc. However when I issue --disable-tls the macro HAVE_TLS is disabled, but the flag USE_EMUTLS is enabled. The code I need on libgomp is the one activated when both  macros are disabled.

jasmin-j:
Rebase done.

jasmin-j:
@josator
I checked "libgcc/configure.ac" in the current gcc git master branch and there USE_EMUTLS is set only, if $enable_tls is also "yes". This code is there since 2010 (8f0f72c0571a7ec64e9ad20bb6211fd44b5af7ee).
This means you can disable "USE_EMUTLS" in libgcc also with "--disable-tls".
But in libgomp there is no such additional code. They use there the result from "GCC_CHECK_EMUTLS" macro (config/tls.m4).
I guess you need to debug the exact behaviour of libgomp's configure and patch it there, or ask on the GCC mailing list how to disable "USE_EMUTLS" in libgomp.
The added USE_EMUTLS in libgomp at Dec 10 13:42:20 2014 (4b9d11e9bb503a98c652a9bd7f685e2f04403d65).
I am not sure if they thought about completely disabling it, like in libgcc.

josator:
@jasmin-j  Thanks for your reply, I did a dirty patch on both libgcc and libgomp configures to disable the macros. I will use your info / ask in the mail lists to find a better solution.

jasmin-j:
I just rebased this branch.
@stilor, please can you add it before you release 1.23.

stilor:
Can you describe how the libgomp issue (that you discussed above) is resolved?

jasmin-j:
> Can you describe how the libgomp issue (that you discussed above) is resolved?

I haven't it resolved at all because I didn't even had this issue when I was playing with the arm-none-eabi compiler.
This pull request adds the possibility to set/clear a compiler option which exists and which I need for the arm-none-eabi compiler.
If libgomp doesn't set "USE_EMUTLS" in a way @josator needs, this has nothing to do with this pull request which adds setting a compiler flag only.

stilor:
Fair enough, will merge. Just curious though, did you encounter a configuration where it was not determined correctly by `configure`?

Issue #374, opened 2016-03-14
CFLAGS_FOR_TARGET/CXX_FLAGS_FOR_TARGET for libstdc++ in baremetal build


jcmvbkbc:
Hello,

I need to build libstdc++ for baremetal targeted cross-compiler with specific compiler options. I tried to pass these options in CT_CC_GCC_ENABLE_CXX_FLAGS, it works partially: they don't apply to C source files in libstdc++/libsupc++, e.g. to libstdc++-v3/libsupc++/cp-demangle.c

Looking at the Buildroot I see that it sets CFLAGS_FOR_TARGET and CXX_FLAGS_FOR_TARGET and they appear in gcc options when libstdc++ is built. If I set CT_TARGET_CFLAGS in the crosstool-NG this doesn't happen.

For a moment I thought that it's due to the way final baremetal cross-compiler is built: do_gcc_core_backend() is used for that and it doesn't pass CFLAGS_FOR_TARGET, CXXFLAGS_FOR_TARGET or LDFLAGS_FOR_TARGET to gcc's configure script. I've changed that, and I can see changed gcc configure invocation in the build log, but flags still don't appear in the libstdc++ build.

Any clues?

jcmvbkbc:
OK, I've missed that I had CT_CC_GCC_ENABLE_TARGET_OPTSPACE=y in my crosstool-NG configuration and that passed --enable-target-optspace to gcc, making it use config/mt-ospace instead of config/mt-gnu, for which I backported a [fix.](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=c037df1be41f8daf4d581d7ffa4ec8cfa640bccf) So I've backported a [fix](https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=9bcf38cd9f382486b3823eb923b50e2e9a89cef7) for config/mt-ospace too and it all works now.
Will post a patch soon.

PR #376, opened 2016-03-15
Pass CFLAGS_FOR_TARGET/CXXFLAGS_FOR_TARGET for libstdc++ in baremetal build

jcmvbkbc:
Hi Bryan,

this series makes it possible to build libstdc++ for baremetal cross-compiler with flags specified in the CT_TARGET_CFLAGS. It backports two gcc fixes that preserve CFLAGS_FOR_TARGET and CXXFLAGS_FOR_TARGET from gcc-5.x to gcc-4.8.x and gcc-4.9.x and passes these environment variables to gcc configure script in do_gcc_core_backend() as it is done in do_gcc_backend().
It fixes issue #374.


bhundven:
Fantastic! This solves another issue I've been running into as well! Thanks for this!

PR #382, opened 2016-04-19
Add support for avr-libc-2.0.0

mbeards:
Introduces support for new avr-libc version 2.0.0, which adds support for gcc 5.


bhundven:
No signed-off-by, sample is not updated, and is a duplicate of #381. Closing.

PR #387, opened 2016-04-29
gettext: update to 0.19.7

jserv:
This change is [Reviewable](https://reviewable.io/reviews/crosstool-ng/crosstool-ng/387)

### DIFF
--- a/config/companion_libs/gettext.in
+++ b/config/companion_libs/gettext.in
@@ -6,9 +6,9 @@ choice
 # Don't remove next line
 # CT_INSERT_VERSION_BELOW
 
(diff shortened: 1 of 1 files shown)


bhundven:
This PR fails to build, please look into it.

jserv:
According to the build message of CI, I don't think the failure resulted from the upgrade of GNU gettext. Instead, the root cause is `do_isl_get`, where the [isl](http://isl.gforge.inria.fr/) was unable to access.

bhundven:
I have restarted the CI test for this PR. It seems that the travis-ci cache was hosed and needed to be cleared out.

PR #393, opened 2016-05-22
arch/sh: adding more SuperH variants

michaelhudak:
adding SH1, SH2, SH2E, SH2A, SH3E as additional SuperH targets

furthermore updating hyperlink

tested SH2, worked  successfully _without_ newlib

Source: [GCC Reference](https://gcc.gnu.org/onlinedocs/gcc/SH-Options.html)

### DIFF
--- a/config/arch/sh.in.2
+++ b/config/arch/sh.in.2
@@ -4,9 +4,29 @@ choice
     bool
     prompt "Variant"
 
+config ARCH_SH_SH1
+    bool
+    prompt "sh1"
+
+config ARCH_SH_SH2
+    bool
+    prompt "sh2"
+
+config ARCH_SH_SH2E
+    bool
+    prompt "sh2e"
+
+config ARCH_SH_SH2A
+    bool
+    prompt "sh2a"
+
 config ARCH_SH_SH3
     bool
     prompt "sh3"
+
+config ARCH_SH_SH3E
+    bool
+    prompt "sh3e"
 
 config ARCH_SH_SH4
     bool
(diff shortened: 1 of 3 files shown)


fenugrec:
Cool, but is it really necessary to specify a variant ? It seems newlib can't build if there's a variant specified.

config/arch/sh.in.2
```diff
@@ -4,10 +4,30 @@ choice
     bool
     prompt "Variant"
 
+config ARCH_SH_SH1
+    bool
+    prompt "sh1"
+
+config ARCH_SH_SH2
```
stilor: Curious which libc variant does your configuration use? Baremetal?
I am asking because this change may affect the multilib PR I posted - gcc for SuperH allows to select the set of MCUs for the set of multilibs to be built - and that requires some extra steps for uClibc (see https://github.com/stilor/crosstool-ng/blob/e42e353a64b518895bf9909b9f966544ead5f255/scripts/build/arch/sh.sh for example).
If using baremetal, probably a good idea would be to save your configuration as a sample - we don't have sh*-*-none samples.

stilor:
No response from originator. I won't merge this without a configuration that exercises this code.

fenugrec:
@stilor you mentioned one of your PRs , at the time I think I gave up and used a prebuilt toolchain, but I'm curious to know if it's now possible to build gcc-sh for all the SH variants ?

stilor:
Do you mean a multilib toolchain for all the SH variants? Well, I'll need to retry but last time I checked it didn't go well with a configuration that includes variants of both endianness. Also, I am not sure if uClibc-ng will build for sh2.

What is that prebuilt toolchain you mentioned and which targets does it support?

fenugrec:
well, my targets of interest were all big-endian (SH2, SH2E mostly) but even just that didn't go well. The toolchain I'm using, albeit dated now, supports many targets -- https://gcc-renesas.com/sh/download-latest-toolchains/

PR #395, opened 2016-05-24
comptools/make: Fix CT_COMP_TOOLS_make_gmake option

bwalle:
Commit 6f8e89cb5ca061e899bf3feaaf3fecf30d366c3e broke that option.
Since ${make} points to /usr/bin/make, making the symlink from gmake
to /usr/bin/make is obviously the wrong decision. gmake should link to
our (old-versioned) self-built make.

Signed-off-by: Bernhard Walle bernhard@bwalle.de

### DIFF
--- a/scripts/build/companion_tools/050-make.sh
+++ b/scripts/build/companion_tools/050-make.sh
@@ -23,7 +23,7 @@ do_companion_tools_make_build() {
     CT_DoExecLog ALL ${make}
     CT_DoExecLog ALL ${make} install
     if [ "${CT_COMP_TOOLS_make_gmake}" = "y" ]; then
-        CT_DoExecLog ALL ln -sv ${make} "${CT_BUILDTOOLS_PREFIX_DIR}/bin/gmake"
+        CT_DoExecLog ALL ln -sv make "${CT_BUILDTOOLS_PREFIX_DIR}/bin/gmake"
     fi
     CT_Popd
     CT_EndStep


scripts/build/companion_tools/050-make.sh
```diff
@@ -23,7 +23,7 @@ do_companion_tools_make_build() {
     CT_DoExecLog ALL ${make}
     CT_DoExecLog ALL ${make} install
     if [ "${CT_COMP_TOOLS_make_gmake}" = "y" ]; then
-        CT_DoExecLog ALL ln -sv ${make} "${CT_BUILDTOOLS_PREFIX_DIR}/bin/gmake"
+        CT_DoExecLog ALL ln -sv make "${CT_BUILDTOOLS_PREFIX_DIR}/bin/gmake"
```
stilor: That's the right change, but it is only half of the fix: the ct-ng will still invoke /usr/bin/make down the road. It looks like https://github.com/crosstool-ng/crosstool-ng/pull/268/commits/6f8e89cb5ca061e899bf3feaaf3fecf30d366c3e by @bhundven went a bit too far: it replaced all "foo" invocations with "${foo}" for make/sed/awk, etc. This is right for the tools that we don't build and rely on the configure to detect the path (sed/awk/...); but for make, we should always be invoking the one in `.build/buildtools/bin`.

bwalle: It will not invoke /usr/bin/make since the directory ${CT_BUILDTOOLS_PREFIX_DIR}/bin is inside $PATH inside the ct-ng build. So it will invoke the make binary inside ${CT_BUILDTOOLS_PREFIX_DIR}/bin which is the correct one.

I tested the change and it worked for me.

stilor: I didn't say your change wasn't tested. What I said is that scripts executed **after** this one will still invoke wrong make. For example, look at `scripts/build/cc/100-gcc.sh`:
````
CT_DoExecLog CFG ${make} ${JOBSFLAGS} configure-libiberty
````
This will again invoke /usr/bin/make rather than the one in `buildtools`.

bhundven:
It seems like there should be more to "fix the problem".
The `crosstool-NG.sh.in` makes the `${CT_BUILDTOOLS_PREFIX_DIR}/bin` early. I think there should be a check after making that directory to symlink the system's `{g,}make` to the buildtools prefix bin directory. So the`paths.{mk,sh}` always points to the make binary, and if we build the companion gmake, then it overwrites the symlink to it's path, and the build uses that.

stilor:
Superseded by PR 479.

PR #404, opened 2016-06-11
cross-gdb: Enable TUI interface

lundmar:
This change enables TUI (text user interface) for cross-gdb. Since this
feature relies on ncurses we also make sure ncurses for host gets built.

### DIFF
--- a/scripts/build/companion_libs/220-ncurses.sh
+++ b/scripts/build/companion_libs/220-ncurses.sh
@@ -33,7 +33,7 @@ do_ncurses_for_build() {
           "--without-cxx-binding" \
           "--without-ada")
     do_ncurses_backend host="${CT_BUILD}" \
-                       destdir="${CT_BUILDTOOLS_PREFIX_DIR}" \
+                       prefix="${CT_BUILDTOOLS_PREFIX_DIR}" \
                        cflags="${CT_CFLAGS_FOR_BUILD}" \
                        ldflags="${CT_LDFLAGS_FOR_BUILD}" \
                        "${opts[@]}"
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -124,6 +124,7 @@ do_debug_gdb_build() {
             --with-build-sysroot="${CT_SYSROOT_DIR}"    \
             --with-sysroot="${CT_SYSROOT_DIR}"          \
             --disable-werror                            \
+            --enable-tui                                \
             "${cross_extra_config[@]}"                  \
             "${CT_GDB_CROSS_EXTRA_CONFIG_ARRAY[@]}"
 


scripts/build/companion_libs/220-ncurses.sh
```diff
@@ -49,7 +49,7 @@ do_ncurses_for_host() {
     # (i.e. in simple cross or native): ncurses may not be needed for
     # host, but we still need them on build to produce 'tic'.
     case "${CT_TOOLCHAIN_TYPE}" in
-        native|cross)   return 0;;
+        native)   return 0;;
```
stilor: In case of the 'cross' type, build==host, so `do_ncurses_for_build` should have built ncurses already. Why wasn't it usable? It seems it is configured in the same way (the only differences are extra directories in cflags/ldflags - but those shouldn't matter as ncurses does not depend on any other library previously installed in buildtools/host_complibs).

In any case, the comment above this check needs to be updated.

lundmar: I've lost my debug data but if I recall correctly, in case of cross gdb, CC="${CC_for_gdb}" and LD="${LD_for_gdb}" in scripts/build/debug/300-gdb.sh expands to include library search and include paths which points to the host sysroot. Since only the build sysroot include ncurses static it does not find it and so TUI has indirectly been disabled for a long time. I assumed host sysroot is the correct one to populate in case of building host tools. Or, we need to fix so that it points to the build sysroot.

stilor: But in case of 'cross' build, library/include paths for host point to the same sysroot as for build. E.g.:
````
CT_CFLAGS_FOR_BUILD='-O2 -g '
CT_CFLAGS_FOR_HOST='-O2 -g -pipe  -I/home/avn/work/ctng/crosstool-ng/.build/sh4-unknown-linux-uclibc/buildtools/include'
CT_LDFLAGS_FOR_BUILD=' '
CT_LDFLAGS_FOR_HOST='  -L/home/avn/work/ctng/crosstool-ng/.build/sh4-unknown-linux-uclibc/buildtools/lib'
````
As you see, it is the same directory where build step places the libraries/headers.

stilor: As a side note, we should probably add -I/-L options to FOR_BUILD variables too, but so far it hasn't bitten us :)

lundmar: With "native|cross", gdb fails to start with TUI:

$ ./arm-cortex_m4-eabi-gdb -tui
Cannot enable the TUI: error opening terminal [TERM=xterm]

With "native" only it works fine.

So there is some sort of difference here which breaks gdb. It seems ncurses related. Needs further investigation.

lundmar:
Ok, by further inspection it looks like the ncurses installation was broken because it used the "--with-install-prefix" instead of the "--prefix" option.

lundmar:
I've updated and tested the commit.

stilor:
The new diff makes me suspect something... Does GDB TUI work if you run 'ct-ng clean' after the build? It looks like the resulting GDB depends on the shared libraries that are installed into a temporary directory.

lundmar:
You are right - gdb TUI stops working after a clean. 

Since the resulting gdb is a static it made me realize and remember that the problem is likely that ncurses normally installs a zillion compiled terminfo binaries. Which one gets loaded depends on the value of TERM at startup of the ncurses interface. So now the question becomes, do we bundle the terminfo files to solve this issue or should we simply rely on the host system for those?

stilor:
Ok, so it's not the shared libraries that are the culprit (although for non-static toolchains, we need to install host companion libs, too, it seems - I just tried and cross GDB indeed uses host's libexpat at runtime, for example). My vote is to bundle everything we build for host (companion libs and their data + I think we should allow building companion tools for host too, not just for build). But I'll try to discuss it with @yann-morin-1998 and/or @bhundven next time I see them on IRC.

lundmar:
In case we want to avoid bloating the toolchain with all the terminfo files our best option might be to configure ncurses to embed the terminal configurations (fallbacks) for the most popular terminals.

According to ncurses doc this means using "--disable-database" in combination with eg "--with-fallbacks=linux,xterm,xterm-color,vt100".

Alternatively, it should also be possible to embed all the terminfo files in a single Berkeley DB file but that will pull in  dependency on the Berkeley DB library so I don't think that is a good option.

davydnorris:
Hi all,

I'm trying to build crosstool-NG for xtensa-lx106-elf using Cygwin64 and am hitting link problems in the final step because of missing references in TUI. Here's the final link line:

`[ALL  ]    x86_64-build_unknown-cygwin-gcc -g -O2   -static-libstdc++ -static-libgcc -Wl,--stack,12582912      -o gdb.exe gdb.o xtensa-tdep.o xtensa-config.o solib-svr4.o ser-base.o ser-unix.o ser-pipe.o ser-tcp.o  remote.o dcache.o tracepoint.o ax-general.o ax-gdb.o remote-fileio.o remote-notif.o ctf.o tracefile.o tracefile-tfile.o  cli-dump.o cli-decode.o cli-script.o cli-cmds.o cli-setshow.o cli-logging.o cli-interp.o cli-utils.o mi-out.o mi-console.o mi-cmds.o mi-cmd-catch.o mi-cmd-env.o mi-cmd-var.o mi-cmd-break.o mi-cmd-stack.o mi-cmd-file.o mi-cmd-disas.o mi-symbol-cmds.o mi-cmd-target.o mi-cmd-info.o mi-interp.o mi-main.o mi-parse.o mi-getopt.o tui-command.o tui-data.o tui-disasm.o tui-file.o tui-hooks.o tui-interp.o tui-io.o tui-layout.o tui-out.o tui-regs.o tui-source.o tui-stack.o tui-win.o tui-windata.o tui-wingeneral.o tui-winsource.o tui.o python.o py-arch.o py-auto-load.o py-block.o py-bpevent.o py-breakpoint.o py-cmd.o py-continueevent.o py-xmethods.o py-event.o py-evtregistry.o py-evts.o py-exitedevent.o py-finishbreakpoint.o py-frame.o py-framefilter.o py-function.o py-gdb-readline.o py-inferior.o py-infevents.o py-infthread.o py-lazy-string.o py-linetable.o py-newobjfileevent.o py-objfile.o py-param.o py-prettyprint.o py-progspace.o py-signalevent.o py-stopevent.o py-symbol.o py-symtab.o py-threadevent.o py-type.o py-unwind.o py-utils.o py-value.o py-varobj.o guile.o elfread.o stap-probe.o dtrace-probe.o posix-hdep.o posix-strerror.o c-exp.o cp-name-parser.o ada-exp.o jv-exp.o d-exp.o f-exp.o go-exp.o m2-exp.o p-exp.o version.o annotate.o addrmap.o auto-load.o auxv.o agent.o bfd-target.o blockframe.o breakpoint.o break-catch-sig.o break-catch-throw.o break-catch-syscall.o findvar.o regcache.o cleanups.o charset.o continuations.o corelow.o disasm.o dummy-frame.o dfp.o source.o value.o eval.o valops.o valarith.o valprint.o printcmd.o block.o symtab.o psymtab.o symfile.o symfile-debug.o symmisc.o linespec.o dictionary.o infcall.o infcmd.o infrun.o expprint.o environ.o stack.o thread.o exceptions.o extension.o filesystem.o filestuff.o inf-child.o interps.o minidebug.o main.o macrotab.o macrocmd.o macroexp.o macroscope.o mi-common.o event-loop.o event-top.o inf-loop.o completer.o gdbarch.o arch-utils.o gdbtypes.o gdb_bfd.o gdb_obstack.o osabi.o copying.o memattr.o mem-break.o target.o target-dcache.o parse.o language.o build-id.o buildsym.o findcmd.o std-regs.o signals.o exec.o reverse.o bcache.o objfiles.o observer.o minsyms.o maint.o demangle.o dbxread.o coffread.o coff-pe-read.o dwarf2read.o mipsread.o stabsread.o corefile.o dwarf2expr.o dwarf2loc.o dwarf2-frame.o dwarf2-frame-tailcall.o ada-lang.o c-lang.o d-lang.o f-lang.o objc-lang.o ada-tasks.o ada-varobj.o c-varobj.o ui-out.o cli-out.o varobj.o vec.o go-lang.o go-valprint.o go-typeprint.o jv-lang.o jv-valprint.o jv-typeprint.o jv-varobj.o m2-lang.o opencl-lang.o p-lang.o p-typeprint.o p-valprint.o sentinel-frame.o complaints.o typeprint.o ada-typeprint.o c-typeprint.o f-typeprint.o m2-typeprint.o ada-valprint.o c-valprint.o cp-valprint.o d-valprint.o f-valprint.o m2-valprint.o serial.o mdebugread.o top.o utils.o ui-file.o user-regs.o frame.o frame-unwind.o doublest.o frame-base.o inline-frame.o gnu-v2-abi.o gnu-v3-abi.o cp-abi.o cp-support.o cp-namespace.o reggroups.o trad-frame.o tramp-frame.o solib.o solib-target.o prologue-value.o memory-map.o memrange.o xml-support.o xml-syscall.o xml-utils.o target-descriptions.o target-memory.o xml-tdesc.o xml-builtin.o inferior.o osdata.o gdb_usleep.o record.o record-full.o gcore.o gdb_vecs.o jit.o progspace.o skip.o probe.o common-utils.o buffer.o ptid.o gdb-dlfcn.o common-agent.o format.o registry.o btrace.o record-btrace.o waitstatus.o print-utils.o rsp-low.o errors.o common-debug.o debug.o common-exceptions.o btrace-common.o fileio.o compile.o compile-c-symbols.o compile-c-types.o compile-object-load.o compile-object-run.o compile-loc2c.o compile-c-support.o inflow.o    init.o       ../readline/libreadline.a ../opcodes/libopcodes.a ../bfd/libbfd.a -L./../zlib -lz  ../libiberty/libiberty.a ../libdecnumber/libdecnumber.a    -lncurses -lm   -liconv  -ldl -lpython2.7 /home/dnorris/crosstool-NG/.build/xtensa-lx106-elf/buildtools/lib/libexpat.a -llzma   ../libiberty/libiberty.a -luser32 build-gnulib/import/libgnu.a
[ALL  ]    tui-win.o:tui-win.c:(.rdata$.refptr.stdscr[.refptr.stdscr]+0x0): undefined reference to `stdscr'
[ALL  ]    tui-win.o:tui-win.c:(.rdata$.refptr.curscr[.refptr.curscr]+0x0): undefined reference to `curscr'
[ALL  ]    tui-win.o:tui-win.c:(.rdata$.refptr.acs_map[.refptr.acs_map]+0x0): undefined reference to `acs_map'
[ALL  ]    tui.o:tui.c:(.rdata$.refptr.COLS[.refptr.COLS]+0x0): undefined reference to `COLS'
[ALL  ]    tui.o:tui.c:(.rdata$.refptr.LINES[.refptr.LINES]+0x0): undefined reference to `LINES'
[ERROR]    collect2: error: ld returned 1 exit status
[ERROR]    make[3]: *** [Makefile:1391: gdb.exe] Error 1
[ALL  ]    make[3]: Leaving directory '/home/dnorris/crosstool-NG/.build/xtensa-lx106-elf/build/build-gdb-cross/gdb'
[ERROR]    make[2]: *** [Makefile:8759: all-gdb] Error 2
[ALL  ]    make[2]: Leaving directory '/home/dnorris/crosstool-NG/.build/xtensa-lx106-elf/build/build-gdb-cross'
[ERROR]    make[1]: *** [Makefile:849: all] Error 2
[ALL  ]    make[1]: Leaving directory '/home/dnorris/crosstool-NG/.build/xtensa-lx106-elf/build/build-gdb-cross'
[ERROR] 
[ERROR]  >>
[ERROR]  >>  Build failed in step 'Installing cross-gdb'
[ERROR]  >>
[ERROR]  >>  Error happened in: CT_DoExecLog[scripts/functions@257]
[ERROR]  >>        called from: do_debug_gdb_build[scripts/build/debug/300-gdb.sh@120]
[ERROR]  >>        called from: do_debug[scripts/build/debug.sh@35]
[ERROR]  >>        called from: main[scripts/crosstool-NG.sh@646]
[ERROR]  >>
[ERROR]  >>  For more info on this error, look at the file: 'build.log'
[ERROR]  >>  There is a list of known issues, some with workarounds, in:
[ERROR]  >>      '/esp8266/share/doc/crosstool-ng/crosstool-ng-1.22.0-56-g7d844e7/B - Known issues.txt'
[ERROR] 
[ERROR]  (elapsed: 76:41.93)
`

I suspect it's a static vs. dynamic link issue but it's got me stumped. I've had to add `--disable-tui` to the options for now

PR #428, opened 2016-08-25
Binutils 2.27

tjohann:
Hi@Briayn,

i added base support for binutils-2.27 (i could build a toolchain and then build linux-4.6.7 for armhf) ...

Add base support for binutils-2.27

Cheers
  Thorsten


bhundven:
This build fails. Also, you're adding more then binutils, so this PR is ambiguous. I'm closing this. Please only add binutils. You also need to port forward patches and add any new patches that may be needed.

PR #429, opened 2016-08-26
Update x86.sh

pedronavf:
Added additional x86 architectures, like core2, that also map to i686. Otherwise, when setting the x86 gcc to build optimized for core2 you get i585.


bhundven:
Thanks!

PR #441, opened 2016-09-07
duma: add SHARED_LIBS dependency for DUMA shared library option

kozyilmaz:
Add SHARED_LIBRARY dependency to DUMA shared library option, partially fixes #440

PR #442, opened 2016-09-07
debug/duma: disable shared libs if necessary

KirillSmirnov:
Propagate CT_SHARED_LIBS option to D.U.M.A. debug tool.

Signed-off-by: Kirill Smirnov <kirill.k.smirnov@gmail.com>


KirillSmirnov:
Whoops, almost the same patch is already PR'ed #441.
Revoking this one.

PR #443, opened 2016-09-09
debug/gdb: properly link with expat

KirillSmirnov:
This patch fixes libexpat detection for gdb-native and
gdb-cross static builds.

For gdb-native build configure should not touch system
/usr/{lib,include} directories while looking for libexpat.
To fix this we pass --without-libexpat-prefix flag
to configure script.

For gdb-cross build configure is allowed to investigate
system /usr/{lib,include} directories, but it does not
hurt to disable this behavior. In this case configure
falls back to -lexpat, which works as expected.

For more info:
http://marc.info/?l=gnulib-bug&m=129660262901148&w=2

Signed-off-by: Kirill Smirnov kirill.k.smirnov@gmail.com

### DIFF
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -68,6 +68,10 @@ do_debug_gdb_build() {
         cd "${CT_BUILD_DIR}/build-gdb-cross"
 
         cross_extra_config=("${extra_config[@]}")
+
+        # For gdb-cross this combination of flags forces
+        # gdb configure to fall back to default '-lexpat' flag
+        # which is acceptable.
         cross_extra_config+=("--with-expat")
         cross_extra_config+=("--without-libexpat-prefix")
 
@@ -169,6 +173,16 @@ do_debug_gdb_build() {
             native_extra_config+=("--with-curses")
         fi
 
+        # Target libexpat resides in sysroot and does not have
+        # any dependencies, so just passing '-lexpat' to gcc is enough.
+        #
+        # By default gdb configure looks for expat in '$prefix/lib'
+        # directory. In our case '$prefix/lib' resolves to '/usr/lib'
+        # where libexpat for build platform lives, which is
+        # unacceptable for cross-compiling.
+        #
+        # To prevent this '--without-libexpat-prefix' flag must be passed.
+        # Thus configure falls back to '-lexpat', which is exactly what we want.
         native_extra_config+=("--with-expat")
         native_extra_config+=("--without-libexpat-prefix")
 
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -69,11 +69,8 @@ do_debug_gdb_build() {
 
         cross_extra_config=("${extra_config[@]}")
         cross_extra_config+=("--with-expat")
-        # NOTE: DO NOT USE --with-libexpat-prefix (until GDB configure is smarter)!!!
-        # It conflicts with a static build: GDB's configure script will find the shared
-        # version of expat and will attempt to link that, despite the -static flag.
-        # The link will fail, and configure will abort with "expat missing or unusable"
-        # message.
+        cross_extra_config+=("--without-libexpat-prefix")
+
         case "${CT_THREADS}" in
             none)   cross_extra_config+=("--disable-threads");;
             *)      cross_extra_config+=("--enable-threads");;
@@ -173,11 +170,7 @@ do_debug_gdb_build() {
         fi
 
         native_extra_config+=("--with-expat")
-        # NOTE: DO NOT USE --with-libexpat-prefix (until GDB configure is smarter)!!!
-        # It conflicts with a static build: GDB's configure script will find the shared
-        # version of expat and will attempt to link that, despite the -static flag.
-        # The link will fail, and configure will abort with "expat missing or unusable"
-        # message.
+        native_extra_config+=("--without-libexpat-prefix")
 
         CT_DoLog EXTRA "Configuring native gdb"
 
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -72,6 +72,12 @@ do_debug_gdb_build() {
         # For gdb-cross this combination of flags forces
         # gdb configure to fall back to default '-lexpat' flag
         # which is acceptable.
+        #
+        # NOTE: DO NOT USE --with-libexpat-prefix (until GDB configure is smarter)!!!
+        # It conflicts with a static build: GDB's configure script will find the shared
+        # version of expat and will attempt to link that, despite the -static flag.
+        # The link will fail, and configure will abort with "expat missing or unusable"
+        # message.
         cross_extra_config+=("--with-expat")
         cross_extra_config+=("--without-libexpat-prefix")
 
@@ -183,6 +189,12 @@ do_debug_gdb_build() {
         #
         # To prevent this '--without-libexpat-prefix' flag must be passed.
         # Thus configure falls back to '-lexpat', which is exactly what we want.
+        #
+        # NOTE: DO NOT USE --with-libexpat-prefix (until GDB configure is smarter)!!!
+        # It conflicts with a static build: GDB's configure script will find the shared
+        # version of expat and will attempt to link that, despite the -static flag.
+        # The link will fail, and configure will abort with "expat missing or unusable"
+        # message.
         native_extra_config+=("--with-expat")
         native_extra_config+=("--without-libexpat-prefix")
 
--- a/scripts/build/companion_libs/200-libelf.sh
+++ b/scripts/build/companion_libs/200-libelf.sh
@@ -69,13 +69,24 @@ if [ "${CT_LIBELF_TARGET}" = "y" ]; then
 
 do_libelf_for_target() {
     local -a libelf_opts
+    local prefix
 
     CT_DoStep INFO "Installing libelf for the target"
     CT_mkdir_pushd "${CT_BUILD_DIR}/build-libelf-target-${CT_TARGET}"
 
+    case "${CT_TARGET}" in
+        *-*-mingw*)
+            prefix="/mingw"
+            ;;
+        *)
+            prefix="/usr"
+            ;;
+    esac
+
     libelf_opts+=( "destdir=${CT_SYSROOT_DIR}" )
     libelf_opts+=( "host=${CT_TARGET}" )
-    libelf_opts+=( "prefix=/usr" )
+
+    libelf_opts+=( "prefix=${prefix}" )
     libelf_opts+=( "shared=y" )
     do_libelf_backend "${libelf_opts[@]}"
 
--- a/scripts/build/companion_libs/210-expat.sh
+++ b/scripts/build/companion_libs/210-expat.sh
@@ -43,12 +43,21 @@ fi
 if [ "${CT_EXPAT_TARGET}" = "y" ]; then
 do_expat_for_target() {
     local -a expat_opts
+    local prefix
 
     CT_DoStep INFO "Installing expat for target"
     CT_mkdir_pushd "${CT_BUILD_DIR}/build-expat-target-${CT_TARGET}"
 
     expat_opts+=( "host=${CT_TARGET}" )
-    expat_opts+=( "prefix=/usr" )
+    case "${CT_TARGET}" in
+        *-*-mingw*)
+            prefix="/mingw"
+            ;;
+        *)
+            prefix="/usr"
+            ;;
+    esac
+    expat_opts+=( "prefix=${prefix}" )
     expat_opts+=( "destdir=${CT_SYSROOT_DIR}" )
     expat_opts+=( "static_build=y" )
 
--- a/scripts/build/companion_libs/220-ncurses.sh
+++ b/scripts/build/companion_libs/220-ncurses.sh
@@ -72,13 +72,23 @@ fi
 
 if [ "${CT_NCURSES_TARGET}" = "y" ]; then
 do_ncurses_for_target() {
+    local prefix
+
     CT_DoStep INFO "Installing ncurses for target"
     CT_mkdir_pushd "${CT_BUILD_DIR}/build-ncurses-target-${CT_TARGET}"
     opts=("--without-sysmouse")
     [ "${CT_CC_LANG_CXX}" = "y" ] || opts+=("--without-cxx" "--without-cxx-binding")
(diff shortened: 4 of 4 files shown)


KirillSmirnov:
This fixes libexpat detection for static builds for me.

scripts/build/debug/300-gdb.sh
```diff
@@ -69,11 +69,8 @@ do_debug_gdb_build() {
 
         cross_extra_config=("${extra_config[@]}")
         cross_extra_config+=("--with-expat")
-        # NOTE: DO NOT USE --with-libexpat-prefix (until GDB configure is smarter)!!!
```
stilor: Isn't absence of `--with-libexpat-prefix` equivalent to `--without-libexpat-prefix`? What is the change in behaviour between these two configurations?

Even if this change is needed, I suggest to keep the comment (since your `--without-libexpat-prefix` only reinforces what this comment states - that `--with-libexpat-prefix` is harmful).

KirillSmirnov: > Isn't absence of --with-libexpat-prefix equivalent to --without-libexpat-prefix?
No, they are not equivalent at all! This is the point of the commit and PR.


> What is the change in behaviour between these two configurations?
--with-expat-prefix=<folder> just adds an additional folder to search for expat, it does not prevent peeking into $prefix/lib
--without-expat-prefix forbids looking to $prefix/lib.


> Even if this change is needed, I suggest to keep the comment (since your --without-libexpat-prefix only reinforces what this comment states - that --with-libexpat-prefix is harmful).
I think this comment is misleading.
GDB configure (more precisely, AC_LIB_HAVE_LINKFLAGS) is smart, but extremely "unfriendly": it is not trivial to come to "--without-expat-prefix" to achieve our goal. The real problem is that AC_LIB_HAVE_LINKFLAGS is not properly documented for cross-compiling with DESTDIR.

KirillSmirnov: > What is the change in behaviour between these two configurations?

Sorry, in prev comment I forgot to describe the main case:
If we do not pass "--without-expat-prefix" nor "--with-expat-prefix" flag, configure will look for expat in $prefix/lib (this is correct behavior), and in our case $prefix=/usr/lib, and voila - we got incorrect libexpat.so

bhundven:
Since most of the samples in the .travis.yml passed without this change, could we add a sample that fails without this change?

bhundven:
Also please add the new sample in a separate PR so we see it fail.

KirillSmirnov:
> Since most of the samples in the .travis.yml passed without this change, could we add a sample that fails without this change?

To trigger this bug we have to have libexpat.so in /usr/lib, which is common for ordinary multilib desktop or 32-bit OS. Is that true for travis-ci test server?

bhundven:
Oh. Hrm, https://github.com/travis-ci/travis-ci/issues/986 ... lame.

KirillSmirnov:
> So, am I right that the test case is to place a bogus libexpat.so into host's /usr/lib?

Yes, I think so.

> I am still a bit surprised that it looks at $prefix/lib rather than at ${CT_SYSROOT_DIR}$prefix/lib - after all, we do pass ${CT_SYSROOT_DIR} as the build sysroot. 

GDB uses AC_LIB_HAVE_LINKFLAGS routine from gnulib. $prefix/lib is a legitimate place to search for a lib unless we build with DESTDIR (and if one uses DESTDIR, he must pass --without-<lib>-prefix flag configure not to search if $prefix/lib). OTOH AC_LIB_HAVE_LINKFLAGS does not search in build sysroot because it is too smart and lazy :-) It knows that linker knows about sysroot and does not need additional hints.

            dnl Didn't find the library; assume it is in the system directories
            dnl known to the linker and runtime loader. <...>

> I'd still mention why --with-libexpat-prefix= does not work - old comment explained how that breaks static builds - we've already had that done once.

Hmm... I tried --with-libexpat-prefix with varions combinations with static/shared expat and linux/mingw target - it all worked for me. How to reproduce the bug with --with-libexpat-prefix ?

stilor:
Wait, I think DESTDIR is only used during the 'make install' step to place the build artifacts into a directory that's going to be used as root at runtime. I don't think it is used during configure/build steps; typical build procedure is `./configure; make; make install DESTDIR=...`

See here, for example: https://www.gnu.org/prep/standards/html_node/DESTDIR.html

The bug with --with-libexpat-prefix was seen when building a static toolchain (note - not the toolchain which does not support shared libraries, but rather a toolchain that is statically linked itself) - in that case, all cross binaries were built with --static but gdb's configure still tried to link with `<prefix>/lib/libexpat.so` explicitly rather than adding `-L<path>/lib -lexpat`. Static link with .so failed, obviously, aborting the configure.

It does seem that the configure indeed looks into $prefix/lib. What bothers me is, how does this work then with other libraries? For example, gdb's configure also checks for libiconv - would a bogus /usr/lib/libiconv.so result in the same failure?

KirillSmirnov:
> Wait, I think DESTDIR is only used during the 'make install' step to place the build artifacts into a directory that's going to be used as root at runtime. I don't think it is used during configure/build steps; typical build procedure is ./configure; make; make install DESTDIR=...

That's right. The problem is that configure has no idea what DESTDIR will be, since it is passed to 'make install' only. Thus, if DESTDIR != '', we must somehow tell configure (precisely, AC_LIB_HAVE_LINKFLAGS) not to inspect $prefix/lib. That's constitutes the necessity of --without-libexpat-prefix flag.

> What bothers me is, how does this work then with other libraries? For example, gdb's configure also checks for libiconv - would a bogus /usr/lib/libiconv.so result in the same failure?

Other libraries are disabled or do not use AC_LIB_HAVE_LINKFLAGS. For example, there is a dedicated macro AM_ICONV for libiconv.

stilor:
On 09/12/2016 10:05 AM, Kirill Smirnov wrote:
>
>     Wait, I think DESTDIR is only used during the 'make install' step
>     to place the build artifacts into a directory that's going to be
>     used as root at runtime. I don't think it is used during
>     configure/build steps; typical build procedure is ./configure;
>     make; make install DESTDIR=...
>
> That's right. The problem is that configure has no idea what DESTDIR 
> will be, since it is passed to 'make install' only. Thus, if DESTDIR 
> != '', we must somehow tell configure (precisely, 
> AC_LIB_HAVE_LINKFLAGS) not to inspect $prefix/lib. That's constitutes 
> the necessity of --without-libexpat-prefix flag.
>
>     What bothers me is, how does this work then with other libraries?
>     For example, gdb's configure also checks for libiconv - would a
>     bogus /usr/lib/libiconv.so result in the same failure?
>
> Other libraries are disabled or do not use AC_LIB_HAVE_LINKFLAGS. For 
> example, there is a dedicated macro AM_ICONV for libiconv.
>
I see. It looks like AC_LIB_HAVE_LINKFLAGS strategy is ill-suited for 
cross-compilation then. Indeed, I googled 'AC_LIB_HAVE_LINKFLAGS 
cross-compile' and there are several reports that AC_LIB_HAVE_LINKFLAGS 
is broken for cross-compilation, with different projects replacing it 
with AC_CHECK_LIB or other mechanisms such as pkgconfig. Here is an 
example: http://marc.info/?l=gnulib-bug&m=129653491822599

Makes me wonder why GDB (which is often cross-compiled) still uses this 
macro. But we'll have to work around whatever upstream packages give us :(

Regards,
Alexey.

KirillSmirnov:
This macro is not broken, rather misused. The next message on the thread (http://marc.info/?l=gnulib-bug&m=129660262901148&w=2) explains the details.

stilor:
On 09/12/2016 10:29 AM, Kirill Smirnov wrote:
>
> This macro is not broken, rather misused. The next message on the 
> thread (http://marc.info/?l=gnulib-bug&m=129660262901148&w=2) explains 
> the details.
>
Ok, understood, thanks!
But a quick grep through GDB's configure shows at least a few more 
instances of AC_LIB_HAVE_LINKFLAGS:

````
gdb/configure.ac:  AC_LIB_HAVE_LINKFLAGS([expat], [], [#include "expat.h"],
gdb/configure.ac:  AC_LIB_HAVE_LINKFLAGS([ipt], [], [#include 
"intel-pt.h"], [pt_insn_alloc_decoder (0);])
gdb/configure.ac:    AC_LIB_HAVE_LINKFLAGS([lzma], [], [#include "lzma.h"],
gdb/configure.ac:  AC_LIB_HAVE_LINKFLAGS([babeltrace], [babeltrace-ctf],
````

And they do come with their own --without-*-prefix flags:

````
   --without-libexpat-prefix     don't search for libexpat in includedir 
and libdir
   --without-libipt-prefix     don't search for libipt in includedir and 
libdir
   --without-liblzma-prefix     don't search for liblzma in includedir 
and libdir
   --without-libbabeltrace-prefix     don't search for libbabeltrace in 
includedir and libdir
````

Shouldn't we be supplying these, too?

KirillSmirnov:
We do not use these libraries, do we? Adding respective --without-XXX-prefix would be just a waste of characters :-)

KirillSmirnov:
Alexey, I try to reproduce a bug with static toolchain and --with-libexpat-prefix, but every time fail. Everything works smoothly, static lib is successfully picked up and passed to gcc. Both for cross and native.

[ALL  ]    checking whether to use expat... yes
[ALL  ]    checking for libexpat... yes
[ALL  ]    checking how to link with libexpat... /home/lich/pkg/repos/crosstool-ng/.build/arm-unknown-linux-uclibcgnueabi/buildtools/lib/libexpat.a

I did:
1) checkout fresh crosstool-ng master.
2) add lines:
cross_extra_config+=("--with-libexpat-prefix=${CT_HOST_COMPLIBS_DIR}")
native_extra_config+=("--with-libexpat-prefix=${CT_SYSROOT_DIR}/usr")
(or native_extra_config+=("--with-libexpat-prefix=${CT_SYSROOT_DIR}/mingw") variant)

to their respective locations in scripts/debug/300-gdb.sh
3) select config:

CT_LOCAL_TARBALLS_DIR="${HOME}/src"
CT_SAVE_TARBALLS=y
CT_LOG_EXTRA=y
CT_ARCH_CPU="arm926ej-s"
CT_ARCH_FLOAT_SW=y
CT_ARCH_arm=y
CT_STATIC_TOOLCHAIN=y
CT_KERNEL_linux=y
CT_LIBC_uClibc=y
CT_CC_LANG_CXX=y
CT_DEBUG_gdb=y
CT_GDB_NATIVE=y
CT_GDB_NATIVE_STATIC=y


I still believe that --without-libexpat-prefix is a better (because of sysroot) solution than --with-libexpat-prefix=<sysroot>/{usr|mingw}, just want to be sure that --with-libexpat-prefix really fails.

bhundven:
This branch needs to be updated to be merge-able. @stilor Is it ok to merge this, to fix the linking issue with expat or is there still work to be done with this change?

KirillSmirnov:
@stilor this PR obsoletes your PR #275 too. It is OK?

stilor:
Re adding the rest of the `--withoug-XXX-prefix` options: ok, as long as gdb's configure does not autodetect them as present (e.g. due to host header presence) and then fail to link (due to incompatible library in /usr/lib).

Ok to merge and obsolete PR #275. I still think it would be better to recreate the environment in the way mingw expects it, but it is pervasive change and I don't have much time to test mingw stuff right now.

KirillSmirnov:
ping?

stilor:
I checked and the issue with --with-libexpat-prefix is still there, so please restore the comments. To reproduce: add back the line:
````
cross_extra_config+=("--with-libexpat-prefix=${CT_HOST_COMPLIBS_DIR}")
````
where you removed the comment, then configure any sample (I used mips-unknown-elf) and select 'Build static cross GDB' option. It then fails to build:
````
[ALL  ]    checking for libexpat... no
[ERROR]    configure: error: expat is missing or unusable
[ALL  ]    Makefile:8791: recipe for target 'configure-gdb' failed
[ERROR]    make[2]: *** [configure-gdb] Error 1
[ALL  ]    make[2]: Leaving directory '/home/avn/work/ctng/crosstool-ng/.build/mips-unknown-elf/build/build-gdb-cross'
````

stilor:
Just to make sure we're on the same page: I agree --without-libexpat-prefix is indeed a better solution, I just don't want anyone to reintroduce `--with-libexpat-prefix=${CT_HOST_COMPLIBS_DIR}`.

stilor:
And in your reproduction scenario, you were missing `CT_GDB_CROSS_STATIC=y` option (I have no idea why we have this as a separate option rather than depending on a CT_STATIC_TOOLCHAIN).

KirillSmirnov:
Done!

stilor:
Ok, I now reproduced the failure you described, so all pieces of the puzzle fell into place. You can ignore this rant (you probably know all that already :)); I am mostly describing the issue in more detail for myself if I ever need to get back to this.

You were a bit misleading with the initial description:
> For gdb-cross build configure is allowed to investigate system /usr/{lib,include} directories, but it does not
hurt to disable this behavior. In this case configure falls back to -lexpat, which works as expected.

For gdb-cross, our build script never did inspect the system /usr/lib directory to begin with - it has --prefix configured as `${CT_PREFIX_DIR}`; and that directory obviously can't contain an incompatible libexpat, as we are populating that directory from scratch. So you were really fixing only the native GDB; the other part is just for the sake of similarity.

And regarding this:
> We do not use these libraries [liblzma, libipt], do we? Adding respective --without-XXX-prefix would be just a waste of characters :-)

It actually should've said that we are not installing these for the target rather than that our GDB is not using them. Hence, configure does not detect them. With libexpat, we install it and its headers into sysroot, so the compiler finds the headers, but then configure uses an explicit path to the library, i.e. /usr/lib/libexpat.so, and fails.

That said, we should examine the rest of the tools/libraries for use of `AC_LIB_HAVE_LINKFLAGS` (and apparently, `AC_LIB_LINKFLAGS_BODY`) and add a respective `--without-*-prefix` to their configure. For example, it looks like gettext's configure similarly checks for some libraries we install (ncurses, libiconv) and may pick them up from the host instead.

KirillSmirnov:
> For example, it looks like gettext's configure similarly checks for some libraries we install (ncurses, libiconv) and may pick them up from the host instead.

libiconv passively keeps bothering my mind, since AM_ICONV macro really uses AC_LIB_LINKFLAGS_BODY from gnulib. Fortunately, libiconv was integrated into libc for ages, thus AM_ICONV has never been a problem on linux. Some time ago I checked msys and cygwin - there is a separate library libiconv.dll, but all my attempts to build failed before reaching gdb stage. IIRC, I did not pass beyond gettext.
I haven't tried out your PR #494 yet, though.

PR #444, opened 2016-09-10
internals: properly strip gdbserver

KirillSmirnov:
On mingw target gdbserver is suffixed with '.exe', thus
plain 'strip gdbserver' fails.

Signed-off-by: Kirill K. Smirnov <kirill.k.smirnov@gmail.com>


bhundven:
Thanks!

PR #446, opened 2016-09-12
elf2flt: add support for a reference as elf2flt version

KirillSmirnov:
elf2flt: add support for a reference as elf2flt version

### DIFF
--- a/config/binutils/binutils.in.2
+++ b/config/binutils/binutils.in.2
@@ -28,8 +28,8 @@ if ELF2FLT_GIT
 
 config ELF2FLT_GIT_CSET
     string
-    prompt "git cset"
-    default "9dbc458c6122c495bbdec8dc975a15c9d39e5ff2"
+    prompt "git cset or ref"
+    default "ref=HEAD"
     help
       Enter the git changeset to use.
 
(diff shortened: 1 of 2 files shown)


config/binutils/binutils.in.2
```diff
@@ -28,8 +28,8 @@ if ELF2FLT_GIT
 
 config ELF2FLT_GIT_CSET
     string
-    prompt "git cset"
-    default "9dbc458c6122c495bbdec8dc975a15c9d39e5ff2"
+    prompt "git cset or ref"
+    default "ref=HEAD"
```
stilor: I am a bit hesitant on this, as this affects the reproducibility of the builds. I.e., if at some point a sample in crosstool-ng starts failing, you cannot easily find a good revision for 'git bisect' in case a failure is due to some change in elf2flt - even the revisions that used to build fine would now be returning failures. Given that we don't have a build archive for all samples, tracing failures would be... fun.

KirillSmirnov: The purpose behind this PR is to incorporate the following cset for elf2flt which fixes gcc5 warnings-errors:
https://github.com/uclinux-dev/elf2flt/commit/d3aa8c887d21dbce6ec59e704bd8b7880fdd5151

Yes, I do not like the "HEAD in production" idea too. Since ref=HEAD is already used for mingw-w64, it would be OK to use it here too.

Later I found a different solution, namely passing "--disable-werror" to elf2flt configure, but have not PR'ed it yet. If this new solution is more suitable, let's close this PR.

KirillSmirnov:
Closing this PR in favor of PR #454

PR #454, opened 2016-09-20
elf2flt: do not treat warnings as errors

KirillSmirnov:
elf2flt is configured with -Werror flag by default. Disable it.

Signed-off-by: Kirill K. Smirnov <kirill.k.smirnov@gmail.com>

### DIFF
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -309,6 +309,7 @@ do_elf2flt_backend() {
         --with-binutils-include-dir=${binutils_src}/include     \
         --with-libbfd=${binutils_bld}/bfd/libbfd.a              \
         --with-libiberty=${binutils_bld}/libiberty/libiberty.a  \
+        --disable-werror                                        \
         ${elf2flt_opts}                                         \
         "${CT_ELF2FLT_EXTRA_CONFIG_ARRAY[@]}"
 


bhundven:
Thanks!

vapier:
you should generally be reporting warnings to the elf2flt project

bhundven:
@vapier +1

I'll revert this, and add some documentation to report errors upstream.

vapier:
i don't think you need to revert this ... it's better i think for end users to not get killed over weird toolchain combos/build settings, but we'd still like to see those warnings upstream

bhundven:
Well, personally if we want to get the warnings upstream, someone has to know about it.
If we disable -Werror and no one notices, they will not inform upstream.

vapier:
that's up to you :)

KirillSmirnov:
Werror is a flag for project developers (elf2flt). There is little reason for end-users (crosstool-ng) to keep the flag, since crosstool-ng is intended to be used in a wide range of environments, gcc etc. Crosstool-ng already disables Werror for binutils and gdb. So I strongly agree with Mike.

BTW some warnings get reported to the upstream eventually :-) https://github.com/uclinux-dev/elf2flt/pull/3 The problem for crosstool-ng is to pick the related elf2flt fix. Always using ^HEAD is not a solution (see PR #446 ). Bumping rev?
Summarizing all abovementioned, I think disabling Werror while building elf2flt is the best approach.

bhundven:
Bonehead change reset to the previous commit.

I'll look into what we can do in crosstool-NG to provide some reporting on warnings and errors.

stilor:
@bhundven As far as I remember, git handbook recommends against using 'reset' on a publicly available branches; the reason is that it breaks anyone's tree that's syncronized via 'git pull'. For example, here's the build tree of my testing script:
````
*   7dcd511 (HEAD, master) Merge pull request #473 from crosstool-ng/revert-454-disable-werror
|\  
| * fa46e37 Revert "elf2flt: do not treat warnings as errors"
|/  
*   5ebbb52 (origin/master, origin/HEAD) Merge pull request #468 from dajamminyogesh/master
````
git now thinks that the local `master` is ahead of `origin/master` and hence, 'git pull' does nothing. I'll reset the local tree, but in the future, please don't use 'git reset' on master.

stilor:
Actually, that warning in git book is for 'git rebase' - "Do not rebase commits that exist outside of your repository". But I think this is applicable to 'git reset', too, since it also removes commits that may exist elsewhere. Of course, I can replace 'git pull' with 'git fetch' + 'git reset', but this seems more like a workaround for a faulty practice.

vapier:
pedantically (because it's git), `git reset` is a local command and you can do whatever you want there.  the problem is pushing a non-fast forward change to a ref that's been published somewhere.

stilor:
My point is that many people would expect 'git pull' to *just work* on a public repo, and doing 'git reset --hard' breaks that assumption.

vapier:
sure, i'm not disagreeing with your overall point ... just clarifying a particular point.

Issue #449, opened 2016-09-14
ppc build places output in the wrong place


jgunthorpe:
This subtly breaks things as the linker will sometime use the dynamic library search path, not the linker search path.

For instance using near master commit 1e4040e01118b21225b46f69b698631be5912376 trying to build for powerpc-405-linux-gnu the compiler is unable to find libgcc_s.so - this is because it is a DT_NEEDED dependency from libstdc++.so:

```
$ echo  "int main(int argc,const char *argv[]) {}" > t.cc
$ bin/powerpc-405-linux-g++ -static-libgcc t.cc
/scratchl/jgg/cx/bin/../lib/gcc/powerpc-405-linux-gnu/5.4.0/../../../../powerpc-405-linux-gnu/bin/ld: warning: libgcc_s.so.1, needed by /scratchl/jgg/cx/bin/../lib/gcc/powerpc-405-linux-gnu/5.4.0/../../../../powerpc-405-linux-gnu/lib/libstdc++.so, not found (try using -rpath or -rpath-link)
```

The file is in the tree, but in the wrong place.

```
powerpc-405-linux-gnu/lib/libgcc_s.so
powerpc-405-linux-gnu/lib/libgcc_s.so.1
```

Since it is a DT_NEEDED search ld restricts itself to sysroot:

```
$ strace -o t powerpc-405-linux-g++ ...
$ grep libgcc_s t
15161 open("/scratchl/jgg/cx/bin/../powerpc-405-linux-gnu/sysroot/src/cx/powerpc-405-linux-gnu/lib32/libgcc_s.so.1", O_RDONLY) = -1 ENOENT (No such file or directory)
15161 open("/scratchl/jgg/cx/bin/../powerpc-405-linux-gnu/sysroot/usr/local/lib32/libgcc_s.so.1", O_RDONLY) = -1 ENOENT (No such file or directory)
15161 open("/scratchl/jgg/cx/bin/../powerpc-405-linux-gnu/sysroot/lib32/libgcc_s.so.1", O_RDONLY) = -1 ENOENT (No such file or directory)
15161 open("/scratchl/jgg/cx/bin/../powerpc-405-linux-gnu/sysroot/usr/lib32/libgcc_s.so.1", O_RDONLY) = -1 ENOENT (No such file or directory)
[...]

PR #459, opened 2016-10-02
Fix issue #449.

stilor:
After building final gcc, place the libraries in non-sysroot libs
into sysroot.

### DIFF
--- a/scripts/build/cc/100-gcc.sh
+++ b/scripts/build/cc/100-gcc.sh
@@ -778,7 +778,7 @@ gcc_movelibs() {
     done
 
     # Move only files, directories are for other multilibs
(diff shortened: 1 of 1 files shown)


scripts/build/cc/100-gcc.sh
```diff
@@ -778,7 +778,7 @@ gcc_movelibs() {
     done
 
     # Move only files, directories are for other multilibs
-    gcc_dir="${CT_PREFIX_DIR}/${CT_TARGET}/lib/${multi_dir}"
+    gcc_dir="${CT_PREFIX_DIR}/${CT_TARGET}/lib/${multi_os_dir}"
```
bhundven: Good catch!

bhundven:
Thanks!

PR #473, opened 2016-11-11
Revert "elf2flt: do not treat warnings as errors"

bhundven:
Reverts crosstool-ng/crosstool-ng#454

Warnings for elf2flt should be reported upstream!

### DIFF
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -309,7 +309,6 @@ do_elf2flt_backend() {
         --with-binutils-include-dir=${binutils_src}/include     \
         --with-libbfd=${binutils_bld}/bfd/libbfd.a              \
         --with-libiberty=${binutils_bld}/libiberty/libiberty.a  \
-        --disable-werror                                        \
         ${elf2flt_opts}                                         \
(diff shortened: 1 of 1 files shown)


KirillSmirnov:
I think this is a bad idea. Following the idea, Werror should be disabled entirely, for gdb and binutils as well. This brings a lot of build errors, some of them are fixed in more recent releases of binutils.

BTW this commit breaks m68k-unknown-elf sample. This may be fixed by bumping elf2flt rev, though.

stilor:
I agree with @KirillSmirnov: by extending that "upstream" logic, we shouldn't be maintaining any patches in crosstool-ng at all, just tell users to go complain to upstream projects. I think it is okay to have workarounds (such as disabling -Werror) in place until upstream fixes their stuff.

@bhundven: I spent a lot of time trying to keep samples building. If this indeed breaks the m68k-unknown-elf sample, please revert your reversal and test the changes before you commit.

As a side note, this change has not been reviewed at all - no one even had a chance to comment on it before you checked it in. What was so urgent with this change? Other trivial PRs that actually fixed some breakages had been hanging here for weeks and months.

bhundven:
It was reverted, so we're back to where we were before this revert.

stilor:
Thanks!

PR #475, opened 2016-11-12
glibc: Add glibc-2.24

troglobit:


troglobit:
Only first target completes build OK, the remaining targets seem to fail fetching Linux 4.4.3 for some reason?  My limited testing with GCC 6.2 for arm-unknown-linux-gnueabi however are positive.

stilor:
It's a problem with Travis CI; we are not relying on it right now.

PR #476, opened 2016-11-13
Fix ltrace under glibc 2.24.

stilor:
GLIBC 2.24 declared readdir_r as deprecated and suggests to use readdir.
uClibc-ng's readdir is thread-safe as well.


stilor:
Reported this to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=844355
(ltrace website says bugs are reported via debian)

PR #484, opened 2016-11-29
Fix canadian crosses

stilor:
Various fixes for canadian-cross samples.


KirillSmirnov:
Commits "Disallow dmalloc for bare metal" and "Sync up with buildroot GCC" are not related to canadian cross build. Maybe, separate PR for them?

stilor:
Syncing up GCC patches (or more specifically, restoring 910-nios2-*.patch that was lost during GCC6 upgrade) was needed to unbreak the canadian cross sample for nios2 that we have in the tree.

Disallow dmalloc/DUMA for bare metal is not, strictly speaking, related - other than just being discovered while testing the new canadian cross sample. But it is a very minor change (2 lines). I can split it up if it causes much pain in the review :D

PR #488, opened 2016-12-01
backport support for gcc 5 to glibc 2.13

cfstras:
This patch modifies the configure script in glibc 2.13 to allow building with gcc 5.

### DIFF
--- a/patches/glibc/2.13/120-support-gcc5.patch
+++ b/patches/glibc/2.13/120-support-gcc5.patch
@@ -1,0 +1,10 @@
+--- a/configure
++++ b/configure
+@@ -5041,7 +5041,7 @@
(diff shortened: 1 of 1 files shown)


cfstras:
I'm working off 1.22.
Crosstool is a wonderful utility, and I would really love it if libc 2.13 stays supported, so I'm stuck on this branch.
My target has Debian wheezy, which in turn is on glibc 2.13, and we don't want to lose vendor support by rolling our own glibc.

Same holds for #489

stilor:
Ah, ok. I don't know if/when @bhundven plans to roll out 1.22.1, so I'll leave this for him.

bhundven:
I will get to this and #489 soon.

KirillSmirnov:
Nitpicking: signed-off-by stanza is missing.

cfstras:
I knew I missed something.

stilor:
See my comment in https://github.com/crosstool-ng/crosstool-ng/pull/489

PR #489, opened 2016-12-01
backport support patch for gcc 5 to glibc 2.13 arm ports

cfstras:
This patch ports back an issue with gcc 5 where it would remove some "unused symbols", to allow building a glibc 2.13 toolchain for ARM.


stilor:
I don't think we have a manpower to maintain an old branch. So instead I decided to restore the old glibc versions on master. I tested a simple ARM-based sample with all versions from 2.12.1 to 2.17, just build-tested - passed. If you have a sample configuration that you use, I'd be happy to add it so that it is not accidentally broken.

See https://github.com/crosstool-ng/crosstool-ng/pull/586

PR #503, opened 2016-12-08
Fix gettext on canadian cross build

modbw:


stilor:
I might be missing something, but where does **libxml for host** come from? I was wrong about `xmlFree` - it is not expat symbol (those start with `XML_` prefix), it must be coming from libxml2.

Anyway, I'll try to give it a try tonight and find out how your patch works :) Thanks!

modbw:
gettext comes with its own version of libxml inside its tree. That's also the reason why adding the static define is needed due to not creating a DLL version.

stilor:
Thanks for the explanation. I'll give it a try in a few configurations and merge.

stilor:
So, I played with different configurations (canadian and simple crosses) and what happens is that even if gettext is built statically, libxml symbols are used as if they were dynamically imported from a DLL.

I am going to switch companion libs to static-only for now (there are other issues with dynamically linked companion libs - they are currently not installed, so the utilities stop working if the build tree is removed). We may fix it properly later.

With static build, the first part of your your patch (`sentence_end_required_spaces`) is not needed at this moment.

But, even if the build gets past gettext, it fails for me later at the 'final gcc for host' stage, when gcc's configure cannot find libdl in mingw. Did you disable GCC plugins, or did you have some other work around for that?

stilor:
This latest commit on your master doesn't have to do anything with gettext problem, does it?
When you push new commits onto the branch, github automatically updates pull requests. You might want to create separate branches for submitting PRs.

modbw:
I see. Sorry for that. Will close this pull request as #506 fixed the problem in a more generic way.

Issue #469, opened 2016-11-04
cross-gdb fails on Mac Sierra


oxoocoffee:
Hello,

I am trying to build arm-cortexa9_neon-linux-gnueabihf on Mac OSX Sierra with latest command line tools. So I did checked out master and I was able to build and install crosstool-ng just fine.
After creating config file and running build all did build fine until I got to cross-gdb.

To get to this point I had to first change crosstool-NG.sh and replace -static with -Bstatic.

Then I was able to build all the way to cross-gdb. I am attaching my config file.
The error I got is this

`[INFO ]  Installing cross-gdb
[EXTRA]    Configuring cross-gdb
[EXTRA]    Building cross-gdb
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gdb-7.10.1/libiberty/fibheap.c:220:30: error: use of undeclared identifier 'LONG_MIN'
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gdb-7.10.1/libiberty/fibheap.c:261:36: error: use of undeclared identifier 'LONG_MIN'
[ERROR]    make[3]: *** [fibheap.o] Error 1
[ERROR]    checking whether build environment is sane... make[2]: *** [all-libiberty] Error 2
[ERROR]    make[1]: *** [all] Error 2
[ERROR]`

Here is top of fib heap.c

`#ifdef HAVE_CONFIG_H`
`#include "config.h"`
`#endif`
`#ifdef HAVE_LIMITS_H`
`#include <limits.h>`
`#endif`
`#ifdef HAVE_STDLIB_H`
`#include <stdlib.h>`
` #endif`
` #ifdef HAVE_STRING_H`
 `#include <string.h>`

I also see this

`215729 [CFG  ]    checking whether <limits.h> defines MIN and MAX... no`
`215730 [CFG  ]    checking whether <sys/param.h> defines MIN and MAX... yes`

Also I am attaching config.log from /Volumes/Yocto/crosstool-ng/.build/arm-cortexa9_neon-linux-gnueabihf/build/build-gdb-cross/

My questions is there a way to add option to just see preprocessor running? My feeling it is picking up wrong file?

[config.log.zip](https://github.com/crosstool-ng/crosstool-ng/files/570471/config.log.zip)

[config.limits.zip](https://github.com/crosstool-ng/crosstool-ng/files/570458/config.limits.zip)

KirillSmirnov:
Hi Robert,

[...]

PR #506, opened 2016-12-11
Make companion libs static; fix gettext build on mingw

stilor:
This follows the trend set by 1*.sh scripts that configure ISL, GMP,
MPFR, CLooG, etc. Building with shared libraries presents all kinds
of problems:
- The shared libraries need to be installed into ${CT_PREFIX_DIR}.
- The binaries linked against companion libs need to have proper
RPATH, or they're looking for shared libs in
.build/${CT_PREFIX}/buildtools/lib.
- All libraries must agree as to whether they're built shared,
static, or both. Otherwise, gettext tries to link in static libncurses.a
into a shared library and fails (since libncurses was compiled without
the -fPIC switch and hence contains relocations that cannot be handled
in a shared library).

So this fixes the current mess. If we decide to re-enable building
the companion libs shared, we should probably make this dependent on
a separate suboption of CT_STATIC_TOOLCHAIN.

I also picked up a part of the suggested fix in issue 274; only one
part of the fix is needed with static libraries.

Also make gettext/libiconv built on linux, too - even though the host likely
has its own, if we don't build the toolchain in an environment where they
are present, NLS is going to be disabled, which defeats the purpose of
having them.

Also, make companion libs for target depend on the CT_SHARED_LIBS
setting.

Add a config loosely based on one reported in the issue 274.


modbw:
I successfully build my four toolchains (two cross, one canadian, one cross-native) with your stilor:canadian-gettext-mingw32 branch. No problems so far.

oxoocoffee:
I was asked to add this issue I just got. Trying to build under SierraOS. 
Attaching config and build log
[build.zip](https://github.com/crosstool-ng/crosstool-ng/files/650250/build.zip)

[INFO ]  =================================================================
[INFO ]  Installing pass-1 core C gcc compiler
[EXTRA]    Configuring core C gcc compiler
[EXTRA]    Building gcc
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gcc-6.2.0/libcpp/system.h:275:10: fatal error: 'libintl.h' file not found
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gcc-6.2.0/libcpp/system.h:275:10: fatal error: 'libintl.h' file not found
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gcc-6.2.0/libcpp/system.h:275:10: fatal error: 'libintl.h' file not found
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gcc-6.2.0/libcpp/system.h:275:10: fatal error: 'libintl.h' file not found
[ERROR]    fatal error: 'libintl.h' file not found
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gcc-6.2.0/libcpp/system.h:275:10: fatal error: 'libintl.h' file not found
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gcc-6.2.0/libcpp/system.h:275:10: fatal error: 'libintl.h' file not found
[ERROR]    gmake[2]: *** [Makefile:224: directives-only.o] Error 1
[ERROR]    gmake[2]: *** [Makefile:224: errors.o] Error 1
[ERROR]    gmake[2]: *** [Makefile:224: identifiers.o] Error 1
[ERROR]    /Volumes/Yocto/crosstool-ng/.build/src/gcc-6.2.0/libcpp/system.h:275:10: fatal error: 'libintl.h' file not found
[ERROR]    gmake[2]: *** [Makefile:224: files.o] Error 1
[ERROR]    gmake[2]: *** [Makefile:224: expr.o] Error 1
[ERROR]    gmake[2]: *** [Makefile:224: directives.o] Error 1
[ERROR]    gmake[2]: *** [Makefile:224: lex.o] Error 1
[ERROR]    gmake[2]: *** [Makefile:224: charset.o] Error 1
[ERROR]    gmake[1]: *** [Makefile:6522: all-libcpp] Error 2
[ERROR]
[ERROR]  >>
[ERROR]  >>  Build failed in step 'Installing pass-1 core C gcc compiler'
[ERROR]  >>
[ERROR]  >>  Error happened in: CT_DoExecLog[scripts/functions@259]
[ERROR]  >>        called from: do_gcc_core_backend[scripts/build/cc/100-gcc.sh@697]
[ERROR]  >>        called from: do_gcc_core_pass_1[scripts/build/cc/100-gcc.sh@271]
[ERROR]  >>        called from: do_cc_core_pass_1[scripts/build/cc.sh@35]
[ERROR]  >>        called from: main[scripts/crosstool-NG.sh@629]
[ERROR]  >>
[ERROR]  >>  For more info on this error, look at the file: 'build.log'
[ERROR]  >>  There is a list of known issues, some with workarounds, in:
[ERROR]  >>      '/opt/local/share/doc/crosstool-ng/crosstool-ng-1.22.0-360-gda3f8c4e-dirty/B - Known issues.txt'
[ERROR]
[ERROR]  (elapsed: 15:15.00)
[15:15] / gmake: *** [/opt/local/bin/ct-ng:152: build] Error 2

bhundven:
Yea, I thought we disabled static linking on macOS (not SierraOS ;) ), because they do not allow static compilation in that environment.

stilor:
@oxoocoffee You reported the same issue (libintl.h missing) in https://github.com/crosstool-ng/crosstool-ng/issues/469
Which means, this is a problem separate from the one addressed in this PR. This should solve your other issue (missing libraries after the build tree is removed) but I understand you cannot test that due to this failure.

oxoocoffee:
Sorry looks like this issue is on master. I reported it https://github.com/crosstool-ng/crosstool-ng/issues/469 again. Thank you for pointing this on IRC (aneyman)

oxoocoffee:
So looks like the "'libintl.h' file not found" was caused by scripts/build/cc/100-gcc.sh. I merged all others to my repo except scripts/build/cc/100-gcc.sh. Now it builds.
The difference is between CT_CFLAGS_FOR_HOST switched to CT_CFLAGS_FOR_BUILD

oxoocoffee:
So I was able to build finally with your changes merged to my repo. But still see this

15:08 $ pwd
/opt/crosstool-ng/linux-3-14-62/arm-cortexa9_neon-linux-gnueabihf/bin
✔ /opt/crosstool-ng/linux-3-14-62/arm-cortexa9_neon-linux-gnueabihf/bin

15:08 $ ldd arm-cortexa9_neon-linux-gnueabihf-gcc
arm-cortexa9_neon-linux-gnueabihf-gcc:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.4.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.0.0)
	**/Volumes/Yocto/crosstool-ng/.build/arm-cortexa9_neon-linux-gnueabihf/buildtools/lib/libintl.8.dylib (compatibility version 10.0.0, current version 10.4.0)**
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)

stilor:
I assume you reverted the CFLAGS_FOR_{HOST,BUILD} change? Well, that explains why your toolchain picked up some link dependencies intended for the *build* machine.

oxoocoffee:
Yes I did reverted all changes in my 100-gcc.sh. So it builds now. Still ldd gcc points to .build. 
Not sure where to start looking to fix this. Any ideas?

stilor:
1. Did you revert CFLAGS, but not LDFLAGS from the https://github.com/crosstool-ng/crosstool-ng/pull/484/commits/2091c5dd2dc01d7462ced6a8617d1f5227764f1b commit?
2. Can you try to build *this* branch, without any modifications, using 'ct-ng build.1' (to disable parallel jobs)? It is going to take a bit longer, but it makes the build log much easier to read. Please attach the resulting build log.
3. Also, can you do `find .build/<target>/buildtools -type f` after the the build and attach the file list here?

oxoocoffee:
Sure thing. I will try it. Can you fix second point "this" it is not URL. I will do clean clone and build "this" branch

oxoocoffee:
Ok So I kicked another build.1 this time. This is from unmodified stilor:canadian-gettext-mingw32
I included config, build, screen output and output from find .build/<target>/buildtools -type f

[config-build.zip](https://github.com/crosstool-ng/crosstool-ng/files/653511/config-build.zip)

stilor:
Ok, it now makes sense. What confused me with your previous build results is that your ldd output referred to dynamic companion libs in the .build directory - while the net effect of this branch is to switch everything to static! Now it complains about missing iconv symbols.

But it is also strange: the linker command line has two instances of iconv (`-liconv /Volumes/Yocto/crosstool-ng/.build/arm-cortexa9_neon-linux-gnueabihf/buildtools/lib/libiconv.a`). I wonder where it picks up the first libiconv and how come that neither of them satisfies the link dependency. Another question is how it even determined the path to the latter, if CT_CFLAGS_FOR_BUILD does not include headers in buildtools. Can you attach the `/Volumes/Yocto/crosstool-ng/.build/arm-cortexa9_neon-linux-gnueabihf/build/build-cc-gcc-core-pass-1/gcc/config.log`, too?

I think I'll try something on a separate branch - this is not an issue related to what this branch is fixing. I'll send you the branch info once I have it.

oxoocoffee:
[build-cc-gcc-core-pass-1-config.zip](https://github.com/crosstool-ng/crosstool-ng/files/654426/build-cc-gcc-core-pass-1-config.zip)

Here is that config file

stilor:
Ok, at least one mystery is solved: `-liconv` comes from libintl-supplied link flags.

Question is, though, which iconv library it picks up with this flag and why there are undefined symbols like `_iconv` even though the library in buildtools/lib is also passed on the command line. When does the C compiler for MacOS prepend an extra underscore? Does the build machine have a system libiconv? Static, dynamic, or both?

Can you do `nm` on all system's libiconv, and .build/<target>/buildtools/lib/libiconv.a and .build/<target>/build/build-cc-core-pass-1/intl/libintl.a?

If all else fails, I can try to set FOR_BUILD flags equal to FOR_HOST in case of a simple cross and see if it helps you. But I feel this is more of a dirty hack, and it is likely that canadian crosses will still break on MacOS in that case. I am reluctant to apply such half-fixes, and since Apple doesn't allow one to download/install MacOS if one doesn't own an Apple computer - I am limited in what I could debug here..

oxoocoffee:
Here is output for both files

[nm-ouput.zip](https://github.com/crosstool-ng/crosstool-ng/files/657420/nm-ouput.zip)

stilor:
Ok, so I think I know what happens. It looks like cc core pass-1 is getting compiled against system headers, but linked with a locally built libiconv. Local libiconv redefines the symbols to use libiconv_ prefix and has `#define iconv libiconv_iconv` in its headers, so this is where the unresolved symbols come from.

I'll try to address that in a separate PR, but I'll need someone to test it on MacOS.

oxoocoffee:
Let me know what an I will do it

stilor:
@oxoocoffee See if https://github.com/crosstool-ng/crosstool-ng/pull/510 makes macos happy (or at least happier)

PR #509, opened 2016-12-17
More gettext-related fixes

stilor:


scripts/build/companion_libs/330-gettext.sh
```diff
@@ -118,6 +118,14 @@ do_gettext_backend() {
         --disable-csharp                                    \
         --without-emacs                                     \
         --disable-openmp                                    \
+        --with-included-libxml                              \
+        --with-included-gettext                             \
+        --with-included-glib                                \
+        --with-included-libcroco                            \
+        --with-included-libunistring                        \
+        --with-libncurses-prefix="${prefix}"                \
```
KirillSmirnov: Very interesting. Gettext is shipped with fixed (w.r.t. #443) version of gnulib, i.e. it properly selects static library libncurses.a located in ${prefix}, ignoring shiny tasty shared library /usr/lib/libncurses.so.

stilor: Well, the real problem observed was with libxml2 - it picked up /usr/lib/libxml2.so - which in Travis environment happened to be of an older version and didn't have a couple of symbols that gettext needed. I threw in the rest of the switches just for good measure, so we may not need these. I just didn't want to test all the possible permutations of host environment vs. build options, so I went for an explicit configuration.

If you feel this is superfluous, feel free to remove these lines and post a PR :)

stilor: Here's a link to an example of the failure:
https://travis-ci.org/crosstool-ng/crosstool-ng/jobs/184700382

I wasn't able to reproduce the *failure* since my host has recent enough libxml2. But I did reproduce gettext linking to the host libxml2.

PR #512, opened 2016-12-18
Kill CT_INSTALL_DIR.

stilor:
We now use CT_PREFIX and expect the resulting toolchain to be relocatable.


stilor:
Minor update - CT_PREFIX_DIR was mkdir'ed twice.

PR #513, opened 2016-12-20
Fix ct-ng show-config.

stilor:
If configured with --enable-local, CT_NG is exported as plain 'ct-ng'
without any path. showSamples.sh then fails to invoke ct-ng (as current
directory is not in $PATH).


stilor:
@bhundven Ping?

bhundven:
I need to run a test on this PR. I'll be able to do that tonight (PST).

PR #519, opened 2017-01-06
Newlib enhancements

deece:
This pull request adds a number of Newlib options, and adds a new sample configuration (arm-nano-eabi), intended to allow the smallest possible binaries to be produced.

The config argument generation has been refactored from a number of if blocks to a list and a loop that traverses it.

This obsoletes #371


scripts/build/libc/newlib.sh
```diff
@@ -87,17 +77,44 @@ do_libc() {
...
+REENT_SMALL:reent-small
+MULTITHREAD:multithread
+WIDE_ORIENT:newlib-wide-orient
+UNBUF_STREAM_OPT:unbuf-stream-opt
+ENABLE_TARGET_OPTSPACE:target-optspace
+    "
+
+    for ynarg in $yn_args; do
+        var="CT_LIBC_NEWLIB_${ynarg%:*}"
+	eval var=\$${var}
```
stilor: Formatting glitch (tab vs space)

scripts/build/libc/newlib.sh
```diff
@@ -125,16 +128,18 @@ ENABLE_TARGET_OPTSPACE:target-optspace
...
-        --target=${CT_TARGET}                           \
-        --prefix=${CT_PREFIX_DIR}                       \
-        "${newlib_opts[@]}"                             \
+    CT_DoExecLog CFG                                               \
+    CC_FOR_BUILD="${CT_BUILD}-gcc"                                 \
+    CFLAGS_FOR_TARGET="${cflags_for_target}"                       \
+    AR_FOR_TARGET="`which ${CT_TARGET}-gcc-ar`"                    \
+    RANLIB_FOR_TARGET="`which ${CT_TARGET}-gcc-ranlib`"            \
+    ac_cv_path_AR_FOR_TARGET="`which ${CT_TARGET}-gcc-ar`"         \
+    ac_cv_path_RANLIB_FOR_TARGET="`which ${CT_TARGET}-gcc-ranlib`" \
```
stilor: Can you explain why both `AR_FOR_TARGET` and `ac_cv_path_AR_FOR_TARGET` are needed? Is it for different versions of newlib, or does newlib indeed require both?

In any case, a comment with an explanation is warranted.

scripts/build/libc/newlib.sh
deece: Ok

scripts/build/libc/newlib.sh
deece: I just tested without ac_cv_path_AR/RANLIB_FOR_TARGET and it builds fine. These were introduced when I was having trouble making the build use the update ar/ranlib, but that was a workflow problem (I didn't realise the scripts were copied sideways). I'll remove them.

stilor:
I missed this in the review, but this new sample actually uses 'unknown' as the vendor part of the tuple, not 'nano'. As a result, 'ct-ng saveconfig' will save this configuration into a wrong directory and the installation path will not be matching the sample name.

Please fix and create a new PR - I'll approve & merge.

PR #530, opened 2017-01-13
Set the vendor name to match in arm-nano-eabi sample

deece:


stilor:
I had posted a change that does the same already: https://github.com/crosstool-ng/crosstool-ng/pull/527
The difference is, my version is also run through 'saveconfig' - so the options with default values are removed from the crosstool.config. Unless you have a reason for keeping such options, I am going to merge mine.

deece:
Ok, thanks :)

PR #544, opened 2017-01-17
Propagate caddr_t patch to other glibc versions

stilor:
... 2.20, 2.21, 2.22 and 2.24.


stilor:
Trivial - merging right away (same patch exists in our tree since 2.19).

PR #545, opened 2017-01-17
Gcc compilation fix when fortran is enabled

dirkvdb:
When building a mingw toolchain with fortran enabled as a language for gcc the build fails due to an implicit declaration of free caused by a missing include in combination with the werror=implicit-function-declaration compilation flag.

This patch adds the missing include


stilor:
1. Can you submit this upstream, please?
2. Signed-off-by line is missing in your commit message.

dirkvdb:
Unfortunately gcc bugzilla account creation is restricted, so I cannot submit the patch upstream

stilor:
1. One needs to have a valid email address; is it the restriction you're talking about?
2. Without signed-off-by, I cannot do anything with your patch: http://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=HEAD#n416

dirkvdb:
I created an issue upstream (my work email was accepted).
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79134
I also modified the commit message

stilor:
Thanks!

PR #575, opened 2017-02-05
Allow some tunables for ncurses

stilor:
- Allow user to specify configure arguments to pass through to host/target
ncurses.
- Checkbox for --disable-database
- String option for --with-fallbacks


stilor:
This should resolve https://github.com/crosstool-ng/crosstool-ng/pull/404 in more generic way (I tested this by running cross-gdb with `-tui` after cleaning up the build tree).
With this change, it is also possible to pass `--with-terminfo-dirs=/usr/share/terminfo` to ncurses configure and make use of the system-wide DB.

PR #586, opened 2017-02-10
Restore old kernel/glibc versions

stilor:
It looks like there's considerable interest in using ct-ng to generate toolchains matching kernel/glibc with some LTS versions. We don't have the manpower to maintain old releases so that they can still download the tarballs, etc. So, restore this on master - at least it is easier to test in one location.


tamird:
I haven't reviewed in detail, but great to see this happen. Thank you!

stilor:
Note that this brings the support for the versions in LTS CentOS/Ubuntu, but not exact libc *sources*. Both distributions heavily patch their libcs, and maintaining these patchsets is not something ct-ng should do. As long as those patches do not affect ABIs of the libraries (i.e., provide backward compatible symbol versions, etc) this should not matter for the binaries produced by the toolchain. But again, it is outside of the purview of ct-ng.

tamird:
Makes sense to me. I would be surprised if those distributions changed the
ABI in a backwards incompatible way. Note that glibc's own releases are
backwards ABI compatible.

stilor:
Well, that's not just about *backward* compatibility but rather compatibility to *stock* libc. The distributions may not have begun with an unmodified glibc - even the first version in CentOS may already have had some patches by RedHat applied.

tamird:
Fair enough. Nevertheless, this is good enough for me.

PR #618, opened 2017-02-27
Require GNU sed on macos

stilor:
See the last comment for the explanation.


stilor:
Ah, forgot to mention. This also makes the configure check 'gsed' and use it if available.

PR #619, opened 2017-02-28
Fix elf2flt build on Cygwin in a different way

stilor:
... after discussion with elf2flt maintainer.


stilor:
Tested on cygwin & macos.

PR #630, opened 2017-03-04
Final batch of updates before 1.23

stilor:


stilor:
I'll merge this immediately after travis has a chance to run this - want my test builds pick this up over the weekend.

Issue #290, opened 2015-11-30
toolchain build fails in glibc localedef on OS X


roolebo:
The issue already popped up [this summer](https://sourceware.org/ml/crossgcc/2015-07/msg00023.html) but never had a follow up:

```
[EXTRA]    Configuring C library localedef
[DEBUG]    Extra config args passed: ''
[DEBUG]    ==> Executing: 'CFLAGS=-O2 -fno-stack-protector -U_FORTIFY_SOURCE' '/Volumes/ct-ng/.build/src/glibc-2.22/configure' '--prefix=/usr' '--cache-file=/Volumes/ct-ng/.build/armv6-rpi-linux-gnueabi/build/build-localedef/config.cache' '--without-cvs' '--disable-profile' '--without-gd' '--disable-debug'
[CFG  ]    configure: loading cache /Volumes/ct-ng/.build/armv6-rpi-linux-gnueabi/build/build-localedef/config.cache
[CFG  ]    checking build system type... x86_64-apple-darwin15.0.0
[CFG  ]    checking host system type... x86_64-apple-darwin15.0.0
[CFG  ]    checking for gcc... gcc
[CFG  ]    checking for suffix of object files... o
[CFG  ]    checking whether we are using the GNU C compiler... yes
[CFG  ]    checking whether gcc accepts -g... yes
[CFG  ]    checking for readelf... readelf
[CFG  ]    checking for g++... g++
[CFG  ]    checking whether we are using the GNU C++ compiler... yes
[CFG  ]    checking whether g++ accepts -g... yes
[CFG  ]    checking whether g++ can link programs... yes
[CFG  ]    checking for sysdeps preconfigure fragments... aarch64 alpha arm hppa i386 m68k microblaze mips nacl nios2 powerpc s390 sh sparc tile x86_64 checking whether gcc compiles in -mx32 mode by default... no
[CFG  ]
[CFG  ]    configure: running configure fragment for add-on libidn
[CFG  ]    *** The GNU C library is currently not available for this platform.
[CFG  ]    *** So far nobody cared to port it and if there is no volunteer it
[CFG  ]    *** might never happen.  So, if you have interest to see glibc on
[CFG  ]    *** this platform visit
[CFG  ]    ***  http://www.gnu.org/software/libc/porting.html
[CFG  ]    *** and join the group of porters
[ERROR]
```

Here's [the full log](https://dl.dropboxusercontent.com/u/669097/build-localedef.log.tar.xz)

[...]

PR #635, opened 2017-03-07
Skip locales on macos

stilor:


stilor:
The solution in https://github.com/crosstool-ng/crosstool-ng/issues/290 does not work with glibc 2.25. It pushes `configure` a bit further, but it still fails.

PR #639, opened 2017-03-10
When moving gcc libs, prefer the same directory as libc

stilor:
If libc ended up in a directory other than /lib, place GCC's libs
in that directory as well.


stilor:
I noticed that the glibc-based toolchains where the glibc's default installation directory is not lib (e.g., aarch64-*-linux-gnu where it is lib64) fail in buildroot because, e.g., libraries end up in /lib64 and dynamic linker is installed into /lib.

Before this change, even libraries were split between both locations. I'll propose a patch to handle ld-so being in a different directory to buildroot (this seems to be the case with multilib-musl as well).

stilor:
Causes a regression in powerpc64le-unknown-linux-gnu, where ld's and gcc's search paths differ: gcc wants to use /lib64 and ld searches /lib64le. As a result, libgcc_s.so (needed by libexpat.so but not specified explicitly via the -lgcc_s) is not found by linker.

PR #661, opened 2017-03-24
Force C preprocessor to toolchain cpp

dankm:
For multilib builds on hosts without cpp installed as cpp, only the
first glibc build completed successfully. Force the use of the
newly built target cpp for all multilib targets.

On systems with gcc installed in the way glibc expects (ie most Linux systems), it eventually falls back to /lib/cpp, and configure continues and eventually succeeds. If it isn't installed, configure fails once the real <sysroot>/usr/include/bits/stubs is installed.

### DIFF
--- a/scripts/build/libc/glibc.sh
+++ b/scripts/build/libc/glibc.sh
@@ -323,6 +323,7 @@ do_libc_backend_once() {
     CT_DoExecLog CFG                                                \
     BUILD_CC=${CT_BUILD}-gcc                                        \
     CC="${CT_TARGET}-${CT_CC} ${glibc_cflags}"                      \
+    CPP="${CT_TARGET}-cpp"                                          \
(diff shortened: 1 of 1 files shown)


dankm:
This goes along with #627. Without it on FreeBSD multilib builds fail. Non-multilib builds still succeed.

stilor:
Have you encountered this issue before or after https://github.com/crosstool-ng/crosstool-ng/commit/8600f3ce56f67a2a0d7b4a14eb9e8b1f0ec49dec was merged to master? I think it should've solved this issue as well, as it makes glibc/musl create symlinks for the target being configured when multilibs have different target tuples.

dankm:
I haven't checked since that commit went in. I'll double check this afternoon. It looks like that commit could also solve the issue.

dankm:
Yes, it looks like 8600f3c also fixed the issue.

PR #672, opened 2017-04-02
Freebsd fixes

stilor:


scripts/build/binutils/binutils.sh
```diff
@@ -199,6 +199,8 @@ do_binutils_backend() {
     CT_DoLog DEBUG "Extra config passed: '${extra_config[*]}'"
 
     CT_DoExecLog CFG                                            \
+    CC_FOR_BUILD="${CT_BUILD}-gcc"                              \
+    CFLAGS_FOR_BUILD="${cflags_for_build}"                      \
```
dankm: I didn't discover this one. Neat.

stilor: Unfortunately, it seems to be necessary in almost every package. I think I'll postpone further freebsd stuff until after 1.23. There are also some weird failures I am seeing in some samples with old glibc versions that I haven't investigated yet, so freebsd support will still be considered experimental in 1.23.

PR #675, opened 2017-04-06
libc/mingw: correct all shellcheck offenses

tamird:
Building with CT_MINGW_TOOLS unset before this change produces:

```
	/usr/local/ct-ng/lib/crosstool-ng-1.23.0-rc2/scripts/build/libc/mingw.sh: line 212: [: =: unary operator expected
```

Also fixed all offenses detected by [shellcheck](https://www.shellcheck.net/) while I'm here.

cc @stilor

### DIFF
--- a/scripts/build/libc/mingw.sh
+++ b/scripts/build/libc/mingw.sh
@@ -50,10 +50,10 @@ do_libc_start_files() {
 
     do_set_mingw_install_prefix
     CT_DoExecLog CFG        \
-    ${CONFIG_SHELL} \
+    "${CONFIG_SHELL}" \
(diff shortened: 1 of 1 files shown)


tamird:
@stilor

stilor:
I have cherry-picked the first commit into https://github.com/crosstool-ng/crosstool-ng/pull/677 . As to the second, I am not sure we need to quote all that stuff: CT_BUILD/CT_HOST tuples, as well as CT_SYSROOT are known not to contain spaces, etc. We either need to follow such strict guidelines everywhere, or allow the use of such "relaxed" unquoted versions where they are known to be safe. Anyway, the 2nd commit deferred until after 1.23.

tamird:
@stilor now that 1.23 is released, I'm going to just close this - if you want to pursue this kind of strictness, you know how to do it :)

PR #676, opened 2017-04-07
MinGW: add --enable-secure-api config option

tamird:
Without this flag, MinGW does not expose secure variants of functions
such as strcpy_s.

See https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/crt/sec_api/string_s.h#l11.

### DIFF
--- a/config/libc/mingw.in
+++ b/config/libc/mingw.in
@@ -128,6 +128,10 @@ config WINAPI_VERSION
     help
       Enter the version number of the Windows API files to use
 
+config MINGW_SECURE_API
+    bool "Expose secure API prototypes"
+    default y
+
 config MINGW_DIRECTX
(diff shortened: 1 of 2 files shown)


tamird:
@stilor

tamird:
@stilor I think this should be in 1.23.0 - it's essentially impossible to use mingw without these functions. In particular, I need to build RocksDB, which needs strncpy_s (https://github.com/facebook/rocksdb/blob/v5.1.4/port/win/port_win.cc#L145), or strcpy_s in the latest version (https://github.com/facebook/rocksdb/blob/6401a8b76be2b74e2721e888f2de2541c8d6b40e/port/win/port_win.cc#L145).

PR #679, opened 2017-04-07
libc/mingw: fix nonsense match

tamird:
@stilor looks like I goofed in the last PR.


stilor:
Already noticed that: https://github.com/crosstool-ng/crosstool-ng/pull/678

I think `case .. esac` is an overkill here, `if` will do :)

tamird:
Nice, thank you!

PR #697, opened 2017-04-23
Makefile enhancements

stilor:
1st wave of restructuring makefiles, create more at the time of 'make all' so that we don't have to recreate unchanging stuff during 'ct-ng build'


bhundven:
👍

PR #728, opened 2017-05-31
gcc: xtensa: backport fix for incorrect code generation

jcmvbkbc:
Hi Alexey,

please pull the following fix for xtensa gcc:

xtensa GCC incorrectly uses %ld/0x%lx format specifiers to output
HOST_WIDE_INT values, which results in incorrect code generation by the
compiler built for armhf host.

The original issue:
  https://github.com/qca/open-ath9k-htc-firmware/issues/130

Thanks.
-- Max


stilor:
Is it applicable to GCC 4.9/4.8?

jcmvbkbc:
You're right, I've updated the patch.

PR #732, opened 2017-06-02
Added MSP430 architecture support

awygle:
Added support for the MSP430 architecture. 

First patch to crosstool-ng, fully anticipating some feedback. In particular the hack added to set `CT_TARGET_SKIP_CONFIG_SUB="y"` in `CT_DoArchTupleValues ` seems likely to produce pushback, but it annoyed me that I couldn't use msp430-elf as my target (the canonicalizer in config.sub overrides the empty `CT_TARGET_VENDOR`).

Signed-off-by: Andrew Wygle <awygle@gmail.com>

### DIFF
--- a/config/arch/msp430.in
+++ b/config/arch/msp430.in
@@ -1,0 +1,8 @@
+# MSP430 specific config options
+
+## select ARCH_SUPPORTS_16
+## select ARCH_DEFAULT_16
+## select ARCH_REQUIRES_MULTILIB
+##
+## help The 16-bit MSP430 architecture, as defined by:
+## help     http://www.ti.com/lsds/ti/microcontrollers-16-bit-32-bit/msp/overview.page?HQS=msp430
--- a/config/target.in
+++ b/config/target.in
@@ -132,6 +132,9 @@ config ARCH_ENDIAN
 config ARCH_SUPPORTS_8
     bool
 
+config ARCH_SUPPORTS_16
+    bool
+
 config ARCH_SUPPORTS_32
     bool
 
@@ -164,6 +167,11 @@ config ARCH_8
     bool
     prompt "8-bit"
     depends on ARCH_SUPPORTS_8
+
+config ARCH_16
+    bool
+    prompt "16-bit"
+    depends on ARCH_SUPPORTS_16
 
 config ARCH_32
     bool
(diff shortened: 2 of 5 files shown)


stilor:
1. Add a sample with this architecture.
2. If it is really needed for this architecture, I'd like the `CT_TARGET_SKIP_CONFIG_SUB` to become an option in config/arch.in, which can then be selected by the architectures that require it (avr is another one). But the question is, does anything break if you use 'msp430-unknown-elf'?  If not, you might want to use `TARGET_ALIAS` or `TARGET_ALIAS_SED_EXPR` instead.

awygle:
Nothing actually breaks, it works fine.

I mostly added it because the documentation for CT_TARGET_VENDOR says that it can be set to empty to remove the vendor string. The actual behavior observed seems to contradict that, since the canonicalizer forces a vendor string. Perhaps there should be an "allow non-canonical names" option in toolchain.in which sets the `CT_TARGET_SKIP_CONFIG_SUB` flag?

stilor:
I'd rather keep the default cpu-vendor-target structure available; some software may be assuming that, e.g., "third dash-separated part is the target". So unless there's is some pressing need to *avoid* the tuple with vendor string, `TARGET_ALIAS` should be the way to go - it will still save you a few keystrokes.

The user-visible documentation on `CT_TARGET_VENDOR` says, quote:
````
      Vendor part of the target tuple.

      A tuple is of the form arch-vendor-kernel-system.
      You can set the second part, vendor, to whatever you see fit.
      Use a single word, or use underscores "_" to separate words.
      Use neither dash nor space, as it breaks things.

      Keep the default (unknown) if you don't know better.
````
I don't see where it is advertising the removal of the vendor string by setting it to empty.

awygle:
I am looking at http://crosstool-ng.github.io/docs/configuration/ , specifically the following:

> CT_TARGET_VENDOR: An identifier for your toolchain, will take place in the vendor part of the target tuple. It shall not contain spaces or dashes. Usually, keep it to a one-word string, or use underscores to separate words if you need. Avoid dots, commas, and special characters. It can be set to empty, to remove the vendor string from the target tuple.

stilor:
Ah, thanks. It seems to have been outdated then - I'll need to update it. Apparently, the text predates the addition of the `config.sub` run on the tuple.

awygle:
Added the sample and removed the config.sub hack. Note that C++ is disabled in the sample because libstdc++ is broken upstream for msp430.

PR #738, opened 2017-06-07
Fix debuild warnings/errors

stilor:
- Update .gitignore, do not place .gitignore into directories installed
in bulk
- Remove executable permissions and shebangs from the scripts that are
supposed to be invoked only via ct-ng frontent; prepend them with $(bash).
Despite what showSamples.sh said, it already has some bashisms.
- Remove --with autotools-dev and override dh_update_autotools_config
to avoid having config.{sub,guess} clobbered with older versions
- Install bash completion where Debian (now) expects it
- Update man page to use .\" as the comment delimiter, instead of
undefined macro (."); also, minor text edits.
- Install kconfig.mk without execute permission.
- Remove shell wrappers from 170-localedef-fix-trampoline.patch, we
do not use that for applying patches
- Revoke execute permissions on 210-expat.sh
- Get flags from dpkg-buildflags if available


stilor:
I'll merge these fixes for now. But, having read a few guides on packaging in debian/ubuntu, it looks like storing the packaging data (`debian/*`) inside the sources is not something that all packages do. Also, I am concerned that `debuild` always creates the version as "1.23.0-1" - even though it has been several dozen commits after "1.23.0". This may add to the confusion - i.e., is debian's "1.23.0" the actual "1.23.0" release or something after that - that hasn't been reflected in debian/changelog.

Given all that, I strongly lean towards separating the debian directory into a separate repository - that will not attempt to build out-of-wc, and will rather use a released version of crosstool-NG.

cpackham:
@stilor yeah that's the impression I get for other packages as well. Once we get _something_ into debian we can probably look at removing the debian directory from this repo. Until then it's nice to have somewhere to track this under version control.

PR #755, opened 2017-06-16
configure.ac: improve stat detection

cpackham:
BSD stat does not take a -c option so we can use this to detect the stat
flavor without relying on setting permission bits that are not portable
across different file systems.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -234,14 +234,13 @@ ACX_CHECK_PROGS_REQ([stat], [stat])
 AC_CACHE_CHECK([whether stat takes GNU or BSD format],
     [acx_cv_stat_flavor],
     [touch conftest
-     chmod 642 conftest
-     attr_bsd=`stat -f '%Lp' conftest 2>/dev/null`
-     attr_gnu=`stat -c '%a' conftest 2>/dev/null`
+     stat -c '%a' conftest 2>/dev/null
+     stat_rc=$?
      rm -f conftest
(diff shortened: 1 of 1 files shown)


stilor:
What if one of BSD (e.g. in MacOS, which is based on BSD's) variants starts recognizing `-c`?
ct-ng uses this to obtain the permissions of some directory, so the test should verify that the permissions can be reported using one of the formats that ct-ng will use.

cpackham:
Is there a homebrew version of stat?

There are cases where ````chmod 642 conftest; stat -c '%a' ```` won't behave as expected even on GNU/Linux due to the filesystem. The usage of stat is around being to remove a subdirectory of readonly parent directory. So perhaps we could make this a non-fatal case and just not preserve the mode on such systems.

Checking for ````GNU coreutils```` in the output from ````stat --version```` would address the "what if BSD stat added -c?" problem.

steakknife:
`"$(brew --prefix)"/bin/gstat` -> `"$(brew --prefix coreutils)"/bin/gstat` is available via `brew install coreutils`

[Here's the macOS `stat` man page](https://gist.github.com//88358d9b3e46f73a1e59b38f9b60be7e)

bhundven:
Thoroughness is appreciated!

stilor:
So, first question is what is failing and why. Does some FS not support the 642 permissions that we check? If that's the case, what does it support? I'd rather have a stat that *takes -c option to specify the format* than a stat that *takes -c option and does something*.

PR #756, opened 2017-06-16
travis updates

cpackham:
Running more jobs in parallel lets us squeak in under the 50 min limit.

Switch to trusty and ensure we have g++ available


cpackham:
Gah. Never to pull requests on master. I'll create a new branch and submit a new PR.

PR #759, opened 2017-06-16
DO NOT MERGE .travis.yml: add support for OSX

cpackham:
I've been playing a bit with enabling osx support in the ct-ng travis configuration. It's not working yet but I figured I'd get this out there so that someone who knows more about osx and/or travis can tell me where I'm going wrong.

Outstanding issues:
* The configure step can't detect the sed version. It correctly picks up gsed but when it checks for the version being >= 4.0 it fails. The output from gsed --version looks like something that should match the ````'GNU sed[^0-9]* [4-9]\.'```` regex but ````ACX_PROG_VERSION_REQ_STRICT```` still fails. I tried switching to GNU grep but still no lock.
* I've had to add || true to all the brew invocations because it's considered an error if you try to install something that has a newer version. I don't know if brew upgrade will work any better for things that don't have a newer version or aren't currently installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@
 dist: trusty
 sudo: false
 
+os:
+    - linux
+    - osx
+
 # 'bash' will define a generic environment without interfering environment
 # settings like "CC=gcc"
 language: bash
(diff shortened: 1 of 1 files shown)


stilor:
Can you post what is the output from `gsed --version`?
Also, I wonder what os:osx selection does with dist:trusty.
But thanks for working on this, it would be useful since macos is a frequent breakage build machine (due to non-GNU utilities and non-default paths). Is Travis supplying a case-senstive FS for macos, btw? Otherwise, it is a pointless exercise.

cpackham:
gsed --version output

````
gsed (GNU sed) 4.4
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Fenlason, Tom Lord, Ken Pizzini,
and Paolo Bonzini.
GNU sed home page: <http://www.gnu.org/software/sed/>.
General help using GNU software: <http://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org
````

According to https://docs.travis-ci.com/user/osx-ci-environment/#File-System it's HFS+. How hard is the case-sensitive file system requirement?

stilor:
We cannot rely on travis-ci anymore, unfortunately.

PR #760, opened 2017-06-16
Obey CONFIG_SHELL when configuring ltrace.

nyetwurk:
According to the INSTALL readme:

 Unfortunately, this technique does not work for `CONFIG_SHELL' due to
 an Autoconf bug.  Until the bug is fixed you can use this workaround:

     CONFIG_SHELL=/bin/bash /bin/bash ./configure CONFIG_SHELL=/bin/bash

### DIFF
--- a/scripts/build/debug/400-ltrace.sh
+++ b/scripts/build/debug/400-ltrace.sh
@@ -46,11 +46,13 @@ do_debug_ltrace_build() {
         ${CONFIG_SHELL}                 \
         ./configure --prefix=/usr
     else
-        CT_DoExecLog CFG        \
-        ${CONFIG_SHELL}         \
-        ./configure             \
-            --build=${CT_BUILD} \
-            --host=${CT_TARGET} \
+        CT_DoExecLog CFG               \
+        CONFIG_SHELL=${CONFIG_SHELL}   \
+        ${CONFIG_SHELL}                \
+        ./configure                    \
+            CONFIG_SHELL=${CONFIG_SHELL}\
+            --build=${CT_BUILD}        \
+            --host=${CT_TARGET}        \
             --prefix=/usr
     fi
 


stilor:
It looks like it is not just ltrace, it is an inherent feature [1] of autoconf 2.68 and older. Quoting from https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/config_002estatus-Invocation.html
[[[
Variable: CONFIG_SHELL

    The shell with which to run `configure` for the `--recheck` option. It must be Bourne-compatible. The default is a shell that supports `LINENO` if available, and `/bin/sh` otherwise. Invoking `configure` by hand bypasses this setting, so you may need to use a command like `CONFIG_SHELL=/bin/bash /bin/bash ./configure` to insure that the same shell is used everywhere. The absolute name of the shell should be passed. 
]]]

They don't mention the need to pass CONFIG_SHELL *after* configure, though. `crosstool-NG.sh` already sets CONFIG_SHELL in the environment (which is equivalent to putting it before configure, so this part of your change is redundant).

Did you see anything break because of missing `CONFIG_SHELL` after `./configure`?

stilor:
[1] A bug, once documented, becomes a feature :)

nyetwurk:
I did indeed see breakage in one of my builds that decided to run libtool with /bin/sh

It did not go well.

nyetwurk:
I should say though i did NOT try all combinations; just went with the recommended syntax in the INSTALL and it worked.

nyetwurk:
As you suspected, the CONFIG_SHELL after ./configure is not required.

stilor:
Well, this should be equivalent:
````
export CONFIG_SHELL=/bin/bash
./configure
````
and
````
CONFIG_SHELL=/bin/bash ./configure
````
since they both put `CONFIG_SHELL` into the environment of all child processes.

Crosstool-NG.sh already does `export CONFIG_SHELL=...`:
````
    # Set the shell to be used by ./configure scripts and by Makefiles (those
    # that support it!).
    export CONFIG_SHELL="${CT_CONFIG_SHELL}"    # for ./configure
    export SHELL="${CT_CONFIG_SHELL}"           # for Makefiles
````
I'd like to understand why it didn't work. What was the `.config`? What was the current shell (from which ct-ng was invoked)? User's login shell?

nyetwurk:
Not sure of root cause. It is easy to reproduce so I will investigate. Looks like somehow CONFIG_SHELL is being unexported.

nyetwurk:
I have verified that CONFIG_SHELL is not exported by the time we hit do_debug_ltrace_build()

nyetwurk:
Doing this via "restart" so I don't have to do the whole process, but when restoring from env.sh in scripts/functions, CONFIG_SHELL (nor any other var in env.sh) is not exported. Is this the expected behavior?

Debugging the issue without using the restart function is extremely painful; not sure there is an easier way.

It could be that it is ONLY a problem when restarting.

nyetwurk:
Verified that it is only a problem when using env.sh to restore the environment when restarting.

stilor:
As explained in #762, a a different issue is at play here.

PR #766, opened 2017-07-01
Nios2 fixes and a sample

KirillSmirnov:


samples/nios2-unknown-elf/crosstool.config
```diff
@@ -0,0 +1,5 @@
+CT_ARCH_nios2=y
+CT_CC_LANG_CXX=y
+CT_DEBUG_gdb=y
+# CT_GDB_CROSS_PYTHON is not set
+CT_GETTEXT=y
```
stilor: Curious, why is it needed?

KirillSmirnov: Hmm... Gettext is definitely a leftover from a canadian sample, that I used as an example.

stilor: Will you update the sample to drop this unneeded dependency?

PR #767, opened 2017-07-05
Packages

stilor:


stilor:
Seems to pass through all-samples testing. Bite the bullet, merging.

PR #793, opened 2017-08-04
[macOS] Patch to fix (hopefully all) gcc builds

steakknife:
Just copied a patch from 4.9.4 to all other builds. This was needed for 5.\*.\* builds to work.

Please let me know if you have any questions, comments, flames ;D, professional feedback, etc.


stilor:
Well, aside from:
- The minor issue that the patches did not apply cleanly (of which Travis CI informed you above)...
- Which, in turn, happened because the releases you tried to patch already contain this change (which is why this patch is only present in 4.8.x releases)
- Also, you didn't put Signed-Off-By line into the commit message - but in this case it shouldn't be an issue as the actual code change was signed off by GCC developers.

PR #806, opened 2017-08-21
Fix Linaro GCC build with glibc 2.26

nathanchance:
These were added by GCC in July but these branches are from May. I
suspect that they will be added to at least the 6.x and 7.x branches
but 5.x is EOL from Linaro it seems (as the base GCC version hasn't
been updated in a year and a half). For right now, these are needed.

This was testing on an arm64 build but the patches have fixes for all
supported architectures.

Signed-off-by: Nathan Chancellor <natechancellor@gmail.com>


stilor:
Will be merged via #817

PR #841, opened 2017-10-05
[binutils] Add 2.29.1

bhundven:


stilor:
I think we should replace 2.29 with 2.29.1; it is a bugfix release, not a feature release. I've only left the latest on each branch for all packages (with the exception of CLooG, which has certain versions of GCC relying on <=0.18.1 or >=0.18.4).

bhundven:
I'll update this PR to remove 2.29

stilor:
LGTM

PR #842, opened 2017-10-08
Redirect cd output to /dev/null in bootstrap

fomichev:
I'm using CDPATH in bash which results in printing the new directory.
Redirect this output to /dev/null so it doesn't mess up the bootstrap script.

Signed-off-by: Stanislav Fomichev <kernel@fomichev.me>

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -557,7 +557,7 @@ enter_fork()
         info[repository_url]=${info[repository]#* }
     fi
 
-    versions=`cd packages/${fork} && \
+    versions=`cd packages/${fork} &>/dev/null && \
         for f in */version.desc; do [ -r "${f}" ] && echo "${f%/version.desc}"; done`
     versions=`sort_versions ${versions}`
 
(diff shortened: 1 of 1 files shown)


stilor:
It is still too easy to trip over if one adds more `cd some-dir` commands elsewhere in the script. I think a more robust fix would be to unset `CDPATH` at the beginning of the script.

stilor:
Closed due to inactivity.

Not that I think it was an issue to begin with: there are countless ways of mangling one's shell environment with aliases, CDPATH, IFS, etc. Should be easy enough to do `CDPATH= ./bootstrap` (or undo whatever other changes one might have).

tsdunc:
The fix for this is pretty simple if you see issue 849. There is a patch provided there - two lines. It is not a big deal to do what you are suggesting here and add the CDPATH but it was not immediately obvious as to what the problem was and took a little digging to point to the CDPATH.

stilor:
If you read my response here, you would've noticed that my concern is not adding two lines but rather hitting this again. I don't use CDPATH so I have no interest in fixing it. If you need it, you're welcome to file a PR with the fix I described above.

tsdunc:
If you had read my response you would realize that I don't care if you fix it. I can work around it. But I suspect that dozens of others that try to use this will hit it and file more bugs.

PR #864, opened 2017-11-13
showSamples.sh: support custom package versions 

ksergeyv:


scripts/functions
```diff
@@ -2213,3 +2213,25 @@ CT_GetPkgVersion()
...
+
+# Closure of a build version selector.
+CT_DoGetPkgBuildVersion()
+{
+    if [ "${src_release}" = "y" ]; then
+        build_version="${version}"
+    elif [ -z "${devel_revision}" ]; then
+        build_version="${devel_branch}"
+    else
+        build_version="${devel_revision}"
```
stilor: Almost: for some version control systems (e.g. Subversion) it needs both branch *and* revision to identify the source.

I think `basename` is exactly the variable your looking for, so `CT_GetPkgBuildVersion` could, perhaps, be replaced with
````
CT_GetPkgVersion()
{
    local rv
    __do_GetBasename() { rv="${basename}"; }
    CT_PackageRun "${1}" __do_GetBasename
    eval "${2}=\"${rv}\""
}
````

Thanks for taking care of `showSamples.sh`!

ksergeyv: As I understand `basename` defined only when a package source code downloaded already.

Initially I tried to make a function that works right after sample selection (simply shows a user configured version from a sample).

Now I see that it is impossible due to dynamic revision selection (for example in `CT_GetVersion_git`). As I know build systems usually use fixed user provided revisions (versions) without any fuzzy logic. For example OpenWRT and LEDE have simple and deterministic download functions (https://git.lede-project.org/?p=openwrt/source.git;a=blob;f=include/download.mk;h=a3595140026e6ac060b858d8508d28704bf01a42;hb=HEAD) that take `URL` and `VERSION` variables without modifications. When the `VERSION` is invalid the build system simply gives an error.

Why did you use so sophisticated revision selection algorithms? Would not it lead to unpredictable (from a user point of view) revision selection and build results?

stilor: I haven't extensively dug into the code you referenced - i.e. how this `dl_pack` method is called and what "version" it gets supplied with. I can list what I thought of requirements when I implemented the downloaders in `ct-ng` in the way I did:
(a) It shall be possible to select the branch and/or revision. Note that some VCS need both to identify a particular version of the software. For example, in Subversion a revision is repository-wide and the same revision ID may refer to different sources when used for different branches, e.g. `^/trunk@1000` and `^/branches/foo-1.x@1000`.
(b) It shall be possible to *omit* the revision - in which case the most recent version shall be fetched on the specified branch (if any) or main line of development (trunk/master/HEAD/...).
(c) In case a version is omitted and a new revision becomes available on the specified branch/main line, the up-to-date version shall be downloaded.

These more or less match the behavior of the VCSes with regard to branch/revision specification and therefore I don't think any of that would be "unpredicatble" from user's POV.

As far as I see from a cursory glance, LEDE code you referenced does not satisfy (b) and (c) above. For (b) and (c), fetching has to be split into two steps - first, querying the revision if it is not specified (and that may result in a cached copy of the check out found locally) and second, actual fetching.

Also, I don't see any svn:// URLs actually used by any of the packages in LEDE. As I stated above, revision alone is not sufficient to fully identify the source in Subversion. Yet, as far as I can see, LEDE will save the checkout into a file that contains just the package name and version.Now how do you know if `gcc-200000.tar.xz` refers to GCC from the trunk, gcc-6-branch or gcc-7-branch, i.e. whether you can use this tarball or need to download it anew?

Back to this specific issue, indeed, `basename` is only going to be set when ct-ng attempts to download package's sources. So, for the sake of simplicity, `showSamples.sh` should perhaps just indicate that certain package is a development version (or a custom location).

There are a couple other issues - for packages where multiple forks are supported (e.g. gcc vs gcc-linaro), this outputs something like "gcc-7.1-2017.08" - i.e. the name of the base package followed by version of the fork. It should be "gcc-linaro-7.1-2017.08" in this case.

Also, just upper-casing the package name is not enough; non-alphanumerics should be replaced by an underscore. Currently, we only have dashes in libc flavors (android-ndk, mingw-w64) or fork names (\*-linaro).

If you don't mind, I'll pick your changes and make these fixes on top of them (and I also want to resolve the TBD for libc flavor selection and apply the same for CC "selection").

ksergeyv: Thanks for explanation.

Sure, please pick my changes and make fixes you deem necessary.

From my side, "gcc-custom"-like descriptions looks very confusing especially when a custom version is really a stable version (for example we always use our uClibc-ng git branch with project specific tags for toolchain releases).

stilor: So, I thought of something like this:
- `gcc-7.2.0` (released version)
- `gcc-devel` (for check out from a VCS)
- `gcc-custom` (for pre-checked out sources)

Alternatively, I can make the checkouts something like
- `gcc-7.2.0-devel-123456` or `gcc-7.2.0-devel-MYTAG` (checkout, if revision/tag is specified; presumably, one would select a matching version)
- `gcc-NEW-devel-123456` or `gcc-OLD-devel-MYTAG` (checkout with version selected as "newer than anything else" or "older than anything else")
- `gcc-NEW-devel`, `gcc-7.2.0-devel` if no revision is specified.
Does this alternative look better to you? It looked a bit too verbose to me, but I don't really object to that either.

ksergeyv: I think technically current crosstool "release" term means simply yet another download method (ftp/http/https) with predefined hardcoded URLs: I can select a custom gcc download specifying a git mirror and tag 7.2.0 which gives the same build as 7.2.0 release. So in my option additional "-devel" mark is excessive and do not indicate a real version status in general. The real source status may always be indicated with special tags or branch names in a VCS.

So I whould prefer

* `gcc-7.2.0` (released version)
* `gcc-custom` (for check out from a VCS)
* `gcc-${unique_id}` (for pre-checked out sources)

But

1. Also as I understand you tries to always make a unique name for a tarball file to know "whether you can use this tarball or need to download it anew?" But according to your approach it is impossible to determine a `${unique_id}` prior to source download when the user does not define branch/revision completely in (b) and (c) cases: you need to prefetch the source every time to make sure the latest revision not changed. Even in cases when the user defines branch/revision completely (branch="master" for git for example) it is also impossible to guarantee uniqueness without preliminary source download for each build run.

2. I did not use svn in the LEDE packages but it seems it uses a URL tail part to select a specific branch: https://searchcode.com/codesearch/view/12163642/ In crosstool to satisfy your (b) and (c) cases a single (say `CT_XXX_DEVEL_VERSION`) parameter can be used to describe svn source in format "\<branch\>@\<revision\>". So only when there is no tarball corresponding to this version an appropriate source selection algorithm may be applied but only once at a time of tarball creation.

Due to (1) I suppose it should be a user responsibility to make sure the tarball content is up-to-date and due to (2) only the one (the same) variable may be used to display, checkout and tarball naming.

stilor: 1. Not "prior to source download"; this is exactly why the fetching sources from VCS is split it two phases - "query" and "download". Query just asks the VCS what is the most recent revision on the specified branch, if needed. And yes, this does guarantee uniqueness (although there may be a slight chance of name clash for subversion because the generated ID does not distinguish branches `/a/b` and `/a_b` - I hope such cases won't happen in the real world).

2. It does not satisfy (b) because it is impossible to *omit* the version specification. Note that `DownloadMethod/svn` passes the `-r` option, which must be followed by a non-empty `$(VERSION)`. And you miss the point - if you change the branch part of the URL, but keep the VERSION (i.e. the revision) - LEDE's download method will use the pre-downloaded tarball even though it refers to a different branch. ct-ng's download implementation will fetch the sources from the newly selected branch (and will keep both tarballs separate, so if you go back to the original branch, the sources are already downloaded).

stilor: Re "-devel" being excessive, I still want to easily spot the unofficial releases being used (e.g. when one submits tickets to crosstool-NG. I decided to use the VCS name as an indicator, though, e.g. "gcc-7.2.0-svn-234567".

ksergeyv: Thanks for detailed explanation.

Will (1) work with git/master example? As I know it is impossible to know an available commit list without a source clone.

stilor: `git ls-remote git@github.com:ksergeyv/crosstool-ng.git --refs master`

ksergeyv: Thanks again.

Now I think I have no more questions.

stilor:
Will be picked via #869

PR #871, opened 2017-11-27
fix offsets of gcc 7.2.0 patches

ksergeyv:


stilor:
I assume you used some kind of a tool (since git commit IDs were also truncated). What is it?

I want to extend test-package.sh to do that for all packages, not just gcc-7.2.0.

ksergeyv:
Unfortunately it is manual work.
Commit IDs probably changed due to `git format-patch` or something I used.
Specifically here https://github.com/crosstool-ng/crosstool-ng/pull/871/commits/fd64d3775c311e44165afc8e0e5d95a3f310f0a5 ID changed due to a context change I suppose.

stilor:
And another question, why did you need this? I thought all these patches still applied, even if at a different offset, so it shouldn't have been a problem for the build.

ksergeyv:
The only nontechnical reason is I do not like warnings (since sometimes these indicate errors) and prefer clean build and compile processes.

ksergeyv:
I suppose `quilt refresh` is an appropriate command to automate such fixes.

stilor:
I figured that much, but I don't want ct-ng to switch to quilt-based patch management completely: unlike debian that needs it only during package build, we'll need it to be available on the end-user's machine. So I guess it would be something like `mkdir tmp-patches; touch tmp-patches/series; quilt setup tmp-patches/series; for p in $CT_LIB_DIR/packages/$pkg/$version/*.patch; do quilt import $p; quilt refresh; done; cp tmp-patches/*.patch $CT_LIB_DIR/packages/$pkg/$version/*.patch`.

I'll merge this for now, but don't waste your time on updating the rest of the patches manually.

PR #875, opened 2017-12-03
Bunch of upgrades: android-ndk, expat, gcc-linaro, ...

stilor:
... mingw-w64, musl, strace, uClibc-ng, linux.

Also, regenerate chksum's after @bhundven's upgrade of binutils.


stilor:
Merging immediately to allow my test engine to run through all the samples (unlike travis which only runs 6).

PR #882, opened 2017-12-14
Fix variables passing to gdb configure scripts

ksergeyv:
These changes mainly fix static linking errors when building static
native gdb and gdbserver (tested with gcc 7.2.0 + uClibc-ng 1.0.27 +
binutils 2.29.1 for MIPS):

[ALL  ]    .../lib/libstdc++.a(eh_throw.o): In function `__cxa_throw':
[ALL  ]    (.text.__cxa_throw+0x64): undefined reference to `_Unwind_RaiseException'
[ALL  ]    (.text.__cxa_throw+0x6c): undefined reference to `_Unwind_RaiseException'
[ALL  ]    .../lib/libstdc++.a(eh_throw.o): In function `__cxa_rethrow':
[ALL  ]    (.text.__cxa_rethrow+0x78): undefined reference to `_Unwind_Resume_or_Rethrow'
[ALL  ]    (.text.__cxa_rethrow+0x80): undefined reference to `_Unwind_Resume_or_Rethrow'
...

The problem is in mixing of CPP, CC, CXX, and LD with CPPFLAGS, CFLAGS,
CXXFLAGS, and LDFLAGS before passing to configure scripts.
gcc is sensitive to argument order and the scripts are normally responsible
to combine the variables in a proper way.

### DIFF
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -11,6 +11,19 @@ do_debug_gdb_extract() {
 do_debug_gdb_build() {
     local -a extra_config
 
+    # These variables should be global and shared between all packages.
+    local CT_TARGET_CPP="${CT_TARGET}-cpp"
+    local CT_TARGET_CC="${CT_TARGET}-gcc"
+    local CT_TARGET_CXX="${CT_TARGET}-g++"
+    local CT_TARGET_LD="${CT_TARGET}-ld"
+    local CT_HOST_CPP="${CT_HOST}-cpp"
+    local CT_HOST_CC="${CT_HOST}-gcc"
+    local CT_HOST_CXX="${CT_HOST}-g++"
+    local CT_HOST_LD="${CT_HOST}-ld"
+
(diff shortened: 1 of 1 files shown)


stilor:
First, thanks! Can you add a sample with the config that was failing?

Just in case I suddenly disappear for a few days - I have busy time at work, so I will have little time to spare for ct-ng until after the New Year. It is not like I am ignoring it.

ksergeyv:
Here are configurations to reproduce the errors: [crosstool-configs.tar.gz](https://github.com/crosstool-ng/crosstool-ng/files/1561601/crosstool-configs.tar.gz).

stilor:
I tried your config and it built fine even without this change.

The change looks okay, but I am trying to understand what the problem was.

ksergeyv:
I found that the errors occur after applying [a LEDE patch](https://github.com/lede-project/source/blob/master/toolchain/gcc/patches/7.2.0/850-use_shared_libgcc.patch#L28). Sorry, but I forgot to test the build without it.

PR #885, opened 2017-12-17
Fix gdb build flags forming

ksergeyv:
Build flags are scalar variables.


stilor:
This partially reverts your last change, right? IIUC, this was causing only the first argument to remain in the list after e.g. `cross_CPPFLAGS=\'echo ${cross_CPPFLAGS}\'`.

ksergeyv:
Yes, it was regression introduced by my last patch.

PR #896, opened 2018-01-09
Fix comment typo

vcgato29:
about ready to start on dyne


stilor:
I couldn't even find where the typo was and the submitter did not respond.

PR #899, opened 2018-01-19
riscv: Add ARCH_SUPPORTS_WITH_ABI and ARCH_SUPPORTS_WITH_ARCH

franzflasch:
The riscv32 (riscv32-unknown-elf) build currently defaults to:
```
  -mabi=                      		ilp32d
  -march=                     		rv32gc
```

Add the possibility to choose the default ARCH and ABI for riscv builds. 

For the sifive hifive1 board the following arch/abi settings are needed:
```
  -mabi=                      		ilp32
  -march=                     		rv32ima
```

Regards.


stilor:
Can you add a sample with these -mabi/-march options?

matthiaslischka:
I think the example is already given in the first post.
By the way, why is the build exceeding its time limit?

BR

cpackham:
@matthiaslischka unfortunately that's a travis-ci thing. With the free service builds have a hard limit of 50mins per job. Which most of the ct-ng configs manage to hit.

stilor:
I meant a sample that is included into the distribution (and hence, in testing matrix). E.g., with your configuration, run `ct-ng saveconfig`. Just set the vendor string to something specific to this board, e.g. `hifive1`.

franzflasch:
OK, I've added a sample config.

PR #912, opened 2018-02-12
cc/gcc: use correct sysroot arg for do_gcc_core_backend

M1cha:
This fixes gccs LIMITS_H_TEST detection for baremetal targets
so limits.h will be installed correctly.

Signed-off-by: Michael Zimmermann <sigmaepsilon92@gmail.com>

### DIFF
--- a/scripts/build/cc/gcc.sh
+++ b/scripts/build/cc/gcc.sh
@@ -315,10 +315,12 @@ do_gcc_core_backend() {
         core1|core2)
             CT_DoLog EXTRA "Configuring core C gcc compiler"
             log_txt="gcc"
+            extra_config+=( "${CT_CC_CORE_SYSROOT_ARG[@]}" )
             extra_user_config=( "${CT_CC_GCC_CORE_EXTRA_CONFIG_ARRAY[@]}" )
             ;;
         gcc_build|gcc_host)
             CT_DoLog EXTRA "Configuring final gcc compiler"
+            extra_config+=( "${CT_CC_SYSROOT_ARG[@]}" )
             extra_user_config=( "${CT_CC_GCC_EXTRA_CONFIG_ARRAY[@]}" )
             log_txt="final gcc compiler"
             # to inhibit the libiberty and libgcc tricks later on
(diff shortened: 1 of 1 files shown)


stilor:
I'll test this and merge. Could you describe what was the LIMITS_H_TEST?

M1cha:
https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;a=blob;f=gcc/Makefile.in;h=f675e073ecc70cb5f9ac81d5ac830196dd5a0a2a;hb=87fb575328cc5d954b91672681aacfc383134b12#l524
https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;a=blob;f=gcc/Makefile.in;h=f675e073ecc70cb5f9ac81d5ac830196dd5a0a2a;hb=87fb575328cc5d954b91672681aacfc383134b12#l2911
When gcc can't find libc's limits.h during build then gccs limits.h will not have "include_next" in it which means that none of newlibs definitions like PATH_MAX will be visible.

stilor:
Wait, I must be missing something. The CT_CC_CORE_SYSROOT_ARG is declared as a simple variable, not as an array, in crosstool-NG.sh.in:

````
        CT_CC_CORE_SYSROOT_ARG="--with-sysroot=${CT_SYSROOT_DIR}"
        CT_CC_CORE_SYSROOT_ARG="--without-headers"
````
So both "${CT_CC_CORE_SYSROOT_ARG}" and "${CT_CC_CORE_SYSROOT_ARG[@]}" should evaluate to the same value. So, your removal of `${CT_CC_CORE_SYSROOT_ARG}` from configure's direct args and adding it to `${extra_config[@]}` should not have changed how configure is invoked. What are the command line arguments passed to configure before and after your change? They should be the same.

M1cha:
you're right about it being a simple value but that has nothing to with its contents.
You looked at CT_CC_CORE_SYSROOT_ARG based on weather CT_USE_SYSROOT is enabled or not.
But this commit fixes the decision between CT_CC_CORE_SYSROOT_ARG(`--without-headers`) and CT_CC_SYSROOT_ARG(`--with-headers=${CT_HEADERS_DIR}`).

Also in case you didn't know: USE_SYSROOT is disabled for bare metal targets.

stilor:
Yes, I know about `CT_USE_SYSROOT` being disabled (I am the author of that change, hehe). Didn't notice you inserted different values for core/final versions, thanks for pointing out. Will merge.

Issue #942, opened 2018-04-14
Compiling crosstool-ng on Arch Linux


xavloose:
I was attempting to build the current development version against the latest arch libraries. I expect this affects all arch users.

crosstool-ng ./configure expects ncurses wide char support in an ncursesw directory but arch doesn't distribute ncurses with non wide char support headers at all.

current master:
bb9a11ae4fe44c22968941edb22264d6f0216dee
[configure.log](https://github.com/crosstool-ng/crosstool-ng/files/1910047/configure.log)

stilor:
Yes, I just saw that too. I took the curses detection macros from autoconf-archive, assuming they would be more portable than I could write/test... apparently, not. Will try to fix this tomorrow.

PR #943, opened 2018-04-14
modified ax_with_curses_extra.m4 to have better regocnition of <panel.h>

xavloose:
#942 I did some research and found a modification to ax_with_curses_extra.m4 that should cause it to recognize <panel.h> better. I found something about it in the changelog of this repo and reproduced the change.
https://github.com/Tlf/tlf
I figured I should give credit where it is due


stilor:
Thanks!

PR #949, opened 2018-04-27
Add compatibility with automake < 1.14

antmak:
As we support CentOS, for example, we have a problem there

    automake: warnings are treated as errors
    kconfig/Makefile.am:26: warning: compiling 'lxdialog/checklist.c' in subdir requires 'AM_PROG_CC_C_O' in
    'configure.ac'
    autoreconf: automake failed with exit status: 1

https://www.gnu.org/software/automake/manual/html_node/Public-Macros.html

CentOS 7, automake 1.13

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -87,6 +87,8 @@ AC_CONFIG_HEADERS([config.h])
 
 # Modern GCC/GDB releases require C++ support in the compiler
 AC_PROG_CC
+# Add for legacy reasons (automake < 1.14)
+AM_PROG_CC_C_O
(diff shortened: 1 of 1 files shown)


stilor:
In general, I am not aiming for perfect portability of bootstrap code (i.e., autoreconf, the bootstrap script that generates kconfig file fragments). For example, the bootstrap script expects bash 4.4+, etc. Instead, I plan to make weekly snapshots from master available for download.

But this fix is small, so will merge - until automake starts to warn about this obsolete option.

PR #952, opened 2018-04-29
debian: include curl and wget in Build-Depends

dlech:
If these are not present during build, we will end up with a ct-ng that can't download anything.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Section: devel
 Priority: optional
 Maintainer: Multiple Candidates <open@example.com>
 Build-Depends: debhelper (>= 9), autoconf, automake, autotools-dev,
- libncursesw5-dev, libncurses5-dev, bison, flex,
+ libncursesw5-dev, libncurses5-dev, bison, flex, curl, wget,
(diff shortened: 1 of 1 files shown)


stilor:
I am not sure about how to handle this.

First, strictly speaking, wget/curl are not dependencies. It could be a valid Crosstool-NG installation without them, e.g. to build from local sources or from repositories.

Second, they are not just "Build-Depends": this would allow the user to uninstall wget/curl after building crosstool-NG, which would be even worse - it would fail during the build.

Then, we really need one of wget/curl. So the dependency should be `wget | curl`.

So, I think it should be `Recommends: wget | curl` - but would that pull these dependencies before the build? If not, it may need both Build-Depends and Recommends.

dlech:
I always build debian packages in a clean chroot (i.e. I use pbuilder) so it is irrelevant whether or not curl and wget are installed on my system. Pbuilder only installs the packages in `Build-depends`, so unless curl and wget are included in the `Build-depends` I can't get a working package to be built.

I have added `Recommends: wget | curl` to this PR because, yes, it needs both.

PR #964, opened 2018-05-15
Add ARC architecture support

abrodkin:
For starters we add basic architecture support and samples for Linux-capable cores of both ARCompact (AKA ARCv1 ISA) and ARCv2 ISA.


stilor:
Thanks!

PR #970, opened 2018-05-16
Add support and sample for uclibc multilib toolchain for ARC

abrodkin:


packages/gcc/8.1.0/0019-ARC-Add-multilib-support-for-linux-targets.patch
```diff
@@ -0,0 +1,64 @@
+From 3d9f0e2c30020c21f2cf4ef30e839fdb08ad7cd3 Mon Sep 17 00:00:00 2001
+From: Alexey Brodkin <abrodkin@synopsys.com>
+Date: Wed, 16 May 2018 13:29:58 +0300
+Subject: [PATCH] ARC: Add multilib support for linux targets
```
stilor: Is it being submitted upstream? If so, can you add PR number for reference?

stilor:
Also, do we need both this and non-multilib sample configurations? I'd like to start paring down redundant configurations - a `ct-ng build-all` test now takes about 40 hours even on an decent machine...

packages/gcc/8.1.0/0019-ARC-Add-multilib-support-for-linux-targets.patch
abrodkin: Not yet. I just sent it for internal review and will shortly send it to GCC's mailing list.
Let's wait then before it gets reviewed and merged in GCC's trunk - I'll update this pull-request then.

abrodkin:
> Also, do we need both this and non-multilib sample configurations?

Well probably not as it is. If those are not really "samples" like "examples" but more of reference "defconfigs" then neither of `archs` or `arc700` make sense as they are because they indeed are subsets of multilibbed uClibc "sample". But I'll tweak a little-bit `arc-arc700-linux-uclibc` with removal of `CT_TARGET_CFLAGS` - that's rather important since that will allow us to:
 - test how `CT_DoArchUClibcCflags` quirk works
 - support one of still used HW configs that is not covered by multilibbed toolchain (see we cannot run stuff built with Load-locked(LLSC)/Store-conditional(SCOND) on HW that doesn't implement those instructions but otherwise LL/SC is recommended base-line so we build multilibbed version right for it)

Also do you mind if I add both ARC multilibbed toolchains to TravisCI?

A little-bit off-topic question - I noticed that  lot of toolchains fail to build including those built by Travis - I guess those failures are known and somebody is looking at them?

stilor:
The 'samples' serve dual purpose - they can be used by the end-users as the starting point for creation of their own configurations, and they're used as a regression test suite (`ct-ng build-all`). Maybe one day, I'll separate these two "classes" of samples to cover more possible permutations of config options with fewer samples, but this hasn't been done yet.

Re Travis, I think we should disconnect it completely: at some point, they (a) introduced a time limit how much time a job can run for (50 minutes) and (b) decreased the CPU share for the VMs running the jobs (which pushed our build time in their VM from ~30 minutes to 45+). The end result is that the jobs randomly run out of time, depending on the load of their test machines - which is hardly reliable.

stilor:
On 1, I currently use a simple shell script that takes an `fnmatch`-pattern and allows to match not by arch but by any portion of the tuple:
````
#!/bin/bash

m=$1
s=`${CTNG:-ct-ng} list-samples | grep -E '^\[.[^B].\]   ' | sed 's,^......,,' | while read n; do eval "case $n in $m) echo -n \" $n\";; esac"; done`

set -e -x
${CTNG:-ct-ng} build-all CT_SAMPLES="$s"
````
E.g., I can build all uClibc-ng samples by running `ctng-build-some '*-uclibc*'` or all AArch64 ones by `ctng-build-some 'aarch64-*'`.

abrodkin:
@stilor Maybe it worth adding that script in CT-NG source tree somewhere?

abrodkin:
@stilor Please check my updated patches.

stilor:
Looks good, merging.

PR #980, opened 2018-06-06
README: Fix spello uClibc-ng 1.21.0 -> 1.0.21

abrodkin:
Note, the same thing should be fixed in the docs here: http://crosstool-ng.github.io/docs/caveats-features/ in "uClibc/uClibc-NG configuration"
>  Before release ***1.21.0***, it was necessary...


stilor:
This text refers to crosstool-NG releases, not uClibc-ng's.

PR #984, opened 2018-06-11
Properly handle newlib configuration flags

slash3g:
Some configuration-time flags for newlib are prefixed with "newlib-", i.e. we need to pass --enable-newlib-FEATURE to the configure script. Unfortunately, newlib.sh do not add this prefix when enabling the features atexit-dynamic-alloc, reent-small, multithread, and unbuf-stream-opt. Thus, enabling these options in crosstool-ng's config has no effect on the generated libc.
This patch fixes this bug.

### DIFF
--- a/scripts/build/libc/newlib.sh
+++ b/scripts/build/libc/newlib.sh
@@ -65,13 +65,13 @@ IO_LL:newlib-io-long-long
 NEWLIB_REGISTER_FINI:newlib-register-fini
 NANO_MALLOC:newlib-nano-malloc
 NANO_FORMATTED_IO:newlib-nano-formatted-io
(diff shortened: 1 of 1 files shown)


slash3g:
I'm currently using newlib 3.0.0.20180226; however, git shows no changes in newlib's configure.in from 2.5.0.20171222, so I think it is safe to assume that the configuration flags did not change between these two versions.

antmak:
A good fix. Just notice that I've looked at full history of newlib/configure.in . I found all old versions of newlib had a `*newlib-*` names as in your PR

slash3g:
Any news on if/when this pull request will be merged? Should I improve it in any way to make it mergeable?

PR #990, opened 2018-06-15
Fix a wrong name for the newlib's multilib option

antmak:
Just an annoying blunder


antmak:
@slash3g PTAL also

slash3g:
Looks good to me. :+1:

PR #1008, opened 2018-07-31
Add GCC 8.2

galak:
Signed-off-by: Kumar Gala <kumar.gala@linaro.org>


stilor:
Done as a part of a bulk upgrade of multiple packages. Besides, there were a few issues with the posted change: the chksum must be updated (by running `manage-packages.sh -D`), patches checked/updated/renumbered (`-a` and `-P` to the same script).

PR #1031, opened 2018-09-25
package/gcc: Don't use --with-glibc-version if version is not 123.123

abrodkin:
If GLIBC_VERY_NEW is used (for example we build Glibc from trunk)
we end-up configuring GCC with "-with-glibc-version=new" which GCC
doesn't like barking like that:
------------------>8--------------
"error: option --with-glibc-version requires a version number M.N"
------------------>8--------------

So let's make sure "-with-glibc-version" is not used at all if
Glibc version is not a numeric value.

Note the fix consists of 2 parts:
 1. Improve SED command such that it doesn't print input line if
    required pattern is not found.
 2. Check if version string is not empty and only then pass
    "--with-glibc-version" to GCC's configuration script.

Signed-off-by: Alexey Brodkin <abrodkin@synopsys.com>

### DIFF
--- a/scripts/build/cc/gcc.sh
+++ b/scripts/build/cc/gcc.sh
@@ -468,8 +468,10 @@ do_gcc_core_backend() {
         local glibc_version
 
         CT_GetPkgVersion GLIBC glibc_version
-        glibc_version=`echo "${glibc_version}" | sed 's/\([1-9][0-9]*\.[1-9][0-9]*\).*/\1/'`
-        extra_config+=("--with-glibc-version=${glibc_version}")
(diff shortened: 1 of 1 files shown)


stilor:
The problem is that without this option, glibc then fails to compile at least for some targets (e.g. powerpc64le-*-linux-gnu). See, for example, the thread here: https://www.sourceware.org/ml/libc-alpha/2017-08/msg01000.html

I'd check if `new` and `old` can be translated to some numerical equivalents, e.g. `999.999` and `0.0` respectively. If that doesn't work, I might try the fallback (`--with-long-double-128`) which should fix glibc at leat on that target - but who knows what other decisions GCC is going to make in absence of this option.

PR #1044, opened 2018-10-01
Disallow linux versions before 4.8 for ARC

abrodkin:
Inspired by a similar fix for AArch64 in 481cbaac9b23.

The point is in Linux v4.8 we introduced new ABIv4 which is now used in
both Linux kernel and up-to-date GNU tools for ARC (based on GCC 6.x+).
See [1] for more details.

[1] https://github.com/foss-for-synopsys-dwc-arc-processors/linux/wiki/ARC-Linux-Syscall-ABI-Compatibility

Signed-off-by: Alexey Brodkin <abrodkin@synopsys.com>

### DIFF
--- a/config/arch/arc.in
+++ b/config/arch/arc.in
@@ -10,6 +10,7 @@
 ## select ARCH_SUPPORTS_WITH_CPU
 ## select GCC_REQUIRE_7_or_later
 ## select BINUTILS_REQUIRE_2_30_or_later
+## select LINUX_REQUIRE_4_8_or_later if KERNEL_LINUX
 ##
 ## help The Synopsys DesignWare ARC architecture, see more info here:
 ## help     https://www.synopsys.com/designware-ip/processor-solutions/arc-processors.html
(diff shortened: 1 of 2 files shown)


stilor:
I wanted to understand what is the consequence of the ABI change: if one selects a kernel before 4.8, does it build (and uses an older ABI)? Or does it build but produces invalid binaries that will use a newer ABI on an older kernel that does not support that ABI? I see that we only support GCC7 and newer for ARC, so I guess none was really interested in building ABIv3.

abrodkin:
> I wanted to understand what is the consequence of the ABI change: if one selects a kernel before 4.8, does it build the toolchain that uses an older ABI? Or does it build but produces invalid binaries that will use a newer ABI on an older kernel that does not support that ABI?

That's exactly what happens. I.e. toolchain based in GCC 6.x+ produces ABIv4 binaries which will be incompatible with kernels prior 4.8. What's worse pre-4.8 kernels won't even recognize a problem as they don't know anything about ARCv4 ABI and will happily run binaries built with new tools and then in run-time we'll see weird issues due to 64-bit data passed improperly from user-space to the kernel.

> I see that we only support GCC7 and newer for ARC, but I don't know if GCC7+ supports both ABIv3/ABIv4 (selecting them based on the kernel version) or just ABIv4.

Nope GCC 6.x, 7.x etc as mentioned above only generate ABIv4 binaries.

PR #1060, opened 2018-10-10
Fix glibc-ports symbolic link

oblique:
It fixes the following error:

```
...
[DEBUG]    ==> Executing:  'mv' '/home/oblique/git/crosstool-ng/.build/arm-none-linux-gnueabi/src/glibc-2.12.1' '/home/oblique/git/crosstool-ng/.build/arm-none-linux-gnueabi/src/glibc'
...
[DEBUG]    ==> Executing:  'mv' '/home/oblique/git/crosstool-ng/.build/arm-none-linux-gnueabi/src/glibc-ports-2.12.1' '/home/oblique/git/crosstool-ng/.build/arm-none-linux-gnueabi/src/glibc-ports'
...
[DEBUG]    ==> Executing:  'ln' '-sf' '/home/oblique/git/crosstool-ng/.build/arm-none-linux-gnueabi/src/glibc-ports-2.12.1' '/home/oblique/git/crosstool-ng/.build/arm-none-linux-gnueabi/src/glibc-2.12.1/ports'
[ALL  ]    ln: failed to create symbolic link '/home/oblique/git/crosstool-ng/.build/arm-none-linux-gnueabi/src/glibc-2.12.1/ports': No such file or directory
```

Signed-off-by: oblique <psyberbits@gmail.com>

### DIFF
--- a/scripts/build/libc/glibc.sh
+++ b/scripts/build/libc/glibc.sh
@@ -23,8 +23,8 @@ do_libc_extract() {
         # we do not support concurrent use of the source directory
         # and next run, if using different glibc-ports source, will override
         # this symlink anyway.
-        CT_DoExecLog ALL ln -sf "${CT_GLIBC_PORTS_SRC_DIR}/${CT_GLIBC_PORTS_BASENAME}" \
-                "${CT_GLIBC_SRC_DIR}/${CT_GLIBC_BASENAME}/ports"
+        CT_DoExecLog ALL ln -sf "${CT_GLIBC_PORTS_SRC_DIR}/${CT_GLIBC_PORTS_DIR_NAME}" \
+                "${CT_GLIBC_SRC_DIR}/${CT_GLIBC_DIR_NAME}/ports"
     fi
 }
 


stilor:
Can you attach the `.config` which was failing? Thanks.

oblique:
It turns out it has to do with

```
CT_PATCH_BUNDLED_LOCAL=y
CT_PATCH_ORDER="bundled,local"
CT_PATCH_USE_LOCAL=y
```

I'm not sure how it should be handled, most probably my patch will break other configs.

My [config](https://github.com/crosstool-ng/crosstool-ng/files/2467840/config.txt).

oblique:
Another solution is to check if `patch_order` is `bundled` or `bundled,local` in the following line, but do you see any side effects or should I create a PR?

https://github.com/crosstool-ng/crosstool-ng/blob/d19c46f8d5d10c9a15ea7022d2e2532911a67e3f/scripts/functions#L2080

stilor:
Ah, it makes sense. The reason is that with "bundled" patch order, the sources for each package are placed in a shared directory, `$CT_COMMON_SRC_DIR`, so that multiple configurations may use the same sources. With "bundled,local" the sources are placed into a per-target directory, `$CT_SRC_DIR`, as the may, for example, use different local patch directories with different sets of patches. The `$CT_xxx_SRC_DIR` variable is set to either `$CT_COMMON_SRC_DIR` or to `$CT_SRC_DIR` depending on the patch order for a given package.

Now, your patch breaks the default, "bundled" patch order for glibc. Instead, it should be patching in the `$CT_SRC_DIR` always: if it is shared, that will be just the symlink to the common source directory. I'll post a patch shortly.

PR #1064, opened 2018-10-21
Try to use 'git fetch --depth 1' if possible

stilor:
It should be possible if fetching a tag/branch, and it may be possible
if fetching a changeset if a server is configured to allow it.

Fixes #986


stilor:
A side note, I tried the commands to do a shallow fetch of a specific commit - it doesn't seem to be allowed on any public servers like github or sourceware. I did configure a local git server with that option, `uploadpack.allowReachableSHA1InWant=true`, and it worked.

PR #1065, opened 2018-10-22
Supply "very old" and "very new" glibc versions

stilor:
... as 1.0 and 99.99, respectively, to gcc configure.

Fixes #1031.


abrodkin:
@stilor Thanks for doing that!

PR #1085, opened 2018-11-05
Regular directory for overlaying

antmak:
Just a useful thing, it adds an ability to use a regular directory for overlaying.

If the directory  doesn't exist, the tarball is still used.


stilor:
I've had the same change in my working copy when I tried your configuration... There's one more fix in that area, will post shortly after merging.

PR #1086, opened 2018-11-06
Provide more tool overrides

roolebo:
glibc's configure script looks for make in the following order:
  gnumake gmake make

but crosstool-ng has an override only for make. So glibc picks up
gnumake from system and fails to proceed with the build because system
make is very old.

It'd more reliable if we also override gnumake and gmake.

Additionally, add an override for bison. The two fixes help glibc build to proceed on macOS.


stilor:
I am not sure how is bison override helping anything. Crosstool-NG finds the first bison in the `$PATH`, so for override to point to the right bison, you'd have to set up `$PATH` with homebrew directories preceding the system directories. But in that case, glibc's configure would find the right bison just as well.

What am I missing?

roolebo:
@stilor bison from homebrew is keg-only, i.e. it's not in PATH. Homebrew does that to avoid mix of system tools with the ones that come from outside under the same names. The override lets glibc build under crosstool-ng to use keg-only bison without changing PATH manually for each `ct-ng build`. Here's a pending crosstool-ng formulae modification that sets BISON to bison from homebrew:
```
diff --git a/Formula/crosstool-ng.rb b/Formula/crosstool-ng.rb
index 9033243d84..a4015f7d6f 100644
--- a/Formula/crosstool-ng.rb
+++ b/Formula/crosstool-ng.rb
@@ -3,7 +3,8 @@ class CrosstoolNg < Formula
   homepage "https://crosstool-ng.github.io/"
   url "http://crosstool-ng.org/download/crosstool-ng/crosstool-ng-1.23.0.tar.xz"
   sha256 "68a43ea98ccf9cb345cb6eec494a497b224fee24c882e8c14c6713afbbe79196"
-  revision 2
+  head "https://github.com/crosstool-ng/crosstool-ng.git"
+  revision 3
 
   bottle do
     cellar :any
@@ -26,9 +27,22 @@ class CrosstoolNg < Formula
   depends_on "ncurses" if DevelopmentTools.clang_build_version >= 1000
   depends_on "xz"
 
+  if build.head?
+    depends_on "bash" => :build
+    depends_on "gettext" => :build
+    depends_on "make"
+    depends_on "bison"
+  end
+
   def install
     ENV["M4"] = "#{Formula["m4"].opt_bin}/m4"
-    ENV["MAKE"] = "/usr/bin/make" # prevent hardcoding make path from superenv
+    ENV["MAKE"] = "#{Formula["make"].opt_bin}/gmake"
+    ENV["BISON"] = "#{Formula["bison"].opt_bin}/bison"
+
+    if build.head?
+      system "./bootstrap"
+      ENV.append "LDFLAGS", "-lintl"
+    end
 
     system "./configure", "--prefix=#{prefix}"
 
```

stilor:
Thanks for the explanation. Merging.

PR #1088, opened 2018-11-07
Moxiebox support (backport from Jim Huang) + fixes

Gbillou:


stilor:
1. What is the status of the patches to gcc and config.sub - have they been submitted upstream? If yes, please add the URLs to the mailing list archive with the submission. If not, please indicate the source of the patches and their purpose.
- Why is config.sub modified? To support being called with `moxie-unknown-elf` instead of just `moxiebox`? We can generate a single-component target tuple, as we already do with the AVR architecture
- Similarly, why is OS selected as `-moxiebox`? For example, a screenshot here [https://github.com/jgarzik/moxiebox] shows GDB configured with `-elf` OS (i.e., `moxie-elf` target).

2. You don't need to change upgrade.sed as this architecture didn't exist in crosstool-NG before your patch. This script is to modify the existing .config files when options are renamed.

stilor:
I am going to cherry pick this on my branch and make changes as I think they should be. We should avoid changes to config.sub/guess if possible because user's update to them (`ct-ng updatetools`) would overwrite them with the stock versions.

stilor:
Instead of cherry-picking, implemented #1102 from scratch. We now have two separate configurations, moxie-unknown-elf (just newlib, like this configuration) and moxie-unknown-moxiebox (newlib+moxiebox runtime). The latter also cross-compiles the `sandbox` application for the host.

PR #1102, opened 2018-11-29
Moxie architecture support

stilor:


bhundven:
I don't mean that for the moxie target, as much as I do for the existing samples.

stilor:
Canadian configurations work - as they have been for the past few years, barring an occasional breakage. Cross-native was never supported by crosstool-NG directly - it is supported as a trivial case of canadian - i.e. you can first build a simple cross and then build a canadian with host==target. This can be optimized to not require a second toolchain to be built as a prerequisite, but that code has never been implemented in crosstool-NG; it requires either reshuffling of the steps or introducing new steps like "cc_for_target".

That said, I tried building the moxiebox library in a canadian cross and indeed it failed; it turned out that  I created wrong `moxiebox-*` symlinks (they point to *host* tools rather than *build* tools). Then the build compiler doesn't find some stuff (because it looks relative to the --prefix it was installed into, and not relative to sysroot); and then I made it build `sandbox` for *build* rather than *host*.

Shame on me.

PR #1128, opened 2019-01-22
Store the uClibc .config alongside the ct-ng .config

dimkr:
This way, a uClibc-based toolchain is reproducible. `.config` has the `ct-ng` version and the `ct-ng` configuration,  but the external uClibc configuration (if supplied) is not part of the build output.

### DIFF
--- a/scripts/crosstool-NG.sh
+++ b/scripts/crosstool-NG.sh
@@ -613,15 +613,7 @@ if [ -z "${CT_RESTART}" ]; then
     rm -f "${testc}"
 
     CT_DoLog EXTRA "Installing user-supplied crosstool-NG configuration"
-    CT_DoExecLog ALL mkdir -p "${CT_PREFIX_DIR}/bin"
-    CT_DoExecLog DEBUG ${install} -m 0755 "${CT_LIB_DIR}/scripts/toolchain-config.in" "${CT_PREFIX_DIR}/bin/${CT_TARGET}-ct-ng.config"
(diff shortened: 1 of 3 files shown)


scripts/crosstool-NG.sh
```diff
@@ -616,6 +616,11 @@ if [ -z "${CT_RESTART}" ]; then
     CT_DoExecLog ALL mkdir -p "${CT_PREFIX_DIR}/bin"
     CT_DoExecLog DEBUG ${install} -m 0755 "${CT_LIB_DIR}/scripts/toolchain-config.in" "${CT_PREFIX_DIR}/bin/${CT_TARGET}-ct-ng.config"
     CT_DoExecLog DEBUG ${sed} -i -e 's,@@grep@@,"'"${grep}"'",;' "${CT_PREFIX_DIR}/bin/${CT_TARGET}-ct-ng.config"
+    if [ -n "$CT_LIBC_UCLIBC_CONFIG_FILE" ]
+    then
```
stilor: If you only need an else-branch from a `-n` conditional, use `-z` instead.

scripts/crosstool-NG.sh
```diff
@@ -616,6 +616,11 @@ if [ -z "${CT_RESTART}" ]; then
...
+        ${install} -m 0755 "${CT_PREFIX_DIR}/bin/${CT_TARGET}-ct-ng.config" "${CT_PREFIX_DIR}/bin/${CT_TARGET}-libc.config"
```
stilor: I'd prefer not to burden the main driver (crosstool-NG.sh) with actions pertinent only to a particular type of C library. Please do this in uClibc.sh (e.g. in `uClibc_post_cc`).

dimkr:
@stilor: done

PR #1147, opened 2019-02-23
Update to gcc 8.3.0

galak:
Had to update the following patches so they'd apply:

0006-libgfortran-missing-include.patch
0010-crystax.patch

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>


stilor:
Thanks. In the future, please run the patches through `./maintainer/manage-packages.sh -P` (e.g., in this case `./maintainer/manage-packages.sh -P -s gcc-8.3.0`) to avoid spurious changes. ct-ng stores quilt-formatted patches.

PR #1167, opened 2019-03-20
riscv64-unknown-linux-gnu: Add C++ Support

palmer-dabbelt:
We've had very solid support for C++ for quite a while now in RISC-V
land, at least in our Linux targets.  This patch set enables C++ support
by default, which I assume most users will want.

Signed-off-by: Palmer Dabbelt <palmer@sifive.com>


palmer-dabbelt:
I'm not sure if it's too late, but I'd like to target this for the next release (1.24, if I understand correctly) if possible.  I think that will be the first released version with riscv64-unknown-linux-gnu support, so it'd be a bit easier on my end if that always builds C++ toolchain.

PR #1172, opened 2019-03-25
overlay: treat overlay as a package

jcmvbkbc:
Hello,

This change allows downloading overlay archive from an URL or getting it from a custom location.
With this change it is possible to make self-contained ct-ng configurations for xtensa toolchains for any xtensa CPU.
Please consider merging it.

### DIFF
--- a/config/arch/xtensa.in
+++ b/config/arch/xtensa.in
@@ -35,3 +35,7 @@ config ARCH_xtensa_fsf
     bool "fsf - Default configuration"
 
 endchoice
+
+if XTENSA_CUSTOM
+source "config/overlay.in"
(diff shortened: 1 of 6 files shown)


stilor:
Why has it been closed? :) I haven't even had a chance to review it yet.

I have been thinking of having a package being able to specify the sources within the ct-ng's tree for another reason: config.{guess,sub}. We need these before we have a working directory, and in some cases even when the working directory does not even exist.

jcmvbkbc:
I've found that it's broken with targets other than xtensa. I've fixed the issue that I saw and opened a different PR. Sorry for the noise.

PR #1185, opened 2019-04-17
 glibc: Add ARC support

abrodkin:
ARC port of Glibc was submitted to the mailing list recently [1]
but due to late submission it didn't make it in Glibc 2.29 release.

Hopefully by the time of next release it will be a part of upstream
release but for now we have to use off-the-tree patch.

Still it's proven to work internally as well as its test-suite
shows brilliant results as might be seen from [1]:
```
Summary of test results:
     24 FAIL
   5124 PASS
     27 UNSUPPORTED
     19 XFAIL
```

Moreover ARC's Glibc port is known to work in Buildroot, OpenEmbedded
and even Automotive Grade Linux distro so we should be good having
this patch for Glibc.

BTW the patch itself is a copy of the one I use in OE, see [2].

[1] https://sourceware.org/ml/libc-alpha/2018-12/msg00678.html
[2] https://github.com/foss-for-synopsys-dwc-arc-processors/meta-synopsys/blob/master/recipes-core/glibc/files/0031-Add-ARC-architecture.patch

P.S. Note I'm not adding multilib Glibc sample for ARC now as there's some problem on building of libatomic during final gcc configuration. Once solved I'll replace currently submitted sample with multilib one.


stilor:
What is the current status of ARC patch? I still don't see it picked up upstream. I'll merge it for now, but I don't want to be forward-porting it to the next glibc release.

PR #1186, opened 2019-04-18
Add arm cortex a7 uclibc sample

alexandrebelloni:
Signed-off-by: Alexandre Belloni <alexandre.belloni@bootlin.com>


jack2015:
Pls see the sample of   "armv7-rpi2-linux-gnueabihf"

stilor:
I am reluctant to add more samples unless they use some configuration options not otherwise tested by crosstool-NG (or have a combination of options that used to break in the past, i.e. as a regression test). I want to actually reduce the number of samples so long as they cover all the choices of architecture/kernel/libc. It has been suggested in the past to have a limited set of samples that are tested regularly and the rest are just for the demo purposes... but it's a slippery slope which has led to a few releases with unbuildable samples in the past.

If you think this configuration is valuable to add to the testing profile, please describe the reason in the sample's description.  

I have a limited computing power and a full test cycle for crosstool-NG is already taking about a week.

alexandrebelloni:
I'm fine with that, this can be dropped

PR #1192, opened 2019-05-07
Add isl 021

nolange:


nolange:
Nope, it is fit for master. The linked commit introduces isl **0.20,** but with checksums for 0.19 archives.

cpackham:
Yep sorry my bad.

PR #1198, opened 2019-05-21
Allow GNU gettext v0.19.7, which is in Ubuntu 16.04 LTS

troglobit:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -287,7 +287,7 @@ AC_FUNC_REALLOC
 AC_FUNC_ALLOCA
 
 AM_GNU_GETTEXT([external])
-AM_GNU_GETTEXT_VERSION([0.19.8])
+AM_GNU_GETTEXT_VERSION([0.19.7])
 
 # For now, curses are needed to build kconfig. We may support a command-line
 # only configuration without curses later. For now, fail in configure but


troglobit:
@cpackham I don't understand your comment?  I just wanted to have the build system accept a slightly older version of gettext that is shipped with Ubuntu.

cpackham:
What I meant was 0.19.8 _may_ have been chosen for a reason. It's part of a big patch so it may just be what was on @stilor's machine when that commit was authored or it may have an actual practical reason.

Does allowing this for Ubuntu 16.04 cause obscure failures on Alpine <3.5? I don't know. Given that users of the former vastly outnumber users of the latter I favor accepting this and dealing with any fallout if/when in occurs. But my acceptance means little, @stilor is the ultimate arbiter of what goes in.

troglobit:
@cpackham I see, that's a valid concern worth taking into consideration.  Hopefully @stilor can weigh in on this when time permits.

stilor:
I don't remember having a particular dependency on 0.19.8+, it was indeed what I had on the machine and what hence has been tested. I don't use Alpine so I didn't test on it. If 0.19.7 has known issues with Alpine, would be nice to have a configure test for this combination so that it fails meaningfully (rather than failing to link later).

troglobit:
Thank you for weighing in @stilor.  Just to be clear, are you asking me to write configure tests for Alpine to make sure gettext 0.19.7 doesn't severely impact versions of Alpine <v3.5, released in 2016, compared to supporting a Ubuntu LTS version, before considering merging this PR?

cpackham:
For my 2c I don't have any objection for merging this as is.

If this does turn out to cause an issue for older alpine users we can tweak
the configure loguc later.

troglobit:
@stilor Would it help speed things along if I update the PR to match the latest crosstool-NG master?

PR #1201, opened 2019-06-05
ARC: Support building of multi-lib Glibc toolchain

abrodkin:
From GCC's standpoint ARC's multilib items are defined by "mcpu" values
which we have quite a few and for all of them might be built optimized
cross-toolchain.

From Glibc's standpoint multilib is just multi-ABI [1] and so very limited
versions are supposed to co-exist (e.g. arc700 & archs).

Here we force Glibc to install libraries in GCC's multilib folder to create
a universal cross-toolchain that has libs optimized for multiple CPU types.

Also here we add a sample which allows to build universal Glibc Linux
toolchain for ARC.

[1] https://sourceware.org/ml/libc-alpha/2019-06/msg00018.html


abrodkin:
Pushed updated version which fixes non-multilib builds.

abrodkin:
@stilor any objections for this one? Otherwise may we get it merged?

abrodkin:
@stilor ping!

PR #1202, opened 2019-06-06
Aarch64 ol7u6 linux gnu

egeyar:
The following patch series adds a sample configuration, aarch64-ol7u6-linux-gnu, to CT-NG and enables users to build Oracle Linux 7.6 toolchain for AArch64 running UEK 5.2. We wish to promote crosstool-NG to internal and external users needing cross tools that they can generate with OL sources according to their own requirements.

Oracle Linux source code is released as src.rpms, but not as tarballs. Extracting sources from src.rpms require additional packages and is not readily supported by CT-NG. We also maintain git branches with toolchain sources per OL update. These patches make use of those branches to pull Oracle Linux toolchain sources. For that purpose, they first enable CT-NG to accept defining packages whose released versions are pulled form vcs branches. These patches later add Oracle Linux 7.6 binutils, gcc and glibc sources as well as UEK4 and UEK5. Lastly, they add the sample configuration aarch64-ol7u6-linux-gnu.


abrodkin:
@egeyar any interest in that series still? Looks similar to what we might be willing to do for 64-bit ARC at some point.

egeyar:
Hi @abrodkin 
Thanks for your response. I am redirecting your question to @jemarch and @ezannoni.

jemarch:
@abrodkin We are indeed still interested in having the support for OL integrated in crosstool-ng.  Is there anything else we would need to change in the original request?  Once this first patch is integrated, we will add support for later versions of the distro (i.e. OL 7.9 and OL 8.3)

abrodkin:
@jemarch first, note I'm not a maintainer of this project, instead we just use CT-NG internally and I found this feature to be interesting for us as well. Well, not exactly Oracle toolchains as we support Synopsys ARC processors which AFAIK are quite far from being of any interest for any kind of server/data-center oriented folks :)

My interest was more in improving CT-NG infrastructure so that we may declare a "release" of some toolchain components even if they go from a fork repo but not from upstream release tarball.

That said I cannot guide you, though if you may make sure that PR is still usable on top of the current master we may ask @cpackham to take a look at that and if all looks good get it merged. Though since it's been a while since this PR was filed I may expect some fix-ups might be needed after re-base, so please rebase and then we'll see how it goes.

cpackham:
Obsoleted by #1415

PR #1229, opened 2019-09-03
zlib: Provide other mirror

imphil:
Currently, SourceForge is down and downloads give a 500 error. That's
not overly uncommon (even less often the case these days). Fortunately,
zlib provides another mirror on their homepage, add that as option to
the package description. (https://www.zlib.net/)


bhundven:
:man_facepalming: again, sourceforge...
https://media0.giphy.com/media/cGu1pGFumy0QnzkPil/giphy.gif

PR #1234, opened 2019-09-05
linux: Add 5.x kernels

cpackham:
Add the 5.x series kernels.


lw65:
no download address

cpackham:
> no download address

Good catch. I'd pre-downloaded them to calculate the checksums hence I already had the tarballs. I'll update CT_Mirrors() to handle the potential kernel versions up to 9 (although Linus does like to be unpredictable).

cpackham:
#1253 makes this obsolete.

PR #1240, opened 2019-09-09
Add musl fts

takumin:
Since [fts\(3\)](http://man7.org/linux/man-pages/man3/fts.3.html) is not implemented in musl libc, it is necessary to link the following musl-fts library.

https://github.com/pullmoll/musl-fts


cpackham:
> Since [fts(3)](http://man7.org/linux/man-pages/man3/fts.3.html) is not implemented in musl libc, it is necessary to link the following musl-fts library.
> 
> https://github.com/pullmoll/musl-fts

Shouldn't this at least use https://github.com/void-linux/musl-fts instead of a users repo. Are there signed tarballs available?

cpackham:
You should rebase -i and push -f to clean up the mess. I don't think this needs to be any more than 1 patch

stilor:
I tend to agree with @richfelker and @bhundven here.

PR #1253, opened 2019-10-02
Updates

stilor:


scripts/functions
```diff
@@ -1786,7 +1786,7 @@ CT_Mirrors()
             '')
                 # Ignore, this happens before .config is fully evaluated
                 ;;
-            [34].*)
+            [345].*)
```
cpackham: Is it worth making this `[3-9].*` just to be future proof?

cpackham:
Had a glance over the series. Looks good to me. I'll do a few tests builds.

stilor:
Yes, thanks. I saw that but the change I had in my tree has added 5.3.x as well - which has since been released.

stilor:
Known breakage is that newer gdbserver fails to build against uClibc-ng due to `in6_addr` not visible. Looking at it right now.

cpackham:
Running:
- [ ] arm-unknown-linux-gnueabi
- [ ] powerpc64-unknown-linux-gnu
- [ ] x86_64-unknown-linux-gnu

PR #1255, opened 2019-10-10
fix toolchain build with linker plugins

nolange:
older binutils dont automatically pick up plugins,
but need to manually use wrappers like gcc-ar.

This fix allows to compile the host toolchain with -ftlo
on debian stretch, which would not work before.

These flags were used:
```
CT_EXTRA_CFLAGS_FOR_HOST="-flto"
CT_EXTRA_LDFLAGS_FOR_HOST="-flto -no-pie"
```


nolange:
sorry, picked wrong branch

PR #1256, opened 2019-10-10
Fix bootstrap

nolange:


bootstrap
```diff
@@ -801,7 +801,7 @@ msg "*** Gathering the list of data files to install"
             scripts/compile | scripts/missing | scripts/depcomp | scripts/ltmain.sh | scripts/install-sh)
                 continue
                 ;;
-            #
```
abrodkin: Do we really need this hunk in this patch?

stilor:
Which version of autoconf is failing? Note that configure.ac requires 2.67 or newer, so versions older than that would still fail, just a bit later.

nolange:
2.64, but I was able to build ct-ng (and then a toolchain) with no other issues

PR #1258, opened 2019-10-10
fix toolchain build with linker plugins

nolange:
older binutils dont automatically pick up plugins,
but need to manually use wrappers like gcc-ar.
This fix allows to compile the host toolchain with -ftlo
on debian stretch.

### DIFF
--- a/scripts/crosstool-NG.sh
+++ b/scripts/crosstool-NG.sh
@@ -437,7 +437,7 @@ if [ -z "${CT_RESTART}" ]; then
             t="${!r}-"
         fi
 
-        for tool in ar as dlltool gcc g++ gcj gnatbind gdc gnatmake ld libtool nm objcopy objdump ranlib strip windres; do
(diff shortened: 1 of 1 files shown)


nolange:
Well debian stretch has 2.28 and it does not work there. But the issue is more complicated as this automatism depends on correct links in `lib/bfd-plugins`. Automatically detecting this is hard enough with lto (think of canadian cross, or having multiple gcc versions installed), and you could theoretically have other plugins too (gold linker AFAIK).

I should add that its generally safe to use `gcc-ar` inplace of `ar` (and so on), and most build systems today will use those unconditionally (if they are available). I cant see anything negative (use this scheme for ~7 years, not a single side-effect). Those are some wrappers for the normal ar, adding the path to the compiler plugins.
You have some specific concern or this is just some unease?

cpackham:
Mainly that the implication that for newer versions the wrappers aren't necessary.

nolange:
It does not hurt using the wrappers (if they are available), and if they truly are useless (I am not sure if there aren't cases where you still need them, there could be multiple gcc installations using the same binutils - using the mathing gcc-ar seems to be the best bet) then future installations can just leave them out and buildsystems will pick up the plain binaries.

bhundven:
This probably needs to be rebased.

PR #1267, opened 2019-10-18
Make musl build use CT_TARGET_CFLAGS

nolange:
I am not really familiar with bash arrays, please check usage.

Also I noted no c-library apparently sets target ldflags?

### DIFF
--- a/config/libc/musl.in
+++ b/config/libc/musl.in
@@ -62,3 +62,10 @@ config LIBC_MUSL_OPTIMIZE
     default "auto" if LIBC_MUSL_OPTIMIZE_AUTO
     default "yes"  if LIBC_MUSL_OPTIMIZE_SPEED
     default "size" if LIBC_MUSL_OPTIMIZE_SIZE
+
+config LIBC_MUSL_EXTRA_CFLAGS
+    string
+    prompt "extra target CFLAGS"
+    default ""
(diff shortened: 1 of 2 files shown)


dimkr:
Please see #1280

nolange:
@dimkr : you could've cherry-picked my MR and placed yours on top.

Since there appears to be some overlap in our goals, and those of @stephanosio, we would work in one fork to get our stuff in shape earlier.
My goal would be to replace my private toolchain-buildscripts one day. Whats missing for that:

-   feature parity with gnu arm toolchain - ie. nano libs (PR: #1279)
-   some flags like `--disable-libstdcxx-verbose` (commit is on my fork)
-   remap debuginfo paths, so you can remote debug and the sources are automatically found (haven't started yet, using some absurdly long CFLAGS setting so far)
-   generate files for CMake, Gdb, pkg-config (untested WIP commit on my fork)

dimkr:
> @dimkr : you could've cherry-picked my MR and placed yours on top.

Yep. I wasted many hours trying to get a working toolchain. When I figured out why my toolchains produced bad binaries I built a working toolchain, gave it to the engineer assigned to the project, cleaned up and pushed my changes to crosstool-ng and only when I opened the PR, I saw yours.

PR #1270, opened 2019-10-27
prevent the glibc build from overwriting flags

nolange:
The glibc will append the content of the CFLAGS variable,
overriding previous flags.
If unset, the CFLAGS variable is not empty, so explicitly set it.
Instead prepend the default CFLAGS flags.

### DIFF
--- a/scripts/build/libc/glibc.sh
+++ b/scripts/build/libc/glibc.sh
@@ -199,7 +199,7 @@ glibc_backend_once()
     printf "%s\n" "${CT_GLIBC_CONFIGPARMS}" > configparms
 
     # glibc can't be built without -O2 (reference needed!)
-    glibc_cflags+=" -O2"
+    glibc_cflags+=" -g -O2"
 
     case "${CT_GLIBC_ENABLE_FORTIFIED_BUILD}" in
         y)  ;;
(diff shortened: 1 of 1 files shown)


scripts/build/libc/glibc.sh
```diff
@@ -199,7 +199,7 @@ glibc_backend_once()
     printf "%s\n" "${CT_GLIBC_CONFIGPARMS}" > configparms
 
     # glibc can't be built without -O2 (reference needed!)
-    glibc_cflags+=" -O2"
+    glibc_cflags+=" -g -O2"
```
abrodkin: 1. Why force adding `-g`? Is that really required
2. glibc requires building with any optimization, i.e. not only `-O2` but any `-Oxxx` should work. This requirement comes from https://sourceware.org/git/?p=glibc.git;a=blob;f=include/libc-symbols.h;h=ac6fe33d236de9c3be29b53a95711ee9f5070f6c;hb=HEAD#l75 where they check for `__OPTIMIZE__` macro which is defined for any optimizations. So we may want to implement check if `CFLAGS` has any `-O*` or not and only then set `-O2` as a safe default.

nolange: 1. Because that are the defaults if nothing is added, changing optimization level should change just  that
2. the default flags are the first flags, if you don't set any `-O*` flag later it will be used, if you set an  `-O*` flag then this (the latest) will be active (that includes `-O0`). Why is a script needed for the same behavior that the de facto standard way of overriding flags provides?

abrodkin: Fair enough. Thanks for explanation of these trivialities... seems like I didn't have enough coffee today thus asking silly things :)

PR #1276, opened 2019-11-18
Select KERNEL_SUPPORTS_SHARED_LIBS on Windows

likema:
Fix building MinGW only generating libstdc++.a

### DIFF
--- a/config/kernel/windows.in
+++ b/config/kernel/windows.in
@@ -7,3 +7,5 @@
 ## select WINDOWS
 ##
 ## help Build a toolchain targeting systems running Windows as host
+
+## select KERNEL_SUPPORTS_SHARED_LIBS


cpackham (changes requested):
I think the right thing to do would be to add `## select KERNEL_SUPPORTS_SHARED_LIBS` to config/kernel/windows.in

likema:
> I think the right thing to do would be to add `## select KERNEL_SUPPORTS_SHARED_LIBS` to config/kernel/windows.in

Fixed

cpackham (approved):

cpackham:
For future commits please make sure to include your `Signed-off-by:`

likema:
> Signed-off-by

Done

PR #1281, opened 2019-12-06
Add Recent Linux Versions

lenary:
This updates the Kernel:
- updates version 5.3 to minor release 15, and 
- adds the 5.4 kernel (at minor release 2)

These are correct according to http://www.kernel.org on 2019-12-06.


stilor:
Obsoleted by #1287, which updated all LTS kernels as well.

PR #1292, opened 2020-01-08
Update nconf.c

angyongen:
Replace legacy bzero with memset
memset is more supported that bzero, this can help with msys2 support


cpackham:
I've just merged a kconfig update in #1430. As I said in my earlier comment the upstream kconfig still uses `bzero()`. I think the way to get this resolved would be to submit a patch against the upstream kconfig and we can pick it up with another sync.

PR #1295, opened 2020-01-22
Add rsync dependency

cpackham:
Recent Linux kernels (5.3 and newer) use rsync to copy headers around.
Add the rsync package to the list of dependencies.

I've only done the debian-based distros because I don't know what the package is called for the others "rsync" would be a pretty good guess but hopefully someone else can confirm the package name


cpackham:
I've tacked on a couple of additions to this PR. For gitlab-ci we need to include sudo so we can run commands as non-root users.

Additionally I've set a default for CTNG_UID and CTNG_GID for convenience (i.e. I keep forgetting to specify them when running `docker build`)

stilor:
The idea behind CTNG_UID/GID is to create the user ID/group ID in the container that match those on the host, so that the built/installed trees are owned by the user running the docker. There's a script, `dmgr.sh`, that handles passing these arguments - I'm curious, why aren't you using it?

As to `rsync`, doesn't ubuntu19.10 need it as well?

cpackham:
Hadn't found dmgr.sh. I've just been manually building the images.

Yes everything needs rsync. I thought I got all the apt users. I think I'll
re do this PR for just rsync. Separate to the sudo/ci stuff.

cpackham:
actually 19.10 already has it in the dockerfile. I'll still re-roll the PR to add the package for the fedora based distros.

PR #1298, opened 2020-01-30
Canadian builds/Add patch for fix "gettext" compilation

Andykaban:
Signed-off-by: Andrey Kostrov <ka_ban@rambler.ru>


stilor:
It is not a correct fix, according to gnulib maintainers. `master` already has a fix backported from upstream, please try again.

PR #1306, opened 2020-02-17
testing/docker: set default value for CTNG_UID/CTNG_GID

cpackham:
Make the creation of docker images easier so that CTNG_UID/CTNG_GID have
a default value if it's not explicitly specified when building. This
will allow publishing of images on various package repositories (e.g.
docker hub, gitlab containers). dmgr.sh can still be used to set the
UID/GID to that of the current user when building a custom container.


cpackham:
The motivation for this is to be able to publish containers suitable for building toolchains to various package repositories (e.g. docker hub) which can then be part of a CI setup for crosstool--ng.

PR #1307, opened 2020-02-17
DO NOT MERGE: gitlab-ci: setup for using gitlab-ci infrastructure

cpackham:
This is based on earlier work done by Bryan. The actual .gitlab-ci.yml
is generated based on the .gitlab-ci.yml.in. A few configurations are
deemed to be keystone builds, if these don't build then there's no point
trying the others. The canadian builds are left for last since they take
the longest. Everything else ends up in the world stage.

I've been playing with gitlab in my own fork, this is the result. With
gitlab we do have the ability to set a project wide timeout for jobs.
I've had to set this to 2h 30m. One thing with gitlab.com is that there
is still an overall limit of 2000 minutes on build runner time. When I
finally got a configuration for a full run this is when I hit the limit
so I'm not sure how close to that limit we'd be generally.

We could probably just to the keystone builds for CI and leave the other
configurations for release time.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,1006 @@
+# Generated file DO NOT EDIT
+
+---
+stages:
+  - build
+  - keystone
+  - world
+  - canadian
+
+cache:
+  paths:
+    - src
+
+# probably worth pushing an image with the required packages to dockerhub.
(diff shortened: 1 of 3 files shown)


cpackham:
Will look into github actions instead

PR #1319, opened 2020-03-18
Update to gcc 8.4.0

cpackham:
https://gcc.gnu.org/legacy-ml/gcc/2020-03/msg00042.html


cpackham:
Actually I think I messed up on this PR. There is now a gcc/8.4.0/8.3.0 directory

bhundven:
Oof, sorry about that!

PR #1322, opened 2020-03-22
uClibc-ng: Fix checksums for 1.0.33 release

abrodkin:
sha256 checksums were taken from release location [1] and match
manually calculated ones.

The rest is manually calculated for tarballs downloaded from [1].

[1] https://downloads.uclibc-ng.org/releases/1.0.33/


abrodkin:
Well there's something weird is going on.
In release archives now I see https://cgit.uclibc-ng.org/cgi/cgit/uclibc-ng.git/commit/?id=6f7c6883b72134d1ae28a6223ab9c9738d033867 already included which I noticed because of this patch (https://github.com/crosstool-ng/crosstool-ng/blob/master/packages/uClibc-ng/1.0.33/0000-missing-header.patch) application failure:
```
[EXTRA]    Patching uClibc-ng-1.0.33
[ALL  ]    patching file libc/stdlib/malloc-standard/malloc.h
[ALL  ]    Hunk #1 FAILED at 23.
[ALL  ]    1 out of 1 hunk FAILED -- saving rejects to file libc/stdlib/malloc-standard/malloc.h.rej
[ERROR]
[ERROR]  >>
[ERROR]  >>  Build failed in step 'Extracting and patching toolchain components'
```

@wbx-github any ideas what was that?

cpackham:
Closing this PR as I believe it's all been covered by #1326

PR #1324, opened 2020-04-01
Add support for no-mmu microblaze

msteveb:
no-mmu architectures need to be explicitly listed
in CT_DoKernelTupleValues

Signed-off-by: Steve Bennett <steveb@workware.net.au>


cpackham (commented):
Hi @msteveb, I think this fell through the gaps (gee I wonder why *gestures vaguely at the world in general*).

I think this just needs rebasing on top of master and it's good to go.

msteveb:
Sure. Over a year later and I'm still here, so I can do that :-)

PR #1325, opened 2020-04-01
Don't force -fPIC when building libelf

msteveb:
microblaze-nommu (BFLT) need libelf to be built without -fPIC,
but crosstool-ng forces -fPIC.

It is not clear if this needs to be configurable per-platform

Signed-off-by: Steve Bennett <steveb@workware.net.au>


bhundven:
I would suggest making this per-platform.
Maybe an option called: `ARCH_SUPPORTS_FPIC`
Then apply that to config/arch/... for everything except microblaze.
```
## select ARCH_SUPPORTS_FPIC
```

PR #1326, opened 2020-04-03
uClibc-ng-1.0.33 latest chksum updated. Patch not required.

ItsNayabSD:
Signed-off-by: Nayab Basha Sayed <basha@nayab.xyz>


nayabbashasayed:
uClibc-ng-1.0.33.tar.lz checksums removed from `chksum` file.

PR #1333, opened 2020-05-03
confdata: fix format overflow warnings for tmpname, newname

THEBOSS619:
Increase the size of tmpname and newname to silence this warning [-Wformat-overflow=] and to prevent further warnings for newer GCC's releases GCC 8.1+


THEBOSS619:
Pretty much makes sense 👍 thank you 🙂

cpackham:
I've just merged a kconfig update in #1430 so this PR should no longer be necessary. I'd appreciate some testing on current master. I'll leave this PR open for another week in case there is any feedback.

PR #1334, opened 2020-05-06
Fix bug where the backspace key didn't work in QTerminal.

Davester47:
I was having a hard time with QTerminal, which is the default on the LXQt. I couldn't use the backspace key to remove any text in a textbox. QTerminal uses the ASCII code 0x08 for backspace, which seems to be different than the other terminals. In this PR, I simply added `case 8:` to the necessary switch statements and `|| keycode == 8` in one if block. I checked the linux kernel's Kconfig, and it already had these exact changes.

### DIFF
--- a/kconfig/lxdialog/inputbox.c
+++ b/kconfig/lxdialog/inputbox.c
@@ -126,6 +126,7 @@ do_resize:
 			case KEY_DOWN:
 				break;
 			case KEY_BACKSPACE:
+			case 8:
 			case 127:
 				if (pos) {
 					wattrset(dialog, dlg.inputbox.atr);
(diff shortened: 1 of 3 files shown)


Davester47:
@stilor, want to take a look at this fix?

bhundven:
@Davester47 I've been meaning to update kconfig from the linux kernel. I'd like to try updating that, and have you test again. In the mean time, could you test linux-5.6.x and see if your menuconfig issue is fix there?

Davester47:
I just tested linux-5.6.5, and it is already fixed. When I was writing this PR, I checked the Kconfig from the kernel to see how they did it. I copied the changes over for the three conditionals that deal with the backspace key. The kernel checks for the keycode 8 in addition to the two codes that Crosstool-ng checks.

Davester47:
Could you accept this PR? I know its a temporary fix since you're upgrading Kconfig, but it would be nice if the backspace key worked.

cpackham:
I've just merged a kconfig update in #1430 so this PR should no longer be necessary. I'd appreciate some testing on current master. I'll leave this PR open for another week in case there is any feedback.

PR #1320, opened 2020-03-18
Update to gcc 9.3.0

egtvedt:
Forward ported patches from GCC 9.2.0 to 9.3.0, refreshed to match
current sources.

Patch 0010-crystax.patch did not apply clean and had minor adjustments
to fit new GCC 9.3.0 release.

Signed-off-by: Hans-Christian Noren Egtvedt <hegtvedt@cisco.com>

PR #1338, opened 2020-05-14
Add gcc 9.3 and 10.1

egtvedt:
Updated pull request of #1320, now also adding GCC 10.1.0 as requested by @graysky2.


egtvedt:
Only tested building GCC 10 for Aarch64, feel free to test broader.

graysky2:
> Only tested building GCC 10 for Aarch64, feel free to test broader.

Very nice, I will build for armv7h and report back here.

graysky2:
On armv7h, I built https://github.com/crosstool-ng/crosstool-ng/commit/5659366bf62b5555bf914b5f55e8a01c92d6c6f1 and applied this PR.  Built fine.  I used [this config](https://github.com/graysky2/crosstool-ng_for_distcc/blob/a3d66e93c1bcf7a294ba2cae055a2d6391f2a433/config) and built the actual toolchain.  Using it to compile seems to be working as expected.

marehr:
Using gcc10 without building GDB for build = host = x86_64; target = powerpc64 worked as well;

But, with gdb (default option), I got a build error (I unfortunately forgot to save the build.log; only have console output):

```
[INFO ]  Installing cross-gdb
[EXTRA]    Configuring cross gdb
[EXTRA]    Building cross gdb
[EXTRA]    Installing cross gdb
[EXTRA]    Installing '.gdbinit' template
[INFO ]  Installing cross-gdb: done in 262.57s (at 39:22)
[INFO ]  =================================================================
[INFO ]  Installing native gdb
[EXTRA]    Configuring native gdb
[EXTRA]    Building native gdb
[ERROR]    /home/marehr/Packages/crosstool-ng-git/powerpc/.build/powerpc64-unknown-linux-gnu/src/gdb/gdb/source-cache.c:37:10: fatal error: srchilite/sourcehighlight.h: No such file or directory
[ERROR]    make[4]: *** [Makefile:1660: source-cache.o] Error 1
[ERROR]    make[4]: *** Waiting for unfinished jobs....
[ERROR]    make[3]: *** [Makefile:9230: all-gdb] Error 2
[ERROR]    make[2]: *** [Makefile:852: all] Error 2
[ERROR]   
[ERROR]  >>
[ERROR]  >>  Build failed in step 'Installing native gdb'
[ERROR]  >>
[ERROR]  >>  Error happened in: CT_DoExecLog[scripts/functions@376]
[ERROR]  >>        called from: do_gdb_backend[scripts/build/debug/300-gdb.sh@281]
[ERROR]  >>        called from: do_debug_gdb_build[scripts/build/debug/300-gdb.sh@143]
[ERROR]  >>        called from: do_debug[scripts/build/debug.sh@35]
[ERROR]  >>        called from: main[scripts/crosstool-NG.sh@695]
```

NiLuJe:
@marehr: IIRC, GDB's source-highlight is incompatible with `-static-libstdc++`. You might want to double-check your config (and, of course, the actual log).

marehr:
@NiLuJe It was actually my first time building crosstools-ng, because I wanted to reproduce some build failures that we saw with gcc10 on ppc64 on CI. Being a newbie, I just used whatever `ct-ng powerpc64-unknown-linux-gnu` was giving me. I actually didn't need / wanted GDB (it seems to use gdb9-1 as default) so my solution was to simply disable it via `ct-ng menuconfig`. Just wanted to share my experience that this PR did not work with a default configuration, but I haven't tried a "stable" build, so this might be completely unrelated.

A bit of off-topic - sorry for that - but I always tried to use cross-compilers given by the distro, but having that good experience I think I will always use crosstools-ng from here on :) One think that I would have liked would have been to reuse my previous failed build when disabling gdb. That would have saved me some time. `build RESTART=phase` said something that I didn't configure it with that feature.

Are you interested in a cmake toolchain file for your documentation? https://gist.github.com/marehr/2cdb461e00a5d24fff12d8539136623f

vanc:
Tried with mingw64, ARMv7, ARM64, x86_64, s390x, all worked out perfectly. Thank you for the PR.

jakublabenski:
Older glibc do not detect GCC 10 properly. Build fails with 'gcc to old' error.
This patch tries to fix this for glibc 2.12.2 to 2.20. 
I've tested it with:
glibc 2.12.2 for x86 and x86_64, glibc 2.19 for mips64el
In glibc GCC detection was reworked and I think it should work fine.
I didn't check older versions.

[Allow-glibc-2.12-2.20-to-work-with-GCC-10.patch.txt](https://github.com/crosstool-ng/crosstool-ng/files/4735805/Allow-glibc-2.12-2.20-to-work-with-GCC-10.patch.txt)

egtvedt:
@jakublabenski this patch sounds like a separate relevant commit that should go in up front. Can you create a pull request?

And on a separate topic, why do we want to support so old glibcs? I would prefer they were removed since they are no longer maintained and could be a security issue for users.

NiLuJe:
Some devices happen to ship horrendously old glibc versions, up to this day.

(Hell, I'm building toolchains against glibc 2.9 over here -_-". I'm not happy about it, but it's unfortunately not all that uncommon).

galak:
@egtvedt any plans to update this to gcc 10.2?

egtvedt:
@galak yes, I can bump this with the new release. It would be great if we could eventually get these pull requests into main.

I also think there is a new binutils, and glibc is soon also releasing...

egtvedt:
Superseeded by #1338, closing.

PR #1373, opened 2020-08-16
Add picolibc support

keith-packard:
This adds support for using picolibc instead of newlib on embedded
systems.

Signed-off-by: Keith Packard <keithp@keithp.com>


scripts/build/libc/picolibc.sh
```diff
@@ -0,0 +1,116 @@
...
+picolibc_start_files()
+{
+    CT_DoStep INFO "Installing C library headers & start files"
+    CT_DoExecLog ALL cp -a "${CT_SRC_DIR}/picolibc/newlib/libc/include/." \
+    "${CT_HEADERS_DIR}"
+    if [ "${CT_ARCH_XTENSA}" = "y" ]; then
+        CT_DoLog EXTRA "Installing Xtensa headers"
+        CT_DoExecLog ALL cp -r "${CT_SRC_DIR}/picolibc/newlib/libc/sys/xtensa/include/."   \
+                               "${CT_HEADERS_DIR}"
+    fi
```
cpackham, line 17: Is this valid for picolibc? Taking a quick look at the sources I don't see a `newlib/libc/sys/xtensa` path.

scripts/build/libc/picolibc.sh
```diff
@@ -0,0 +1,116 @@
...
+    # Note: picolibc handles the build/host/target a little bit differently
+    # than one would expect:
+    #   build  : not used
+    #   host   : the machine building picolibc
+    #   target : the machine picolibc runs on
+	meson_cflags=""
+	for cflag in ${cflags_for_target}; do
+		meson_cflags="${meson_cflags} '${cflag}',"
+	done
+	cat << EOF > picolibc-cross.txt
```
cpackham, line 81: I see this adds a new host dependency on meson and ninja. Which is fine but we probably should back that up with some kind of configure time check that only allows picolibc to be selected if these tools are present. We've got such an option for lzip, wget and curl.

scripts/build/libc/picolibc.sh
keith-packard: Ah, I didn't really review what that line did -- it's from the newlib script. Not useful for picolibc.

scripts/build/libc/picolibc.sh
keith-packard: ok, I'll work on adding those checks. I'm also working on getting picolibc building on Windows using mingw and msys tools. Is that something that would be useful here, or is this just for building tooling for real computers?

cpackham: There currently isn't much real support for anything other than Linux.

With the right tools (i.e. GNU sed, awk, stat) various BSDs are supportable. macOS support has somewhat atrophied due to lack of use. I'd be interested in hearing from anyone who has tried building crosstool-ng under WSL, it might just work. Cygwin/msys2/mingw will probably be limited by the file system adaptation as ct-ng is a bit picky about being able to set permissions.

keith-packard: ok, checks added for meson and ninja. I've done some limited testing and things 'seem' to work. It would be cool to set up CI here on github and run the picolibc test suite using the built tooling.

cpackham (approved):

scripts/build/libc/picolibc.sh
cpackham: Yeah CI is a bit of a problem for ct-ng with the sheer number of permutations and the time it takes to build a compiler. travis-ci times out at 50 minutes for a build job so we were plagued by failures due to timeout, I managed to get shadow banned on gitlab-ci while testing out using that (it allows timeout configuration but has a per-user allocation of CPU time, It think I'm still about 9000 minutes in debt).

Azure looks promising and somewhat friendly to FLOSS projects (as long as I don't try my world build on it).

keith-packard: I'm running 200 separate configurations of picolibc on github these days and it seems to be working; it takes about 90 minutes to run, depend on how many github runs in parallel. Not using autotools is probably saving me though; I'm compiling somewhere around 1.2 million files.

cpackham: Hmm OK. I obviously haven't been keeping up with github actions.

keith-packard: Yeah, once I had someone show me how to avoid depending on an external dockerhub image, I've been pretty happy with how easy it is to use them for CI. Picolibc now runs tests on every push and PR. https://github.com/picolibc/picolibc/actions

keith-packard:
Is there anything else you need from me before merging this in? I've rebased it to current master and it still works fine.

cpackham:
On Wed, Sep 2, 2020, 07:12 Keith Packard <notifications@github.com> wrote:

> Is there anything else you need from me before merging this in? I've
> rebased it to current master and it still works fine.
>
It's pretty much next in the queue. Just having some issues with
powerpc64+glibc2.32 that I wanted to get sorted before merging the next
batch of PRs.

keith-packard:
> It's pretty much next in the queue. Just having some issues with powerpc64+glibc2.32 that I wanted to get sorted before merging the next batch of PRs.

Yeah, saw that. Does not look like fun :-)

 There's no hurry at all; just wanted to make sure I wasn't blocking you. Thanks for accepting my contribution!

PR #1385, opened 2020-09-10
Patches: search package patches in the top package directory

falaleevms:
Before patches for specific package were searched in packages/${pkg_name}/${version}. This means that with usage of custom
version, patches wont be applied. This commit makes ct-ng search bundled patches also in packages/${pkg_name} directory. That means that we can put some patches in this directory, that will be applied to any version of this component.

This change is useful for development purposes when you build toolchain from custom sources and you need some patches to be applied, so you can put required patches in the top package directory.

### DIFF
--- a/scripts/functions
+++ b/scripts/functions
@@ -2233,6 +2233,7 @@ CT_DoExtractPatch()
     local archive ext
     local -a patch_dirs
     local bundled_patch_dir
+    local bundled_common_patch_dir
     local local_patch_dir
     local overlay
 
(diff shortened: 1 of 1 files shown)


cpackham:
One concern I have is that it may make it harder to identify patches which need to be different between versions as such we'd end up with random breakages (particularly of older versions). On the other hand I'm really in favor of anything that makes contributing to crosstool-ng easier.

falaleevms:
@cpackham, I agree that separation of patches by versions prevents from random breakages, that's why this change does not remove this possibility. But instead gives a mechanism to apply patches to custom versions of packages. The idea is not to place patches in a top directory in a git tree, but to have a possibility to put needed patches there locally.

PR #1404, opened 2020-10-08
Don't unconditionally disable XLOCALE support.

lancethepants:
Using a custom config for uClibc-ng I wanted to enable XLOCALE but found out it is being unconditionally disabled regardless.

Signed-off-by: Lance Fredrickson <lancethepants@gmail.com>

### DIFF
--- a/scripts/build/libc/uClibc.sh
+++ b/scripts/build/libc/uClibc.sh
@@ -303,7 +303,6 @@ manage_uClibc_config()
             CT_KconfigEnableOption "UCLIBC_HAS_LOCALE" "${dst}"
             CT_KconfigDeleteOption "UCLIBC_PREGENERATED_LOCALE_DATA" "${dst}"
             CT_KconfigDeleteOption "UCLIBC_DOWNLOAD_PREGENERATED_LOCALE_DATA" "${dst}"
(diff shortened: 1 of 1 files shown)


cpackham:
I'm not sure about the impact of this. Disabling xlocale dates back a long way in crosstool-ng so I can't find any specific reason why its disabled.

lancethepants:
Yes, looking back in the code it looks like disabling xlocale predates this repo and was added in an ancestral one.

From what I observe, removing this line doesn't change anything in building the toolchain except to block the addition of xlocale in a custom config. 

With this commit I've built a vanilla toolchain without using a custom config for uClibc-ng and it continues to be built without xlocale support (no xlocale.h header to be found).  I think that enabling locale in uClibc-ng does not automatically enable xlocale, it seems it needs to be explicitly enabled.  It still is labeled as a experimental/incomplete feature.

I've only used crosstool-ng so far for arm toolchains, so please only proceed if you're feeling comfortable with the change.

What all does the Travis-ci do? Does it actually build a set of toolchains? If so, could it be checked to see if any other arches suddenly enable xlocale with this pull request?

cpackham (approved):

PR #1407, opened 2020-10-13
Clarify tree layout of local patches directory

bagasme:
Currently the help for LOCAL_PATCH_DIR did not specify the tree layout
of custom patches directory. This commit adds such explanation.

For example, the bundled patches for GCC are placed under
packages/gcc/<gcc-version>, thus custom (local) GCC patches should be
placed under $LOCAL_PATCH_DIR/gcc/<gcc-version>.

Signed-off-by: Bagas Sanjaya <bagasdotme@gmail.com>

### DIFF
--- a/config/global/extract.in
+++ b/config/global/extract.in
@@ -107,5 +107,7 @@ config LOCAL_PATCH_DIR
     help
       Enter the custom patch directory here.
       
-      Note that you must ensure that the directory containing your custom
-      patches is arranged the same way the official directory is.
+      Note that you must ensure that tree layout of the directory containing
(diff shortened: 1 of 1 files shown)


cpackham (approved):
I wonder if we should apply the same logic from #1385 to local patches.

bagasme:
@cpackham on the PR you mentioned, I interpreted `bundled common patches` as common patches bundled with crosstool-ng
(though none in reality).

bagasme:
@cpackham I agree that ct-ng should also search local patches on top package directory.

PR #1412, opened 2020-11-01
Make picolibc a companion library instead of the only C library

keith-packard:
This makes picolibc build as a companion library for the target rather than the only C library, permitting both picolibc and newlib to be built at the same time.

You can build only picolibc by selecting 'none' for the regular C library.


keith-packard:
I'd like to figure out how picolibc could also be selected as the regular C library, but I couldn't make that work without duplicating all of the picolibc configuration variables (with slightly different names), and that doesn't seem ideal.

cpackham:
Hi @keith-packard are you wanting this to go in as-is? In hindsight it kind of makes sense for a new libc to start out as a companion library so I'm happy with this.

cpackham (approved):

keith-packard:
Ideally there'd be a way to select this either as the primary C library *or* a companion library, instead of only allowing one or the other -- as a primary C library it could be used in building libstdc++, once I've got patches for that working. It's really more interesting as a companion library if you've also got dual newlib/newlib-nano support as that lets one toolchain support all three common embedded options. So, perhaps just leave it as a 'regular' C library for now and work on how to get all three libraries working at the same time?

keith-packard:
I've got patches to add picolibc support to libstdc++ under review; I'm hoping to have these merged upstream, perhaps in time for GCC 11. When those are ready, I'll create a local patch for crosstool-ng so that we don't have to wait for gcc to release them.

keith-packard:
I've created a new branch and PR that implements both picolibc and newlib-nano as companion libraries, including libstdc++ support for both. I'll close this PR

Issue #808, opened 2017-08-21
Two-pass build for GLIBC


stilor:
In [1], Joseph Myers described how GLIBC should be built with only one bootstrap pass.

Highlights:
- pass-1 (which does not build libgcc at all) should not be needed; instead, pass-2 should build a static libgcc which should suffice for glibc build
- use `--with-glibc-version`
- current GLIBC requires GCC4.9 or newer - verify and record as a dependency
- if GLIBC test suite is to be run (currently not supported by ct-ng), need to rebuild glibc with a final compiler with C++ support; perhaps, a good thing to be offered as an option anyway.

[1] https://sourceware.org/ml/libc-alpha/2017-08/msg01026.html

patrakov:
This bug also affects human-readable text at http://crosstool-ng.github.io/docs/toolchain-construction/

PR #1418, opened 2020-11-11
CI: Add MacOS X host testing

galak:


cpackham:
I'm conflicted on this one.

I know we've had issues reported by macOS users previously but with no regular crosstool-ng contributors using macOS they are often met by radio silence.
 
I note that the CI run for this took 5hrs (normally it clocks in around 2hrs) but that could also be because we've effectively doubled the number of builds. Some of the toolchains took about 30 minutes longer on macOS. This could be improved by getting round to #808 which would improve things across the board.

I also don't know if any of this is a problem, at some point github might start limiting the amount of compute time but right now there's no sign of that.

I'll give this PR a week to solicit any other comments for for or against. I'm tending towards merging it then possibly dropping out some of the longer builds.

galak:
> I'm conflicted on this one.
> 
> I know we've had issues reported by macOS users previously but with no regular crosstool-ng contributors using macOS they are often met by radio silence.
> 
> I note that the CI run for this took 5hrs (normally it clocks in around 2hrs) but that could also be because we've effectively doubled the number of builds. Some of the toolchains took about 30 minutes longer on macOS. This could be improved by getting round to #808 which would improve things across the board.
> 
> I also don't know if any of this is a problem, at some point github might start limiting the amount of compute time but right now there's no sign of that.
> 
> I'll give this PR a week to solicit any other comments for for or against. I'm tending towards merging it then possibly dropping out some of the longer builds.

One thought I had was reducing the targets that are built on macOS to `*-elf`, `*eabi` variants.  I'm not sure how many people are looking for a cross toolchain to build linux kernel/apps v doing more embedded development for a RTOS or something.

PR #377, opened 2016-03-25
Updating RPI configuration for harware float and kernel version

stef3569:
Because armv6-rpi-linux-gnueabi is outdated (hardware FPU and kernel version)
Exact copy of current armv6-rpi-linux-gnueabi configuration except for hard float and kernel version.

---
This change is [Reviewable](https://reviewable.io/reviews/crosstool-ng/crosstool-ng/377)

PR #1424, opened 2020-11-24
Update kernel and hard float for RPI 1

aaujon:
Revive https://github.com/crosstool-ng/crosstool-ng/pull/377 with samples renaming.
Kernel header used was too old to compile some tools in Busybox, works well with the new sample.


aaujon:
HI @YVbakker, the renaming was asked in this PR : https://github.com/crosstool-ng/crosstool-ng/pull/377. This was asked because there is other boards with armv6 working with this configuration. This configuration should work for raspberry pi 1 and zero.
About glibc version, I guess you mean 2.29 and 2.28 ? 
If you have to stick to a specific glibc version because of an OS, downgrading it in the toolchain compilation is the way to go. I will check later to compile it with glibc 2.28.
With the previous toolchain it was not possible to compile the latest kernel.

YVbakker:
Ah thank you for the swift elaboration @aaujon, and yes I meant 2.29 and 2.28, my bad. I just double checked and the current version of glibc (libc6) that ships with raspberry pi os (raspbian) buster is 2.28 (2.28-10+rpi1). There is a new Debian release on the way, but this is still in stage 'testing' (I'm assuming they do use a newer glibc version there).

Also I just investigated the crosstools build process some more and the error seems to occur when building the 'C library localedef' after building the c library itself successfully. The specific error is: ./../include/libc-symbols.h:534:26: error:  '__EI___errno_location' specifies less restrictive attributes than its target '__errno_location': 'const', 'nothrow' [-Werror=missing-attributes]

Should I create an issue for this?

PR #1430, opened 2020-12-10
Kconfig update

cpackham:
This updates ct-ng as far as v5.7. From v5.8 onward the way `if` directives are handled within `choice` entries has changed so the ct-ng menu definitions would need updating.


egtvedt:
Nice, have not looked at the details, but love keeping things up to date.

cpackham:
Annoyingly the updates are incompatible with the way we insert the package mirrors (hence all the CI failures). I think we could trivially go as far as v4.18. Going further requires some changes on our end. I also think I can do a better job of filtering the changelogs to just cover the bits we use (i.e. not qconf/gconf). I'll spend some more time on this next week.

PR #1438, opened 2020-12-23
libgmp: Fix cross-canadian build for macOS

abrodkin:
Currently when building cross-canadian toolchain for macOS the folowing error happens when GCC is configured:
```
ld: illegal text-relocation to '___gmp_binvert_limb_table' in
... /.build/... /buildtools/complibs-host/lib/libgmp.a(mp_minv_tab.o) from '___gmpn_divexact_1' in
... /.build/... /buildtools/complibs-host/lib/libgmp.a(dive_1.o)
```

Apparently this might be solved with GMP configured with `--with-pic`, even though we're talking about static library here.

That solution was found here: https://github.com/Homebrew/homebrew-core/pull/25470

### DIFF
--- a/scripts/build/companion_libs/100-gmp.sh
+++ b/scripts/build/companion_libs/100-gmp.sh
@@ -87,6 +87,15 @@ do_gmp_backend() {
         extra_config+=("--enable-mpbsd")
     fi
 
+    # To avoind “illegal text-relocation” linking error against
+    # the static library, see:
+    #     https://github.com/Homebrew/homebrew-core/pull/25470
+    case "${host}" in
+        *darwin*)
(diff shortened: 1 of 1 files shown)


cpackham:
@galak any comment?

galak:
@abrodkin is there a different config you are building that runs into this?  No issue with the change but want to understand when its showing up vs the Mac OS X builds we are doing in CI right now.

abrodkin:
@galak note I bump into this problem when cross-compiling (actually Canadian cross-compiling) for Mac. While in your case I guess you're building embedded cross-toolchain (i.e. for ARC, ARM etc) natively on Mac for Mac and that seems (based on your experience) to fork fine.

PR #1441, opened 2021-01-05
Allow for major bash version >= 3

isaachaller:
Build will fail at configure stage on a system that has bash >3.2 without this change.


isaachaller:
I should have read the contributing guidelines and made a different branch, I apologize.

PR #1442, opened 2021-01-09
Add patch to work round missing mips ld command line flags

majenkotech:
Signed-off-by: Matt Jenkins <matt@majenko.co.uk>

The GCC compiler passes `-mips<level>` to `ld` erroneously.  `ld` has *some* code to ignore those flags, but misses some of the more modern ones. This simply adds those extra flags into `ld`. Of course, better would be to try and get these changes in to the `binutils` upstream, but here is easier to get a "quick fix" in place.

### DIFF
--- a/packages/binutils/2.35.1/0011-mips-fix-missing-emulations.patch
+++ b/packages/binutils/2.35.1/0011-mips-fix-missing-emulations.patch
@@ -1,0 +1,13 @@
+--- a/ld/ldmain.c	2021-01-09 16:47:26.109063089 +0000
++++ b/ld/ldmain.c	2021-01-09 16:44:00.426128832 +0000
+@@ -700,6 +700,9 @@
+ 		   || strcmp (argv[i], "-mips5") == 0
+ 		   || strcmp (argv[i], "-mips32") == 0
+ 		   || strcmp (argv[i], "-mips32r2") == 0
++		   || strcmp (argv[i], "-mips32r3") == 0
++		   || strcmp (argv[i], "-mips32r4") == 0
++		   || strcmp (argv[i], "-mips32r5") == 0
+ 		   || strcmp (argv[i], "-mips32r6") == 0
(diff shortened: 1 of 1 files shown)


cpackham:
I'd like a bit more info on the flags gcc is passing. Why is it passing them on? What is the effect of them being passed? Is there a disconnect between certain versions of binutils and newer gcc? My preference would be an upstream patch for binutils or gcc which we carry for existing versions.

majenkotech:
I have no idea quite why it passes them on - GCC is very much a mystery to me.  As it says in the `ld` source code (`ldmain.c`):

```c
 710         {
 711           /* FIXME: The arguments -mips1, -mips2, -mips3, etc. are
 712          passed to the linker by some MIPS compilers.  They
 713          generally tell the linker to use a slightly different
 714          library path.  Perhaps someday these should be
 715          implemented as emulations; until then, we just ignore
 716          the arguments and hope that nobody ever creates
 717          emulations named ips1, ips2 or ips3.  */
 718         }
```

And there's similar thing for `-m486` too. 

I have raised it upstream (https://sourceware.org/bugzilla/show_bug.cgi?id=27167) already, but of course it would need them to then release a new version of binutils and you guys to then include that in ct-nt, etc.

TBH, though, since submitting this patch I now have a set of other patches to GCC for configuring multilib for a specific family of MIPS based microcontrollers, so it rather looks like I'll be running my own fork for that anyway, so this patch could just reside there until the upstream is fixed.  Feel free to close this if that's simpler.

majenkotech:
Looks like my upstream patch has been approved and merged into the main source tree, which is nice. That solves the problem for future, and I'm happy to run my own fork anyway for now, so I guess this is now redundent.

cpackham:
If it's upstream I'm happy to carry it in ct-ng. Could you update this pr
with the patch from upstream (i.e. run git format-patch so it has the
metadata of the commit upstream).

PR #1444, opened 2021-01-12
Picolibc companion

keith-packard:
Finishes making picolibc a pure companion library, while also adding C++ support from upstream gcc and then merging picolibc 1.5 support.


cpackham (approved):
Looks good to me. I'll wait for the builds to finish before merging the PR

PR #1447, opened 2021-01-14
Build newlib-nano as a companion library

keith-packard:
This uses the same process as for the picolibc companion library. One tricky part is how to share the same sources as are used for newlib; the proposed patches allow the user to select a different version for newlib and newlib-nano. If the same version is selected, then the source files will be shared. To ensure that all patches are shared, newlib-nano has a symbolic link to the newlib package version directory.

I've tested this by creating an arm toolchain including both picolibc and newlib-nano and then running zephyr's test suite (twister), which builds against a handful of arm architectures and executes some of those using qemu.


cpackham (approved):

cpackham:
@keith-packard these look good to me. You mentioned on IRC that you were having some issues with g++ and multiple libcs. Did you still want this PR merged? We can follow up in subsequent PRs.

keith-packard:
yes, these patches work fine for C apps, and "mostly" work for C++ apps; there are some C++ headers in gcc that are a bit broken, but we can sort those out with subsequent patches.

abrodkin:
@keith-packard I just realized there's one funny corner-case with that approach. If you try to link something with some existing `.specs` which do rename of `-lc` to `-lc_nano` that way, see https://www.sourceware.org/git/?p=newlib-cygwin.git;a=blob;f=libgloss/arm/elf-rdimon.specs:
```
*rdimon_libc:
%{!specs=nano.specs:-lc} %{specs=nano.specs:-lc_nano}

...

*link_gcc_c_sequence:
%(rdimon_link_gcc_c_sequence) --start-group %G %(rdimon_libc) %(rdimon_libgloss) --end-group
```
it won't work :)

```
$ arm-bare_newlib_cortex_m3_nommu-eabi-gcc test.c --specs=rdimon.specs

$ arm-bare_newlib_cortex_m3_nommu-eabi-gcc test.c --specs=rdimon.specs --specs=nano.specs
.../x-tools/arm-bare_newlib_cortex_m3_nommu-eabi/arm-bare_newlib_cortex_m3_nommu-eabi/bin/ld.real: cannot find -lc_nano
.../x-tools/arm-bare_newlib_cortex_m3_nommu-eabi/arm-bare_newlib_cortex_m3_nommu-eabi/bin/ld.real: cannot find -lrdimon_nano
```

One not very elegant solution might be in creation of `*_nano.a` symlincs so that LD is happy, but in theory not all libs may need their suffixed versions.

P.S. And it's not only ARM, ARC is also affected due to the same constructions in .specs files :(

keith-packard:
Keeping things compatible with the old method for using two C libraries is probably a good plan; if a simple symlink for libc (and libm, I assume) is sufficient, that seems like a fine solution.

PR #1448, opened 2021-01-15
Add PRU target configuration

dinuxbg:


cpackham (commented):
I'd prefer if we had tarballs to build from by default. That way the content can be independently verified (and we can be compromised with a $5 wrench).

cpackham:
On Wed, Jan 20, 2021, 07:16 dinuxbg <notifications@github.com> wrote:

> I added a commit to switch from GIT tag to a tarball.
>
> I also added small fix for a linker issue I found at the last moment.
>
> Let me know if I should squash and resubmit.
>
Yes please squash the fixes into the appropriate commit(s) and push -f to
update the PR.

cpackham:
@dinuxbg thanks for the PR and the quick update. Sorry it took me a while to get round to merging it.

PR #1450, opened 2021-01-25
Libstdc++ compat

keith-packard:
Ok, I've managed to find a work-around to having multiple C libraries installed in the toolchain.

The root cause was C++ headers using '#include_next' to find C library header files. Those would always use the 'default' C library header files because those were after the C++ directories in the search path. Because we're building the whole toolchain, we don't need #include_next as we can make things work correctly without them.

The key patch in this series just does that -- switches some of the #include_next references to simple #include so that you get the first headers from the pre-pended C lib directory rather than some random headers from the middle of the include search path.

The other two patches bring picolibc to version 1.5.1 and then add a couple of patches on top of that to make the C++ stuff work.

I've tested this by hacking zephyr to use picolibc for all of the tests, which now pass on the qemu ARM platforms (except for cmsis tests, which I have no idea how to get running).


cpackham (commented):
I'm not sure about this one. I've read the docs on #include_next and I think I get it's use-case and I could see how it could affect supporting multiple libcs.

What I'm not sure about is the consequences of such a change. It sounds like we could hit other unintended side-effects when building some exotic projects that use odd filenames.

I'm also not terribly keen on carrying around patches in ct-ng if they're not going to land upstream. Although to be fair there are others in the category of stuff in ct-ng that never went upstream (and this is much better documented than "Apply Android-related modifications").

keith-packard:
That's very reasonable. The other option I considered was to install *none* of the libc variants in the default search path. That remains tricky because of #include_next, but I will explore that and see how far I get.

I can also ask upstream and see if the gcc folks have any suggestions.

cpackham:
On Tue, Feb 2, 2021, 07:27 Keith Packard <notifications@github.com> wrote:

> That's very reasonable. The other option I considered was to install
> *none* of the libc variants in the default search path. That remains
> tricky because of #include_next, but I will explore that and see how far I
> get.
>
> I can also ask upstream and see if the gcc folks have any suggestions.
>
Thanks. I'm inclined to merge this while you're exploring other options.
That way at least the feature gets some exposure. Thoughts?

keith-packard:
Let me explore a 'no default' libc option and see if I can't make that work without gcc patches. That will avoid any risk of breaking existing c++ users. Given how fragile the header files are in that environment, I have to agree with you that any changes are fraught with peril.

cpackham (approved):

cpackham (commented):
Is a similar change needed for gcc-11.1.0 or is this a back-port?

cpackham:
@keith-packard I'm pretty happy with these. Are they good to go from your end?

keith-packard:
> Is a similar change needed for gcc-11.1.0 or is this a back-port?

It's a back-port, I got the changes upstream to gcc before bringing them here.

keith-packard:
> @keith-packard I'm pretty happy with these. Are they good to go from your end?

Hrm. Maybe I should do a 1.6.2 picolibc release that has the patch included so ct-ng doesn't need to carry that separately? Would be useful for other people too. How about I do that and update this patch to use it?

keith-packard:
> Hrm. Maybe I should do a 1.6.2 picolibc release that has the patch included so ct-ng doesn't need to carry that separately? 

I've done a picolibc 1.6.2 release and also added the libstdc++ header patch for gcc version 11.1.0. Things built fine for me, and the toolchain appears to work.

cpackham (approved):

PR #1452, opened 2021-02-02
Add GCC patch for ms_printf to GCC 10.2.0

melven:
Patch from https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c51f1e7427e6a5ae2a6d82b5a790df77a3adc99a

Fixes cross-native compilation errors in gomp for mingw-w64 and threads=posix of the form:
```
error: unknown conversion type character 'l' in format [-Werror=format=]
```


cpackham (commented):
Could you add your sign-off to the commit. Other than that looks good to me

cpackham:
Commit lacks sign-off, PR would need updating for gcc 10.3 and 11.0 (assuming it's not already there).

PR #1456, opened 2021-02-13
CI: Add miltilibbed Elf32 & glibc configs for ARC

abrodkin:
As of today baremetal (AKA "Elf32") & Linux glibc toolchains are even more important than Linux uClibc one for ARC, so adding them.

@cpackham if +2 more full multilib toolchains for just one architecture is way too much given amount of resources we have for per-PR testing here, I may drop existing uClibc config.

PR #1455, opened 2021-02-12
Add binutils 2.36.1 and glibc 2.33

egtvedt:

PR #1457, opened 2021-02-13
glibc: Add ARC700 support to v2.32 & 2.33

abrodkin:
That fixes build of `arc-multilib-linux-gnu` sample for ARC700. We didn't see that problem previously since we used off-the-tree patch for all ARC processors with glibc pre-2.32 (where ARCv2 port was accepted), since proposed here fix was a part of that patch. Now fix this with ARC700-related diff as compared to vanilla sources.

As of today ARCompact (AKA "ARCv1 ISA") processors (mostly those are ARC770) are not officially supported by upstream glibc as it adds quite some burden on release ans support of yet another "architecture" port.

But given on ABI and ISA front ARCompact is very close to ARCv2 we may easily retrofit its support off-the-tree, which we do here. Thanks to @naheb for taking care of that here: https://github.com/openwrt/openwrt/commit/33646a51abcf15ff5c5363848287e1ed778b7467.

Also given amount of changes we need to apply there's a hope it will be easy and straight-forward to apply the same to later versions of glibc.

Shoudl fix CI failure in https://github.com/crosstool-ng/crosstool-ng/pull/1456.

### DIFF
--- a/packages/glibc/2.32/0002-Add-ARC700-support.patch
+++ b/packages/glibc/2.32/0002-Add-ARC700-support.patch
@@ -1,0 +1,82 @@
+From 6349ae7c3d96c8d00179e290d1ccf8a2d8438cc8 Mon Sep 17 00:00:00 2001
+From: Rosen Penev <rosenp@gmail.com>
+Date: Sat, 13 Feb 2021 17:08:21 +0300
+Subject: [PATCH] Add ARC700 support
+
+glibc does not officially support ARC700 so this adds the missing
+pieces. I looked at uClibc-ng and a patch by Synopsis for glibc.
+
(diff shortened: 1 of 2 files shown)


cpackham:
As discussed now that I've merged #1455 this needs updating. Not sure if we want glib 2.32 and glib2.33 support for ARC700. I'm not too fussed but only the latter will get regular CI builds.

abrodkin:
@cpackham added v2.33 support (with exactly the same patch as for 2.32).
As for the CI - that's totally fine to get only 2.33 covered, though given v2.32 support is there for no extra efforts why not just keep it there in case we ever want to get back to v2.32 for whatever reason.

cpackham:
No problem. I agree it doesn't do any harm to keep the 2.32 support. I'll
take a look at the updated PR tomorrow.

cpackham (approved):

PR #1465, opened 2021-03-07
Add musl 1.2.2

dimkr:


packages/musl/1.2.2/chksum
```diff
@@ -0,0 +1,4 @@
+aed8ae9e2b0898151b36a204088292dd  musl-1.2.2.tar.gz
```
cpackham, line 1: These should be in the format <algo> <filename> <hash>

i.e.
`md5 musl-1.2.2.tar.gz aed8ae9e2b0898151b36a204088292dd`

cpackham (approved):

cpackham:
@dimkr  would it be possible to you to rebase on master and do one more force-push. I've just added a CI config for musl so it'd be good to check that it still works with the new version.

cpackham:
This failed building strace

```
Error:     /home/runner/work/crosstool-ng/crosstool-ng/.build/arm-unknown-linux-musleabi/src/strace/xlat/inet_protocols.h:239:1: error: static assertion failed: "IPPROTO_MAX != 256"
Error:     make[3]: *** [libstrace_a-net.o] Error 1
Error:     make[2]: *** [all-recursive] Error 1
Error:     make[1]: *** [all] Error 2
```

I haven't dug into the error yet

cpackham:
Looks like strace/kernel-headers/musl causes some problems. Any time we touch one of those the musl build fails.

cpackham:
@dimkr I think the kernel-headers/strace issues are sorted now. If you rebase this series against master it should be good to go (I've tested doing so locally).

PR #1484, opened 2021-04-12
gdb: Don't mess with gdbserver config permissions

abrodkin:
Note, this is more of precautious "fix", though I'd say pretty meaningful: why touching something we're not going to really use.

Now, w/o that fix it's not possible to build even host/target GDB out of GDB v10.x sources. To get gdbserver built in v10.x we'll
need to do much more (see https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=919adfe8409211c726c1d05b47ca59890ee648f1) as source structure was changed (gdbserver was "promoted" to the top-level, see https://github.com/abrodkin/crosstool-ng/commit/2f9b431228c58fe98704a72390672942935ac8e3).

Yet host or target GDB could be nicely built of v10.x sources just with that one-liner fix.

### DIFF
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -8,11 +8,6 @@ do_debug_gdb_get()
 do_debug_gdb_extract()
 {
     CT_ExtractPatch GDB
-
-    # Workaround for bad versions, where the configure
-    # script for gdbserver is not executable...
-    # Bah, GNU folks strike again... :-(
-    chmod a+x "${CT_SRC_DIR}/gdb/gdb/gdbserver/configure"
 }
 
 do_debug_gdb_build()


scripts/build/debug/300-gdb.sh
```diff
@@ -9,10 +9,12 @@ do_debug_gdb_extract()
...
 
-    # Workaround for bad versions, where the configure
-    # script for gdbserver is not executable...
-    # Bah, GNU folks strike again... :-(
-    chmod a+x "${CT_SRC_DIR}/gdb/gdb/gdbserver/configure"
+    if [ "${CT_GDB_GDBSERVER}" = "y" ]; then
+        # Workaround for bad versions, where the configure
+        # script for gdbserver is not executable...
+        # Bah, GNU folks strike again... :-(
+        chmod a+x "${CT_SRC_DIR}/gdb/gdb/gdbserver/configure"
```
cpackham, line 16: Does anyone know which gdb versions this applies to? Perhaps the best approach would be to remove this entirely if it doesn't affect any of the gdb versions we support.

Also perhaps we should drop some gdb versions, maybe keep latest 7,x, 8,x, 9.x. That would mean we need to worry less about compiling old versions when we get round to adding 10.x

abrodkin: Wow, so that fix was introduced back in 2007 as a part of initial addition of debug facilities (see https://github.com/crosstool-ng/crosstool-ng/commit/8d3f0a8781cc25e75db3db4a9195816e7d3da902).
So yeah, I think it might not be really needed any longer.
FWIW I cannot find the same fix in Buildroot v2009.02 (which was the first release in their git, see https://git.buildroot.net/buildroot/tree/toolchain/gdb/gdb.mk?h=2009.02) so maybe it was long obsolete.

Maybe @yann-morin-1998 may remember something? ;)

cpackham: I'm tempted to say lets remove it and see if anyone complains. Particularly if this is the only thing stopping us supporting gdb 10.1

abrodkin: Well for GDB 10.x we'll need more things to do, see:
1. https://git.buildroot.net/buildroot/commit/?id=81ef52b6306651559fa269b34ef79b1e42141f98
2. https://git.buildroot.net/buildroot/commit/?id=39267b5b1e19971b0e0e3bc06820a53b31c5eadf

yann-morin-1998: > Wow, so that fix was introduced back in 2007 as a part of initial addition of debug facilities (see [8d3f0a8](https://github.com/crosstool-ng/crosstool-ng/commit/8d3f0a8781cc25e75db3db4a9195816e7d3da902)).
> So yeah, I think it might not be really needed any longer.
> FWIW I cannot find the same fix in Buildroot v2009.02 (which was the first release in their git, see https://git.buildroot.net/buildroot/tree/toolchain/gdb/gdb.mk?h=2009.02) so maybe it was long obsolete.
> 
> Maybe @yann-morin-1998 may remember something? ;)

14 years later, a gray-beard wakes up from a long sleep, scratches his head...

So yeah, I do remember that part: the configure script did not always have the executable bit set. I looked at all versions from tha era that were known to ct-ng: 6.2.1, 6.3, 6.4, 6.5, and 6.6.

However, in August 2011, GNU re-released all the gdb tarballs (as well as those of quite a fe other packages) because of a license oversight: https://sourceware.org/legacy-ml/gdb/2011-09/msg00002.html

It is quite possible that they also fixed (incidentally) the execute bit for that script...

> FWIW I cannot find the same fix in Buildroot v2009.02 (which was the first release in their git, see https://git.buildroot.net/buildroot/tree/toolchain/gdb/gdb.mk?h=2009.02) so maybe it was long obsolete.

But it was there, way back from before we had the first release:

https://git.buildroohttps://git.buildroot.net/buildroot/tree/toolchain/gdb/gdb.mk?id=2d523c23175b8e8772c2e223d92513f836cde8a6#n29

So, that was for gdb 5.3...

There, we have our answer... Guess we can drop that workarond, now! ;-)

abrodkin: Thanks for the historical overview, @yann-morin-1998!
Posted a re-spin with removal of that old hack.

Issue #1491, opened 2021-04-15
Companion Newlib nano libs make built toolchain non-relocatable


abrodkin:
With https://github.com/crosstool-ng/crosstool-ng/pull/1447 we may now build & install Newlib (and in fact `libstdc++` as well) optimized by size along with the "normal" Newlib. But the way it is done now make toolchain non-usable if it was moved to another location, or we may call it non-relocatable.

That's because we hard-code paths in the top-level `nano.specs` that way:
```bash
    cat > "${CT_SYSROOT_DIR}/lib/nano.specs" <<EOF
%rename link	newlib_nano_link
%rename cpp	newlib_nano_cpp
%rename cc1plus	newlib_nano_cc1plus
*cpp:
-isystem ${CT_PREFIX_DIR}/newlib-nano/${CT_TARGET}/include %(newlib_nano_cpp)
*cc1plus:
-idirafter ${CT_PREFIX_DIR}/newlib-nano/${CT_TARGET}/include %(newlib_nano_cc1plus)  
*link:
-L${CT_PREFIX_DIR}/newlib-nano/${CT_TARGET}/lib/%M -L${CT_PREFIX_DIR}/newlib-nano/${CT_TARGET}/lib
EOF
```

Obviously `${CT_PREFIX_DIR}` & `${CT_TARGET}` get resolved in build-time and we get in the resulting `nano.specs` something like:
```
*cpp:
-isystem ...MY_CTNG_BUILD_LOCATION.../arc-multilib-elf32/newlib-nano/arc-multilib-elf/include %(newlib_nano_cpp)

*cc1plus:
-idirafter ...MY_CTNG_BUILD_LOCATION.../arc-multilib-elf32/newlib-nano/arc-multilib-elf/include %(newlib_nano_cc1plus)··

*link:
-L...MY_CTNG_BUILD_LOCATION.../arc-multilib-elf32/newlib-nano/arc-multilib-elf/lib/%M -L...MY_CTNG_BUILD_LOCATION.../arc-multilib-elf32/newlib-nano/arc-multilib-elf/lib
```

If we move toolchain, those paths become irrelevant and linker fails to find what's needed -> fail :(

If we want to keep Newlib's "nano" version separately the following hack with run-time path calculation seem to work.
Though I'm not sure if there're any side-effects we might get, so @keith-packard @cpackham your input here is very welcome!

```patch
[...]

PR #1495, opened 2021-04-16
newlib-nano: Use run-time calculated paths in top-level nano.specs

abrodkin:
In currently generated top-level `nano.specs` we resolve paths during toolchain building and then use those pre-defined
full paths once the toolchain got built.

That's OK until the toolchain is used right where it was built, otherwise paths used in the top-level `nano.specs` become irrelevant and linker fails to find "nano" libs reverting to non-"nano" libs in the default location.

See https://github.com/crosstool-ng/crosstool-ng/issues/1491.

### DIFF
--- a/scripts/build/companion_libs/350-newlib_nano.sh
+++ b/scripts/build/companion_libs/350-newlib_nano.sh
@@ -199,13 +199,13 @@ ENABLE_TARGET_OPTSPACE:target-optspace
 %rename cc1plus	newlib_nano_cc1plus
 
 *cpp:
--isystem ${CT_PREFIX_DIR}/newlib-nano/${CT_TARGET}/include %(newlib_nano_cpp)
+-isystem %:getenv(GCC_EXEC_PREFIX ../../newlib-nano/${CT_TARGET}/include) %(newlib_nano_cpp)
 
 *cc1plus:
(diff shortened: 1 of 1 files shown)


scripts/build/companion_libs/350-newlib_nano.sh
```diff
@@ -199,13 +199,13 @@ ENABLE_TARGET_OPTSPACE:target-optspace
 %rename cc1plus	newlib_nano_cc1plus
 
 *cpp:
--isystem ${CT_PREFIX_DIR}/newlib-nano/${CT_TARGET}/include %(newlib_nano_cpp)
+-isystem %:getenv(GCC_EXEC_PREFIX ../../newlib-nano/${CT_TARGET}/include) %(newlib_nano_cpp)
 
 *cc1plus:
--idirafter ${CT_PREFIX_DIR}/newlib-nano/${CT_TARGET}/include %(newlib_nano_cc1plus)  
+-idirafter %:getenv(GCC_EXEC_PREFIX ../../newlib-nano/${CT_TARGET}/include) %(newlib_nano_cc1plus)
```
cpackham, line 205: This had a minor conflict with a commit from @galak that removed the trailing space. I think I resolved it correctly.

cpackham (approved):

nolange:
I ran into this problem, does not seem to be fixed.

I use gcc 8.5.0, I suppose it's only a problem for compilers lacking a certain commit.

Issue #1499, opened 2021-04-28
Add support for GCC 11.1


mustafakemalgilor:
GCC 11.1 is finally released yesterday. It would be good to be able to use it via crosstool-ng.

cpackham:
I've merged the work from @nikkon-dev (thanks). Feel free to take it for a spin.

PR #1501, opened 2021-04-28
Initial commit for gcc11

nikkon-dev:
Initial commit for #1499

Brought and checked some patches from gcc10 for x86_64 and ARM.
Checked just x86_64 cross build on Linux with kernel 3.16 and glibc 2.12.

Did not check or validate MacOS/Windows/MinGW/etc builds.
Did not check targets other than x86_64.

Signed-off-by: Nik Konyuchenko <spaun2002mobile@gmail.com>


cpackham (commented):
Typo in commit message "MaxOS" should be "MacOS"

*sigh* github can't let me comment on commit messages. The typo is in the commit message of "Pull patch for macOS from GCC 10.3.0 package"

cpackham:
Thanks for the PR. I gave it a quick skim.

I think perhaps we could squash the `-Werror` related changes but Ideally we wouldn't need them. I think some of it might be due to the fact that we're still doing more passes than necessary (see #808). I also wonder if there is an update to DUMA that we could use instead of patching it.

nikkon-dev:
I squashed the -Werror related changes.
The Glibc sources have logic to disable warnings for gcc11, but it looks like -Werror or --enable-error have higher priority. 

>I also wonder if there is an update to DUMA that we could use instead of patching it.

The original DUMA project on SourceForge that is used for builds looks abandoned. The last change was in 2009.
I found a couple of mirrors on GitHub, but both have the same issue I patched.
[hayguen/duma](https://github.com/hayguen/duma) - the author of this one seems to be one of the original DUMA authors.
[johnsonjh/duma](https://github.com/johnsonjh/duma)

Floessie:
I've been trying this branch at 0cf01b5 for `powerpc-e500v2-linux-gnu`. Some random outcomes:

- glibc <2.30 fails to build
- glibc 2.31's `scanf()` won't match anything (even with locales in place and working – really strange…)
- glibc 2.33 is fine

HTH,
Flössie

nikkon-dev:
@Floessie, could you check if you have `-mfloat-gprs=double` in your CFLAGS and try to build without that if you do?

WBR,
Nik

Floessie:
@nikkon-dev No, it's nowhere to be found in my `.config`. You're alluding to "glibc <2.30 fails to build", I suppose...

graysky2:
@nikkon-dev - Any free time to refresh for gcc 11.2 that was just released?

QBos07:
> @nikkon-dev - Any free time to refresh for gcc 11.2 that was just released?

I would do it in the next week

graysky2:
@QBos07 - Thanks, how to determine whether or not extra patches are needed?  Is it as simple as bumping the version numbers and checksums?

nikkon-dev:
> 
> 
> @QBos07 - Thanks, how to determine whether or not extra patches are needed? Is it as simple as bumping the version numbers and checksums?

Most likely, the answer is yes for the GCC 11.1 to 11.2. In general, in addition to the Github CI validation tests, I build GCC with ancient GLlibc versions for x86_64/aarch64/ppc64le architectures.

>@nikkon-dev - Any free time to refresh for gcc 11.2 that was just released?
>I would do it in the next week
@QBos07, that would be great! I will not be able to do that earlier.

graysky2:
@nikkon-dev @QBos07 - I followed the commit adding gcc11 as a guide: https://github.com/graysky2/crosstool-ng/commit/7106561d29b43453e1733aa78baed9a70dc8ee4c

Building now.  I did see that is pulling the correct source.  If it is successful, I will submit a PR.  Any feedback?

```
...
[EXTRA]    Retrieving 'gcc-11.2.0'
[EXTRA]    Verifying SHA512 checksum for 'gcc-11.2.0.tar.xz'
[EXTRA]    Saving 'gcc-11.2.0.tar.xz' to local storage
...
```

QBos07:
> Any feedback?
Do you test if patches nessesairy and if there is any incompatibility or new dependency? (e.g. looking into source code)
If you done it, I think it's fine

PR #1513, opened 2021-05-19
Add config sample of Raspberry Pi 4

Bensuperpc:
Add config sample of Raspberry Pi 4:
Add armv8-rpi4-linux-gnueabihf
Add aarch64-rpi4-linux-gnu

Signed-off-by: Bensuperpc <bensuperpc@gmail.com>


Bensuperpc:
Is it relevant to add armv8-rpi4-linux-gnueabihf or aarch64-rpi4-linux-gnu and to CI tests?

cpackham:
> Is it relevant to add armv8-rpi4-linux-gnueabihf or aarch64-rpi4-linux-gnu and to CI tests ?

We already have aarch64 and a few 32-bit arms covered so I don't see it really being necessary.  Perhaps if we had some follow on tests that actually used the resulting toolchains, but as that's not currently the case it'll just take up CI resourses.

cpackham (approved):

Bensuperpc:
> > Is it relevant to add armv8-rpi4-linux-gnueabihf or aarch64-rpi4-linux-gnu and to CI tests ?
> 
> We already have aarch64 and a few 32-bit arms covered so I don't see it really being necessary. Perhaps if we had some follow on tests that actually used the resulting toolchains, but as that's not currently the case it'll just take up CI resourses.

It is especially that there are examples for the raspberry pi 1,2 and 3, but not for the version 4 ^^

PR #1517, opened 2021-05-24
Update utils to latest stable

graysky2:
* autoconf from 2.69 to 2.71
* isl from 0.22 to 0.24

Signed-off-by: John Audia <graysky@archlinux.us>


packages/autoconf/2.71/0001-Update-autoconf-and-add-new-version.patch
```diff
@@ -0,0 +1,35 @@
+From a09e46a5413066896e7a18c0032b5e2da35df1f9 Mon Sep 17 00:00:00 2001
```
cpackham, line 1: This file seems to have been added in error

cpackham (changes requested):
I'd also prefer real names in the Sign-off line.

packages/autoconf/2.71/0001-Update-autoconf-and-add-new-version.patch
graysky2: Fixed both.

cpackham:
Looks better thanks

cpackham (approved):

PR #1523, opened 2021-06-06
gcc: Add CT_CC_GCC_TM_CLONE_REGISTRY config

stephanosio:
This commit adds a new gcc config `CT_CC_GCC_TM_CLONE_REGISTRY` that
enables the GCC transactional memory clone registry feature for libgcc.

Note that the gcc option to control this feature is only available in
gcc 10 and above.
(see gcc commit 5a4602805eb3ebddbc935b102481e63bffc7c5e6)


scripts/build/cc/gcc.sh
```diff
@@ -396,6 +396,12 @@ do_gcc_core_backend() {
         extra_config+=("--disable-__cxa_atexit")
     fi
 
+    if [ "${CT_CC_GCC_TM_CLONE_REGISTRY}" = "y" ]; then
+        extra_config+=("--enable-tm-clone-registry")
+    else
+        extra_config+=("--disable-tm-clone-registry")
+    fi
```
cpackham, line 403: I'm wondering if we should just omit the else case.  I presume tm-clone-registry defaults to disabled in gcc.

cpackham: I'd presumed wrong. So we do need to disable this. Older GCCs will just see an option they don't recognize (that will emit a warning but should be otherwise harmless).

cpackham (approved):

PR #1527, opened 2021-06-09
Move SRC_CUSTOM inside choice

cpackham:
Prior to commit 30bffa96 SRC_CUSTOM was inside the choice with
SRC_RELEASE and SRC_DEVEL making the 3 options mutually exclusive.
Restore this behaviour keeping the "fix" to keep kconfig happy.

Fixes #1151
Fixes: 30bffa96 ("don't nest choices")


cpackham:
@bagasme, @bensuperpc care to take this for a spin?

bensuperpc:
looks good to me, i will test this when i have more time

PR #1552, opened 2021-07-08
Bionic : disable large files when unsupported

abraaocsantana:
Fixes https://github.com/crosstool-ng/crosstool-ng/issues/1553

Building with bionic as the libc gives me an error of undefined symbols for fgetpos and fsetpos.
This is caused by setting large file support on Api Levels < 24.

Disabling large file support fixes the error.

The patches are from my personal fork of gcc, it's not in the main repository (it may seem like it due to the commit style).

### DIFF
--- a/packages/gcc/10.3.0/0024-android-Disables-large-files-when-unsupported.patch
+++ b/packages/gcc/10.3.0/0024-android-Disables-large-files-when-unsupported.patch
@@ -1,0 +1,35 @@
+From 769efb97889e45df2601f4351259c127d5e9dfbb Mon Sep 17 00:00:00 2001
+From: Abraao de Santana <abraaocsantana@gmail.com>
+Date: Thu, 8 Jul 2021 15:26:04 -0300
+Subject: [PATCH] [android] Disables large files when unsupported
+
+libstdc++-v3/config/os/bionic
+        * os_defines.h: Undefines _FILE_OFFSET_BITS when it's 64-bit and Api Level < 24
+---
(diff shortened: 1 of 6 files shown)


cpackham:
Has this patch been sent upstream to gcc? I'm not a fan of carrying patches that aren't back-ports of upstream fixes.

abraaocsantana:
Hey, @cpackham it has not, sorry! As mentioned in the PR's definition https://github.com/crosstool-ng/crosstool-ng/pull/1552#issue-686245298 .

The fix is simple enough, in the seemingly appropriate place, and executes what's asked in the bionic doc: https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md

But yes, it's from my fork of GCC and has not been passed through GCC's regression testing yet because I lack the free time to it right now (that's why I didn't submit the patch yet).

So feel free to refuse it if it's not up to quality, no hard feelings!
It's just that I'm using those patches on my fork of Crosstool-ng and felt like sharing them right away.

cpackham:
I don't have a problem with the patches they certainly seem reasonable to me.However I do think submitting them to upstream gcc will be better for everyone (still happy to carry them as back-ports in ct-ng).

abraaocsantana:
Seems fair!
I really don't have the time right now, but I'll do it asap.

abraaocsantana:
Patch sent: https://gcc.gnu.org/pipermail/gcc-patches/2021-July/575046.html
I don't know if the GCC community will have an interest in reviewing Android patchs, though.

cpackham:
since we've now dropped support for bionic this change is no longer relevant

PR #1558, opened 2021-07-12
Openrisc

cpackham:


cpackham:
Needs testing. Probably shouldn't be merged until someone confirms it's actually working

cpackham:
Nothing has been done with this since I posted it so I'm going to close the PR without merging. It can be re-opened if someone surfaces that is willing to test openrisc.

johnsonjh:
@cpackham Just wanted to confirm that this works for me, at least for musl-libc built binaries.

PR #1560, opened 2021-07-13
Support common local patch directory

stephanosio:
This commit updates the patching process such that the local patches
can be applied in a version-independent manner, as with the patches
provided by the crosstool-ng packages.

This is done by reading the patch files from
`${CT_LOCAL_PATCH_DIR}/(package_name)` rather than from
`${CT_LOCAL_PATCH_DIR}/(package_name)/(version)`.

### DIFF
--- a/scripts/functions
+++ b/scripts/functions
@@ -2240,6 +2240,7 @@ CT_DoExtractPatch()
     local bundled_patch_dir
     local bundled_common_patch_dir
     local local_patch_dir
+    local local_common_patch_dir
     local overlay
 
     # Inherit global value if requested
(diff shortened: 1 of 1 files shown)


cpackham (approved):

abrodkin:
@stephanosio, @falaleevms made a similar improvements to solve the same problem back in the day via https://github.com/crosstool-ng/crosstool-ng/pull/1385. So do we need yet another fix?
I mean I'm fine with having it, but just to make sure we don't introduce multiple substances implementing the same functionality w/o good reason.

stephanosio:
@abrodkin This is not adding the same fumctionality as the referenced PR.

PR #1570, opened 2021-08-03
packages/glibc: Add 2.34

cpackham:
Add glibc 2.34. Bring through patches for canadian build and ARC700.

https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html


cpackham:
The x86_64 build has a weird error when building gdbserver (a.k.a. native gdb).

````
[INFO ]  Installing native gdb
[EXTRA]    Configuring native gdb
[EXTRA]    Building native gdb
[ERROR]    build-gnulib-gdbserver/import/unistd.h:135:3: error: #error "Please include config.h first."
[ERROR]    build-gnulib-gdbserver/import/unistd.h:137:1: error: '_GL_INLINE_HEADER_BEGIN' does not name a type
[ERROR]    build-gnulib-gdbserver/import/unistd.h:1929:1: error: '_GL_INLINE_HEADER_END' does not name a type
[ERROR]    make[1]: *** [Makefile:635: amd64-linux-siginfo.o] Error 1
[ERROR]    make[1]: *** Waiting for unfinished jobs....
[ERROR]
````

The error doesn't happen with glibc-2.33 but darned if I can figure out what the problem is. There must be something subtly different between which is causing gcc to pick up a different file (or to compile something I didn't need to previously). I know some of the changes to glibc-2.34 involve making some libraries part of the main C library so there are some subtle differences in what `./configure` does.

cpackham:
Some more analysis of the problem https://sourceware.org/pipermail/libc-help/2021-August/005975.html basically a problem with the gnulib bundled with gdb.

PR #1584, opened 2021-08-22
contrib: Add contrib/gen-chksum.sh

dinuxbg:
When updating a single package, it does not make sense to touch the
configuration for all packages:
  $ ./maintainer/manage-packages.sh -D

Instead of this, provide a script to update the checksum for a single
tarball only:

  $ ./contrib/gen-chksum.sh gnuprumcu-0.6.0.tar.gz > packages/gnuprumcu/0.6.0/chksum

Signed-off-by: Dimitar Dimitrov <dimitar@dinux.eu>

### DIFF
--- a/contrib/gen-chksum.sh
+++ b/contrib/gen-chksum.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Helper script to generate package chksum from a given tarball.
+
+die()
+{
+  echo "ERROR: $@"
+  exit 1
+}
+
+[ $# != 1 ] && die "Usage: $0 <package.tar.gz>"
+
+for s in md5 sha1 sha256 sha512
+do
+  echo "${s} `basename ${1}` `${s}sum ${1} | cut -f1 -d' '`"
+done


bensuperpc:
I had made a python version, if you are interested:

```
import sys
import glob
import hashlib

filenames = glob.glob(sys.argv[0])
filenames.reverse()

def hash_file(fname, hash_type):
    hash_result = getattr(hashlib, hash_type)()
    with open(fname, "rb") as f:
        for chunk in iter(lambda: f.read(2 ** 20), b""):
            hash_result.update(chunk)
    return hash_result.hexdigest()

for filename in filenames:
    hash_list = ['md5', 'sha1', 'sha256', 'sha512']
    for hash_type in hash_list:
        print(hash_type, filename, hash_file(filename, hash_type))
```

dinuxbg:
I just found out that manage-packages.sh  already supports both of the suggested enhancements. Example:

```
$ ./maintainer/manage-packages.sh --digest --select m4
```

It recreates packages/m4/*/chksum for all tarball compression formats listed by the package.

Sorry for the noise - closing since feature is already present.

PR #1595, opened 2021-09-11
gdb10: Backported fixes & improvements for ARC

abrodkin:
Upstream GDB v10.2 lacks some important fixes & improvements for ARC processors as those were submitted late for v10 (good example here is support of native debugging on Linux) as well as there's non-upstreamable fix for compatibility with ARC proprietary toolchain.

And here we back-port quite a few patches from what would be v11.1 at some point to the current version so that our users may enjoy those now. And with v11.1 release majority of those will be gone as they are already in upstream `master branch` (see links in each patch's commit message).

And while we're here also we refresh existing patches for GDB 10.2 with `./maintainer/manage-packages.sh --update-patches --select gdb-10.2`.


cpackham (changes requested):
I think there is a legitimate failure in aarch64 (and possibly architectures) from the CI builds

```
Error:     /home/runner/work/crosstool-ng/crosstool-ng/.build/aarch64-unknown-linux-gnu/src/gdb/gdb/aarch64-tdep.c:3563:9: error: invalid conversion from 'bool (*)(gdbarch*, gdb_byte, dwarf2_frame_state*)' {aka 'bool (*)(gdbarch*, unsigned char, dwarf2_frame_state*)'} to 'bool (*)(gdbarch*, gdb_byte, dwarf2_frame_state*, const gdb_byte**, const gdb_byte*)' {aka 'bool (*)(gdbarch*, unsigned char, dwarf2_frame_state*, const unsigned char**, const unsigned char*)'} [-fpermissive]
```

packages/gdb/10.2/0000-musl_fix.patch
```diff
@@ -24,7 +24,7 @@
  /* This comment documents high-level logic of this file.
  
  Waiting for events in sync mode
---- a/gdb/stopcode.h
+--- /dev/null
 +++ b/gdb/stopcode.h
 @@ -0,0 +1,4 @@
 +#ifndef W_STOPCODE
```
cpackham, line 30: Not a problem with your change just re-generating the patch. But I get the feeling that this patch is no-longer needed in newish gdb `W_STOPCODE` is defined this way in `gdbsupport/gdb_wait.h` so we can probably just drop the patch entirely. I'm getting a new PC so I can't dig into why this change was first needed at the moment.

abrodkin: Good catch!

But let me start from an interesting story about the second hunk of the same patch:
```diff
+#ifndef __SIGRTMIN
+#define __SIGRTMIN SIGRTMIN
+#endif
```

[...Putting an archeologist hat on yet again...]

In CT-NG the first appearance of this entire patch seems to happen together with GDB 8.0.0 introduction made by @stilor, see https://github.com/crosstool-ng/crosstool-ng/commit/72a5a1d6de43503579cb77daacdf17a77a315c03. Ans since then it was re-added for all the subsequent versions exactly in the way I did it just now.

Unfortunately there's no information on reasons for this patch to exist in this commit. Maybe @stilor remembers something 4 years ago...

But as often OpenEmbedded is to rescue - we may see something very similar done by @kraj here: https://github.com/openembedded/openembedded-core/commit/87376746d64aea47cce327d951692c31d4acf872. What's also good there's a reference to a glibc's bug touching this problem: https://sourceware.org/bugzilla/show_bug.cgi?id=13012.

And aforementioned fix became a part of 7.11 back in Feb 2016 via https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=682b25469e66ea45b214e95962671373983c118f. So this hunk in no longer needed.

Now back to the third one you're asking about:
```diff
--- a/gdb/stopcode.h
+++ b/gdb/stopcode.h
@@ -0,0 +1,4 @@
+#ifndef W_STOPCODE
+#define W_STOPCODE(sig) ((sig) << 8 | 0x7f)
+#endif
```

This one was fixes the same day 5 minutes before the one discussed above via https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=963843d4d07aef6caa296dacf191f8adc9518596 and also made its way in GDB 7.11.0.

That said this entire patch is no longer needed and should be dropped, which I'll do on a re-spin (there seem to be a failure for `aarch64`, see https://github.com/crosstool-ng/crosstool-ng/pull/1595/checks?check_run_id=3574676717#step:9:356, which might be due to my other patches). Another question is if we want that patch to be removed from older GDB version as well?

P.S. And while at it, do we really want that entire zoo of GDB versions?

cpackham: > [...Putting an archeologist hat on yet again...]

(aside I wonder if there is an Indiana Jones emoji)

> 
> In CT-NG the first appearance of this entire patch seems to happen together with GDB 8.0.0 introduction made by @stilor, see [72a5a1d](https://github.com/crosstool-ng/crosstool-ng/commit/72a5a1d6de43503579cb77daacdf17a77a315c03). Ans since then it was re-added for all the subsequent versions exactly in the way I did it just now.
> 
> Unfortunately there's no information on reasons for this patch to exist in this commit. Maybe @stilor remembers something 4 years ago...

My own digging leads me to c7b9a9bd which appears to be adding musl support prior to upstream gdb releasing a version that was musl aware.

<snip>

> That said this entire patch is no longer needed and should be dropped,

Yeah looks like nothing newer than 7.11 needs the musl patch.

>  Another question is if we want that patch to be removed from older GDB version as well?

I think at least if we remove it from the newest version (maybe 9.2 as well) it'll disappear over time. I note gdb 11.1 was just released. So after the dust settles on this series we should look at adding that.

> P.S. And while at it, do we really want that entire zoo of GDB versions?

I'm not sure I've quite gone from care-taker merger of patches to setting direction for the project, but I agree it's a bit silly to have as many as we do.

I think having just the newest 7.x, 8.x, 9.x and 10.x (and 11.x) would be more than enough, and I don't think it'd be unreasonable to drop the extras. I do wonder if there's some obscure micro that only has support in one of those older versions but it's not as if they are receiving any testing (even building) right now.

Looking at those a bit closer. Most of them are already marked obsolete so I'm less concerned about just dropping them. I'll get a patch ready for that but I'll wait until this lands before submitting it.

abrodkin:
@cpackham in this re-spin:
1. Dropped `0000-musl_fix.patch` as discussed here above & re-generated remaining patches
2. Dropped `0010-dwarf2-Add-arguments-to-gdbarch_execute_dwarf_cfa_ve.patch` & `0011-arc-Ignore-Metaware-extension-.cfa_info.patch` since the former touches generic code and changes signature of `default_execute_dwarf_cfa_vendor_op()` and the latter only updates ARC part, leaving all the rest as they were. That caused build failures for non-ARC toolchains. Until we wrap all the change with "ifdefs" we'll keep it change locally. And refreshed patchset again to have all in the right order etc.

cpackham (approved):

PR #1600, opened 2021-09-15
GCC 11.2: Fix cross-native/cross-canadian builds

abrodkin:
With this we may finally build Windows and "native" toolchains if host tools are also GCC11 based. For example:
1. You build cross toolchain with all the recent components by CT-NG
2. You build cross-canadian toolchain for Windows or ARC, ARM, whatever board

See upstream bug report https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100017 for more details.

Basically when we do cross-canadian build with use of the same GCC11 as a "host" compiler we're seeing an error like that:
```
mingw-w64-cross/gcc/x86_64-w64-mingw32/libstdc++-v3/include/fenv.h:58:11: error: 'fenv_t' has not been declared in '::'
   58 |   using ::fenv_t;
```

This is a solution proposed by Yujie Yang in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100017#c20

Note, though it's not the final fix merged upstream, that's just an attempt to fix this by casual GCC users. There's a hope it will be fixed anyways a bit later, maybe by the time of GCC 11.3...

And while at it, refreshing existing GCC11 patches.


cpackham (approved):

cpackham (approved):

abrodkin:
For the record, there's a proper upstream fix for this bug, see https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=4fde88e5dd152fe866a97b12e0f8229970d15cb3. I guess we'd like it to be applied instead of our temporary fix. Needs to be done via separate PR though.

PR #1606, opened 2021-09-28
gcc: Build static libgcc in core_pass1

cpackham:
Per https://github.com/crosstool-ng/crosstool-ng/issues/808 build static
libgcc in the first pass which lets us skip the second one.

Very lightly tested at this point. Creating PR to make use of CI builds.

### DIFF
--- a/config/cc.in
+++ b/config/cc.in
@@ -2,11 +2,6 @@
 
 menu "C compiler"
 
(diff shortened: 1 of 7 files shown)


cpackham:
mingw-w64 doesn't appear to like this.

PR #1608, opened 2021-09-28
Update crosstool-NG.sh due to recent coreutils change

b1f6c1c4:
Since GNU coreutils 9.0, trying to change the mode of any symbolic link will cause chmod to exit with status 1.
See https://github.com/coreutils/coreutils/commit/bbb19b18687d5572dcbe7bbb1a48c8ee0f21786b for the change.

The built toolchain usually contains symbolic link so, without the `|| true` trick, `chmod` will cause failure to `ct-ng build` at the very last stage that is extremely hard to spot.


cpackham:
The idea seems reasonable (I'll have to have a think about it because we're basically swallowing all errors not just this specific one). I would like to see something in the commit message (the info you've put in the PR is a good start but PRs only exist on github whereas commit messages are part of the repository and are available regardless of where it is hosted).

I'd also like to see a proper Sign-off line with a real name and email address (i.e. `git commit --amend -s`) as part of the [DCO](https://developercertificate.org/). It's also a sign that you're a real person not a spammer trying to land a bitcoin miner in the CI workflow.

scripts/crosstool-NG.sh
```diff
@@ -733,11 +733,15 @@ if [ "${CT_LOG_TO_FILE}" = "y" ]; then
     fi
 fi
 if [ "${CT_PREFIX_DIR_RO}" = "y" ]; then
-    chmod -R a-w "${CT_PREFIX_DIR}"
+    # Since GNU coreutils 9.0, trying to change the mode of any symbolic link will cause chmod to exit with status 1.
+    # https://github.com/coreutils/coreutils/commit/bbb19b18687d5572dcbe7bbb1a48c8ee0f21786b
```
cpackham, line 737: I don't think this comment adds anything. If anyone wants to know the reason for this change they can look at the commit message.

egtvedt:
@cpackham coreutils are aware of the regression, and apparently the behavior will change in upcoming coreutils releases:
https://www.mail-archive.com/bug-coreutils@gnu.org/msg32108.html
https://www.mail-archive.com/bug-coreutils@gnu.org/msg32112.html

b1f6c1c4:
The bug has been fixed from coreutils but not yet released. https://github.com/coreutils/coreutils/commit/e8b56ebd536e82b15542a00c888109471936bfda

PR #1626, opened 2021-11-03
cloog: remove gnu mirror

cpackham:
The GNU mirror doesn't contain cloog-18.4. Remove it.


cpackham:
@galak the macOS makers seem unhappy. Some searching suggests something might need to be done with brew but I really have no idea (and no access to a macOS system).

cpackham:
Explicitly installing make seems to avoid the problem https://github.com/crosstool-ng/crosstool-ng/pull/1629. I wonder if its a generic problem with using a make we build. I'll see if I can find a sufficiently old Linux docker image to try it out with.

PR #1638, opened 2021-11-25
Fix GCC 10.3.0 build on MacOS 12

jackqu7:
MacOS 12 (Xcode 13) fails to build GCC due to an issue with genconditions. This occurs on both x86_64 and arm64 platforms.

This has been discussed upstream and someone proposed a patch which fixes the issue: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92061

This is particularly important for ARM (M1) Macs, as 10.3.0 is the only GCC version with other required patches to build on this architecture.

Signed-off-by: Jack Saunders <jackqu7@gmail.com>

### DIFF
--- a/packages/gcc/10.3.0/0024-genconditions-macos-12.patch
+++ b/packages/gcc/10.3.0/0024-genconditions-macos-12.patch
@@ -1,0 +1,39 @@
+Fix build on MacOS 12 (Xcode 13)
+see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92061 for details
+
+---
+ gcc/genconditions.c | 9 +++++----
+ 1 file changed, 5 insertions(+), 4 deletions(-)
+
+--- a/gcc/genconditions.c
++++ b/gcc/genconditions.c
+@@ -57,8 +57,9 @@ write_header (void)
+ \n\
+ /* It is necessary, but not entirely safe, to include the headers below\n\
+    in a generator program.  As a defensive measure, don't do so when the\n\
+-   table isn't going to have anything in it.  */\n\
+-#if GCC_VERSION >= 3001\n\
++   table isn't going to have anything in it.\n\
++   Clang 9 is buggy and doesn't handle __builtin_constant_p correctly.  */\n\
++#if GCC_VERSION >= 3001 && __clang_major__ < 9\n\
+ \n\
+ /* Do not allow checking to confuse the issue.  */\n\
+ #undef CHECKING_P\n\
+@@ -170,7 +171,7 @@ struct c_test\n\
+    vary at run time.  It works in 3.0.1 and later; 3.0 only when not\n\
+    optimizing.  */\n\
+ \n\
+-#if GCC_VERSION >= 3001\n\
++#if GCC_VERSION >= 3001 && __clang_major__ < 9\n\
+ static const struct c_test insn_conditions[] = {\n");
+
+   traverse_c_tests (write_one_condition, 0);
+@@ -191,7 +192,7 @@ write_writer (void)
+ 	"  unsigned int i;\n"
+         "  const char *p;\n"
+         "  puts (\"(define_conditions [\");\n"
+-	"#if GCC_VERSION >= 3001\n"
++	"#if GCC_VERSION >= 3001 && __clang_major__ < 9\n"
+ 	"  for (i = 0; i < ARRAY_SIZE (insn_conditions); i++)\n"
+ 	"    {\n"
+ 	"      printf (\"  (%d \\\"\", insn_conditions[i].value);\n"


cpackham:
I assume this affects more than just GCC 10.3.0. Have you check other versions of GCC?
 
I notice the upstream bug report is ~2years old and says "RESOLVED INVALID". Has it been superseded by something else or did they just black-list certain versions of clang?

I'm not overly keen on ct-ng carrying patches that aren't just back-ports from upstream. Particularly if there's little chance they'll ever land. If anything I'd prefer to add a configure check and refuse to use know affected versions of clang.

RyanHir:
I was never able to find why the issue was considered resolved, but I never had this issue with x86_64 macOS. But I am attempting to build a toolchain with the M1 as its host (using Xcode 13 on an Intel mac) and the issue comes back.

This patch appears to resolve the issue for GCC 7 through GCC 10. GCC 11 does not appear to have any issues. I can't explain why builds are failing for @jackqu7 when targeting Intel.

I find it especially odd that the patch appears to work because it should be no-op with Xcode 13 as it is conditional with the version of Clang.

I tested this patch in a sister project to crosstool-ng which can be found [here](https://github.com/RyanHir/toolchain-builder). The only reason I could believe there being a difference in Intel builds is because I am removing the kernel version from the GNU triple of the host (`x86_64-apple-darwin21.4.0` ->`x86_64-apple-darwin`) and setting `MACOSX_DEPLOYMENT_TARGET` to `10.9`.

cpackham:
Looks like gcc11 has the same code (genconditions.c has been renamed genconditions.cc) but the content is mostly the same so if this is the problem then gcc11 is probably affected.

I'd still like to know why the GCC bug is marked as RESOLVED INVALID. There is some mention of blacklisting certain clang versions but I can't see how they do this. I'd much prefer to carry a cherry-pick/back-port from upstream gcc.

bhundven:
TL;DR: Not a gcc bug, a clang/llvm bug, this should be ported to other gcc versions. The gcc bug is marked RESOLVED INVALID because it's a clang/llvm bug, and someone should open a llvm bugzilla report to get it fixed in the future.

As the gcc bug [says](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92061#c1), it's a bug in clang/llvm.
The patch they provided was a "workaround" for that version of clang/llvm (clang/llvm 9).

Further on in the bug, someone mentions that it is still an issue with xcode 13 (clang/llvm 13), and mentioned that the patch still worked for gcc up to 11.3.0.

To get this patch to 12.1.0 and newer, I forward ported the patch:
```diff
--- a/gcc/genconditions.cc	2022-06-26 20:15:07.357605102 -0700
+++ b/gcc/genconditions.cc	2022-06-26 20:16:18.589252259 -0700
@@ -58,7 +58,7 @@
 /* It is necessary, but not entirely safe, to include the headers below\n\
    in a generator program.  As a defensive measure, don't do so when the\n\
    table isn't going to have anything in it.  */\n\
-#if GCC_VERSION >= 3001\n\
+#if GCC_VERSION >= 3001 && __clang_major__ < 9\n\
 \n\
 /* Do not allow checking to confuse the issue.  */\n\
 #undef CHECKING_P\n\
@@ -170,7 +170,7 @@
    vary at run time.  It works in 3.0.1 and later; 3.0 only when not\n\
    optimizing.  */\n\
 \n\
-#if GCC_VERSION >= 3001\n\
+#if GCC_VERSION >= 3001 && __clang_major__ < 9\n\
 static const struct c_test insn_conditions[] = {\n");
 
   traverse_c_tests (write_one_condition, 0);
@@ -191,7 +191,7 @@
 	"  unsigned int i;\n"
         "  const char *p;\n"
         "  puts (\"(define_conditions [\");\n"
-	"#if GCC_VERSION >= 3001\n"
+	"#if GCC_VERSION >= 3001 && __clang_major__ < 9\n"
 	"  for (i = 0; i < ARRAY_SIZE (insn_conditions); i++)\n"
 	"    {\n"
 	"      printf (\"  (%d \\\"\", insn_conditions[i].value);\n"
```

Please update this PR with the patch for all crosstool-ng supported gcc versions.
I'm running into another issue with gLibc... which I think is not related.

bhundven (changes requested):
Please provide this patch for all crosstool-ng supported versions of gcc.

bhundven:
So, interesting note. I have clang/llvm 13.1.6, and didn't have this problem, without this patch.

cpackham:
FYI GCC 10.4.0 is out. I intend to update ct-ng to this when I get a chance (probably this weekend) I'm not sure if anything in the release will fix this specific issue but at the very least it'll mean any patch might need regenerating.

cpackham:
GCC 10.5 has now been added as well. I'm closing this PR for now as it definitely needs updating but the bigger issue was really that the problem being solved wasn't made clear.

PR #1642, opened 2021-12-06
Update config.{sub,guess}

gnaygnil:
Add support for loongarch64 architecture hosts

Signed-off-by: gnaygnil <gnaygnil@gmail.com>

### DIFF
--- a/scripts/config.guess
+++ b/scripts/config.guess
@@ -996,6 +996,9 @@ EOF
     k1om:Linux:*:*)
 	echo "$UNAME_MACHINE"-unknown-linux-"$LIBC"
 	exit ;;
+    loongarch64:Linux:*:*)
+	echo "$UNAME_MACHINE"-unknown-linux-"$LIBC"
+	exit ;;
     m32r*:Linux:*:*)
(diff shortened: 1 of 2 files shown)


cpackham:
@gnaygnil I see loongarch support has landed upstream http://git.savannah.gnu.org/gitweb/?p=config.git;a=commit;h=c8ddc8472f8efcadafc1ef53ca1d863415fddd5f

I think we'd be better of running `ct-ng updatetools` to pick up the change (and others). I'll post a patch shortly it'd be good if you could try it out an verify that it works for you.

gnaygnil:
> @gnaygnil I see loongarch support has landed upstream http://git.savannah.gnu.org/gitweb/?p=config.git;a=commit;h=c8ddc8472f8efcadafc1ef53ca1d863415fddd5f
> 
> I think we'd be better of running `ct-ng updatetools` to pick up the change (and others). I'll post a patch shortly it'd be good if you could try it out an verify that it works for you.

It has been verified that using "ct-ng updatetools" works very well. Thanks!

PR #1644, opened 2021-12-12
Update config.{sub,guess}

cpackham:
Run `ct-ng updatetools` to pick up the latest config.{sub,guess} from
upstream. This picks up support for some new architectures (e.g.
loongarch) and some new variants of existing ones. There is some
refactoring that makes the diff a bit larger but it's fairly easy to
follow.


cpackham:
@gnaygnil care to take this for a spin.

PR #1652, opened 2022-01-12
gcc: Use upstream fix for cross-canadian builds

cpackham:
Replace the patch from #1600 with the change that landed upstream.


cpackham:
@abrodkin I don't think the CI builds cover the cross-native/cross-canadian scenario. Are you able to take this for a spin (also given I've just merged #1606 canadian builds may be impacted)

abrodkin:
@cpackham wow, that was fast! I'm going to try that now locally.

Note though there's the same patch cherry-picked for 11.3 release, see https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=01a70ccd723eb9a479186fe37c972b0d0f8676cf. And while technically it's the same, we may want to have a back-ported one here so that it applies better (not tried yet) and has more relevant meta-data... though I'm obviously nitpicking here ;)

As for the CI part, we discussed that with @temap internally on how that could be done in the most efficient way, as I'd really like it to be in place - otherwise this and more issues will eventually slip through the cracks and will be only found much later. And at the same time canadian cross is not that rare these days since Linux build machines are cheap and simple, while we want to produce toolchains for Windows/Mac etc... that said if you're not against adding of yet another build target, we may think how to implement it.

abrodkin:
@cpackham successfully built "native" tools for ARC HS processors via canadian cross process from your `native-cross-canadian` branch. I guess it's good to merge now then. Again, thanks for your work!

temap:
Hi @abrodkin and @cpackham,
I added Canadian Cross testing with x86_64-w64-mingw32,x86_64-pc-linux-gnu sample. Could you please review the change?

https://github.com/crosstool-ng/crosstool-ng/pull/1681/files

PR #1662, opened 2022-02-02
binutils-2.37: add more patches from upstream release branch

egtvedt:
Fetch in various patches from binutils-2_37-branch upstream.

The most vital change is the
0012-pr28391-strip-objcopy-preserve-dates-a-cannot-set-time.patch which
allows building large upstream projects like Qt WebEngine without need
100k's of file descriptors open.

### DIFF
--- a/packages/binutils/2.37/0011-bfd-close-the-file-descriptor-if-there-is-no-archive-fd.patch
+++ b/packages/binutils/2.37/0011-bfd-close-the-file-descriptor-if-there-is-no-archive-fd.patch
@@ -1,0 +1,232 @@
+From 1c611b40e6bfc8029bff7696814330b5bc0ee5c0 Mon Sep 17 00:00:00 2001
+From: "H.J. Lu" <hjl.tools@gmail.com>
+Date: Mon, 26 Jul 2021 05:59:55 -0700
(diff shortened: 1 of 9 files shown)


cpackham (approved):

egtvedt:
All these patches will be fine to have, they are already present in upstream binutils.

However, turned out I still could not use binutils 2.37 to link Qt's WebEngine :/ I had to switch back to binutils 2.36.1, that worked fine.

cpackham:
I noticed that there were even more patches upstream than the ones you brought in. Is it worth bringing in the rest?

egtvedt:
https://sourceware.org/bugzilla/show_bug.cgi?id=28138 AFAICT this is the relevant upstream bug.

The bug mentions two patches:
`bfd: Close the file descriptor if there is no archive fd`
and
`bfd: Set error to bfd_error_malformed_archive only if unset`

Only the former is backported to binutils-2.37 release branch, might need the latter as well. I will give that a go tomorrow hopefully.

egtvedt:
> I noticed that there were even more patches upstream than the ones you brought in. Is it worth bringing in the rest?

There was quite a lot of patches related to ppc. I no longer have powerpc architecture available :/ So nothing I can test other than patch is applying.

PR #1672, opened 2022-02-09
packages/binutils: add binutils-2.38

egtvedt:
Binutils version 2.38 has been released
https://www.phoronix.com/scan.php?page=news_item&px=GNU-Binutils-2.38
https://sourceware.org/pipermail/binutils/2022-February/119721.html

Patches 0000 to 0008 from binutils-2.37 applied clean, so forward ported
them.

The following patches are now applied from upstream, hence dropped:
0010-change-uint-to-unsigned.patch
0011-bfd-close-the-file-descriptor-if-there-is-no-archive-fd.patch
0012-pr28391-strip-objcopy-preserve-dates-a-cannot-set-time.patch
0013-ld-pru-fix-resource_table-output-section-alignment.patch
0014-pr28422-build_id-use-after-free.patch
0015-pr28459-readelf-issues-bogus-warning.patch
0016-pr28417-std-string-no-longer-allows-accepting-nullptr_t.patch
0017-pr28540-segmentation-fault-on-NULL-byte_get.patch
0018-gold-place-note-gnu-property-section-before-other-note-sections.patch
0019-waddress-warning-in-ldelf.c.patch


cpackham:
> @cpackham: where can I find the failure logs? Seems like sparc-unknown-linux-gnu fails to build on macos-10.15, but I can not find any information in the log archive?

When they're captured they show up under the artifacts link on for the run (e.g. https://github.com/crosstool-ng/crosstool-ng/actions/runs/1819740433#artifacts) in that particular case looks like the macos runners just died (as they have a tendency to do). As you got some runs through for other architectures on macos I'm not too worried.

stilor:
When you bring new versions, can you please run the packages through `./maintainer/manage-packages.sh -D -P -s binutils-2.38` (in this case, or select the appropriate package/version as needed). This makes sure the patches not just apply cleanly, but do so without fuzz/offsets (which makes further upgrades easier) and that the checksums are generated in a consistent format and do not have typos in them. I just posted a pull request that does such clean-ups on a lot of packages.

egtvedt:
> When you bring new versions, can you please run the packages through `./maintainer/manage-packages.sh -D -P -s binutils-2.38` (in this case, or select the appropriate package/version as needed). This makes sure the patches not just apply cleanly, but do so without fuzz/offsets (which makes further upgrades easier) and that the checksums are generated in a consistent format and do not have typos in them. I just posted a pull request that does such clean-ups on a lot of packages.

Will do the script thing next time, I'll redo this as well and force push an update.

I have automated the checksum part locally on my end from way back, but everything is better with scripts committed to source.

Issue #1477, opened 2021-04-01
Cut new release (last release from 2019)


paulmenzel:
The last/current release 1.24.0 is from 2019. For projects depending on releases, it’d be great if a new release could be made. Just doing time based releases with no promises would help most of the projects.

cpackham:
I have been thinking about cutting a release but I need to sort out some details first. I'd also really like to get some traction on the 2 pass gcc build (#808) and make that the big new feature.

abrodkin:
@galak just FYI if you'd like something special to be included in the discussed release.

Bensuperpc:
It would be a good idea to do a new version, GCC11 support would be perfect for dockcross ^^

qhaas:
I have been using 1.24.0 since it is the latest tagged (a general rule I follow for most FOSS projects), is most of the community just using the head of master these days?

bensuperpc:
> I have been using 1.24.0 since it is the latest tagged (a general rule I follow for most FOSS projects), is most of the community just using the head of master these days?

Some things are not available with this version, I am using the master branch for this reason, but not the last commit, to avoid some problems.
It's better to use a stable version but not too much choice on my projects :/

benesch:
@cpackham, does #1625 move the needle here at all? The current stable release of crosstool-ng is broken out of the box since the isl tarballs moved. I know it wouldn't be a very exciting release without the 2-pass GCC build, but I'd personally take a functional stable release over an exciting one. :)

paulmenzel:
For the record: https://github.com/crosstool-ng/crosstool-ng/releases/tag/crosstool-ng-1.25.0

PR #1674, opened 2022-02-11
Updates to make `ct-ng build-all` pass

stilor:
I am still re-running build-all after the latest rebase and fixing the last two samples, but barring any surprises, it should be now passing at least on my Ubuntu 21.10.

Finish incomplete removal of obsolete milestones. Finish the conversion of GCC build into two-pass (the "libc_headers" step mentioned in that commit is actually needed to unbreak xtensa samples).

Also, update/fix docker images that can be used to run the build in various Linux flavors. Randomly tested with a single sample in each, I hope to resurrect my dedicated build machine for running `ct-ng build-all` repeatedly.


cpackham (approved):

cpackham:
@stilor looks good to me. Let me know if you want me to do merge the PR.

stilor:
Just to confirm, `ct-ng build-all` passed in the commit I posted. Hope new binutils don't break anything :)

cpackham:
I think it should be fine. The differences between 2.37 and 2.38 aren't huge and both series had clean CI runs.

Are there any configurations you think are worth adding to the CI builds? We don't have any coverage for bionic so that's an obvious one. I'd also like to include a canadian build but I haven't quite figured out the best way to go about that.

mbulut:
@stilor This breaks the toolchain build for me on a debian bullseye host.
In particular, it fails in the "Install C Library" step, saying that it could not find the target gcc.
Revision ba680a3 does not have that issue.
The toolchain config is attached.
[armv8-rpi3-linux-gnueabihf.txt](https://github.com/crosstool-ng/crosstool-ng/files/8064877/armv8-rpi3-linux-gnueabihf.txt)

messense:
> In particular, it fails in the "Install C Library" step, saying that it could not find the target gc

Same issue, https://github.com/messense/homebrew-macos-cross-toolchains/actions/runs/1840835602

stilor:
@mbulut  @messense You guys didn't run `ct-ng oldconfig` (or `ct-ng menuconfig`) after the update, did you? Some internal configuration variables have been changed due to unification of compiler passes, and your .configs have the _old_ variables in them. This is the reason the pass to build the target compiler was not done before ct-ng attempted to build the C library.

cpackham:
@mbulut, @messense the problem with your failures is that the configs you are using need updating (your configs will say `CT_CC_CORE_PASS_1_NEEDED=y` but after these changes it needs to say `-CT_CC_CORE_NEEDED=y`). I was able to successfully build after running @mbulut 's config through `ct-ng oldconfig`.

stilor:
@cpackham I actually think we should mark bionic obsolete. They dropped support for gcc some 3-4 years ago, which is why we have a quite old version of bionic in ct-ng. The patches to support bionic have not been ported to GCC11 (which is why I had to limit GCC selection for it to GCC10 and below). Frankly, I am not eager to port them - they have been a real pain to maintain in the past, as they are very invasive.

We should also mark obsolete glibc 2.12.1 (CentOS6 has reached EOL, I didn't want to obsolete it before that) and by extension, the whole glibc-ports package (of which 2.12.1 is the last remaining version). Maybe you have some other candidates for obsoletion, too?

After marking these obsolete, we should release 1.25.0 and remove the obsolete packages afterwards.

How does this plan sound?

cpackham:
After I posted that I went and actually looked at it and now I'm all in favour of dropping bionic. It's probably not terribly useful to people wanting to develop for modern android-ndk and having it in ct-ng gives the impression that it's usable which it probably isn't. It also downloads binary images from the interwebs which is probably a bad idea.

Your plan sounds pretty good. Let's move the conversation over to #1477 (or a new issue if you want to start a fresh thread).

PR #1676, opened 2022-02-14
strace: Add 5.16

cpackham:
https://lists.strace.io/pipermail/strace-devel/2022-January/010884.html


cpackham:
@stilor how do you feel about the number of strace versions. Given that strace tries to maintain backwards compatibility there doesn't seem to be a reason to keep as many versions as we do (I know I'm the one who has added the most recent ones).

cpackham:
I'll follow up with another PR. Basically what I propose is any version added between 1.24 and now (except for the most recent) I'll just remove. Older versions present in the 1.24 release I'll mark as obsolete and we can drop them after 1.25.

PR #1681, opened 2022-02-16
CI: add Canadian Cross build job

temap:
Add reusable GitHub workflow to build Canadian Cross toolchain where
the x86_64-w64-mingw32 cross-compiler is used which is built with ct-ng.

Signed-off-by: Artem Panfilov <artemp@synopsys.com>

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -89,97 +89,40 @@ jobs:
 
   toolchains:
     needs: [ crosstool, tarballs ]
-    runs-on: ${{ matrix.host }}
-    strategy:
-      matrix:
-        host: [ "ubuntu-latest", "macos-10.15" ]
-        sample:
-          [
-            "aarch64-unknown-linux-gnu",
-            "arc-multilib-elf32",
-            "arc-multilib-linux-gnu",
-            "arc-multilib-linux-uclibc",
-            "arm-picolibc-eabi",
-            "arm-unknown-linux-gnueabi",
-            "arm-unknown-linux-musleabi",
-            "armv6-nommu-linux-uclibcgnueabi",
-            "avr",
-            "i686-w64-mingw32",
-            "mips-unknown-elf",
-            "mips-unknown-linux-gnu",
-            "mips64-unknown-linux-gnu",
-            "powerpc-unknown-linux-gnu",
-            "powerpc64-unknown-linux-gnu",
-            "riscv32-unknown-elf",
-            "riscv64-unknown-elf",
-            "s390-unknown-linux-gnu",
-            "sh-unknown-elf",
-            "sparc-unknown-linux-gnu",
-            "x86_64-multilib-linux-uclibc",
-            "x86_64-unknown-linux-gnu",
-            "x86_64-w64-mingw32",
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -9,10 +9,7 @@ jobs:
     runs-on: ${{ matrix.host }}
     strategy:
       matrix:
-        host: [
-          "ubuntu-latest",
-          "macos-10.15",
-        ]
+        host: ["ubuntu-latest", "macos-10.15"]
```
abrodkin, line 12: I'd move that beautification to a separate commit to not pollute real meat of the functional change.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -54,9 +51,7 @@ jobs:
     runs-on: ${{ matrix.host }}
     strategy:
       matrix:
-        host: [
-          "ubuntu-latest",
-        ]
+        host: ["ubuntu-latest"]
```
abrodkin, line 54: Ditto

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -122,71 +113,16 @@ jobs:
           "s390-unknown-linux-gnu",
           "sh-unknown-elf",
           "sparc-unknown-linux-gnu",
-          "x86_64-unknown-linux-gnu",
           "x86_64-multilib-linux-uclibc",
+          "x86_64-unknown-linux-gnu",
```
abrodkin, line 117: Ditto

abrodkin (approved):
I'd say it's a long awaited feature as we discussed that previously - testing of canadian cross tools is good to have. And now when it's possible to re-use parts of the existing flow, implementation is very simple and clean.

The only concern I have it's a limitation of GitHub CI for non-paid accounts so that we cannot see this check in action - lots of jobs are pending still after many hours. But w/o funding what else can we do?

cpackham:
I certainly like the look of this. We might be able to leverage azure, circle-ci, gitlab-ci to spread the load a bit. I am surprised no-one from GitHub has complained about us chewing up resources.

abrodkin:
> I am surprised no-one from GitHub has complained about us chewing up resources.

Well I guess the answer is simple - we consume all the given for free resources and then we're being put on halt. So they don't see resources consumption higher that they expect or allow us to use. But given what we do here is a true open source development and results of our works are being used in some other projects, maybe it worth contacting GitHub folks asking for more resources?

.github/workflows/continuous-integration-workflow.yml
temap: OK, I will revert code changes, not related to PR.

abrodkin: There's no need to revert, just move them to a separate patch to be applied before the bigger functional change in the series.

temap:
@cpackham, all checks passed. Do you have any concerns related to this PR except @abrodkin comments?

cpackham:
No concerns from me. If you update to address @abrodkin's points I'll take
a look this weekend.

cpackham (approved):

PR #1686, opened 2022-02-21
gcc: add gcc libstdcxx-verbose option

nolange:
Rather important option for arm cortex toolchains supporting c++,
avoids pulling in all printf/iostream code by default.


QBos07:
Why do you use a tristate? Is bool not more accurate in this case?

QBos07:
Looks good

cpackham (approved):

PR #1695, opened 2022-03-25
Bugfix/gcc11 cxx fix anon aggregated struct

antmak:
Fixes

```c++
struct A {
    int a;
    int aa;
};

struct B {
    int b;
};

struct C {
    int c;
    union {
        struct A ma;
        struct B mb;
    };
};

int main() {

    struct C c = { .c = 2, .ma = { .a = 3, .aa = 4, }, };

    return 0;
}
```

```
<source>:20:56: error: 'C::<unnamed union>' has no non-static data member named 'a'
```

https://godbolt.org/z/3W5oso1GM

the regression appeared only GCC 11.2.0. All details in patches' description. 

I hope the crosstool-ng can move to gcc 11.3.0 soon (after gcc release...), because the version 11.2.0 is buggy IMHO


egtvedt (approved):
Thank you for identifying problem and backporting.

cpackham (approved):

cpackham (approved):

PR #1705, opened 2022-04-02
Add sample arm-none-eabi

keith-packard:
This is the triplet used by the Arm Embedded Toolkit as well as
Debian.

Signed-off-by: Keith Packard <keithp@keithp.com>


cpackham (approved):

cpackham:
@keith-packard were you hoping to get this in the 1.25 release?

keith-packard:
> @keith-packard were you hoping to get this in the 1.25 release?

Not critical for me -- I use it for testing and can carry it externally as long as you like.

cpackham:
OK I'll tag this for after the release.

keith-packard:
A fine plan. I just realized I'd been carrying that patch for a long time and wondered if you'd find it a useful thing to have upstream at some point. I think it's kinda nice to be able to generate a GNU Arm Embedded Toolkit -ish toolchain.

mid-kid:
How "compatible" is this configuration with the official sources from ARM themselves? I'm having trouble building the upstream toolchain from source since they started using the linaro ABE framework and crosstool-ng has always been a joy to use (and is usually quicker to introduce new compiler versions), so having this as a compatible replacement would be awesome.

Specifically, I'm wondering:
* I had trouble using --specs=nano.specs with the arm-unknown-eabi and my previous attempts at using crosstool-ng for this. Does this configuration provide a similar set of specs files?
* Is picolibc used anywhere by default or does it need a set of flags? I'd rather not have differing behavior compared to the real thing.
* Is the official toolchain multilib? If not, does this install aarch64-none-elf symlinks or do I need to use `-m64? How should I call the compiler in my software's build scripts, such that it works with both?
* What does the `-ftls-model=local-exec` flag do? Would this introduce differing behavior compared to the "real thing"?

keith-packard:
> How "compatible" is this configuration with the official sources from ARM themselves? I'm having trouble building the upstream toolchain from source since they started using the linaro ABE framework and crosstool-ng has always been a joy to use (and is usually quicker to introduce new compiler versions), so having this as a compatible replacement would be awesome.
> 
> Specifically, I'm wondering:
> 
>     * I had trouble using --specs=nano.specs with the arm-unknown-eabi and my previous attempts at using crosstool-ng for this. Does this configuration provide a similar set of specs files?

I don't think this builds a config which is exactly compatible with how the ARM tools work, it does try to build something that supports the same targets using the same triplet.

>     * Is picolibc used anywhere by default or does it need a set of flags? I'd rather not have differing behavior compared to the real thing.

No, picolibc requires the use of `--specs=picolibc.specs`.

>     * Is the official toolchain multilib? If not, does this install aarch64-none-elf symlinks or do I need to use `-m64? How should I call the compiler in my software's build scripts, such that it works with both?

Yes, the official toolchain is multilib, but only for arm32, not aarch64.

>     * What does the `-ftls-model=local-exec` flag do? Would this introduce differing behavior compared to the "real thing"?

This only affects how applications using thread local storage (TLS) are compiled by supporting the standard ARM ABI for TLS.  The ARM version doesn't support TLS at all, so yes, this is a change, but it should be upware compatible as code which works with the ARM toolchain cannot be using TLS at all.

mid-kid:
I see. Thanks a lot!

However, I'm having trouble using `--specs=nano.specs`, as it complains about GCC_EXEC_PREFIX being unset, and setting it to `$CT_PREFIX/newlib-nano/lib/gcc` doesn't work, as then it fails to find `-lgcc`.

Looking at the layout of the official ARM toolchain and comparing it with the ct-ng one, I can see that both have nano.specs in the arm-none-eabi/lib directory, but ARM has libc_nano.a in `arm-none-eabi/lib` while ct-ng has it in `newlib-nano/arm-none-eabi/lib`. Maybe these two directories should be merged, instead?

PR #1710, opened 2022-04-21
gcc: bump GCC 11 to latest release 11.3.0

egtvedt:
Remove patches applied upstream we no longer need to maintain here.


cpackham:
You beat me to it :). Unfortunately timing wise this is just a bit too late for the 1.25 release (which is really a pain). I'll merge this shortly after the release.

Hopefully we can end up with a release cadence that follows gcc. As gcc-12 appears imminent croosstool-ng 1.26 might not quite establish that pattern but hopefully it won't be as long as the gap between 1.24 and 1.25.

cpackham (approved):

egtvedt:
> You beat me to it :). Unfortunately timing wise this is just a bit too late for the 1.25 release (which is really a pain). I'll merge this shortly after the release.

I always build off master branch anyway, so for me personally it is not important to get into the 1.25 release.

IMHO I feel GCC 11.3.0 is quite the bugfix on the previous release, so it would be good to provide in release form in not too distant future. Can we consider doing dotreleases? I would understand if that creates too much additional work.

> Hopefully we can end up with a release cadence that follows gcc. As gcc-12 appears imminent croosstool-ng 1.26 might not quite establish that pattern but hopefully it won't be as long as the gap between 1.24 and 1.25.

Yes, tracking GCC release train is tricky IMHO. The first major release is seldom something you want to use for production system. I have had good experience waiting for the X.2.0 release, and switch to X.3.0 or later as soon as they become available.

bhundven:
Actually,

I'll be getting some time soon. I'm hoping I can contribute more
soon. @cpackham Let's talk soon about automating the releases.

-Bryan

cpackham (approved):

PR #1721, opened 2022-05-09
newlib-nano: Fix include path for CT_NEWLIB_NANO_INSTALL_IN_TARGET

stephanosio:
When `CT_NEWLIB_NANO_INSTALL_IN_TARGET=y`, the `nano.specs` file
emitted by the newlib-nano build script contains an invalid include
path, resulting in the full `newlib.h` being included instead of the
nano `newlib.h` by the application.

`=/include/newlib-nano` is not a valid path (`=` does not mean anything
and that string is taken as an include path as-is) and GCC ignores this
include path, resulting in application including the `newlib.h` from
`include/` which contains the newlib build configurations for the full
newlib.

This commit modifies the newlib-nano build script to emit a proper
newlib-nano include path relative to the `GCC_EXEC_PREFIX`.

NOTE: This `nano.specs` file was originally taken from the GNU Arm Embedded toolchain. Not really sure where they got the `=` from, but I can say for sure that it does not work with the recent GCC versions (GCC 10 and above).

### DIFF
--- a/scripts/build/companion_libs/350-newlib_nano.sh
+++ b/scripts/build/companion_libs/350-newlib_nano.sh
@@ -203,7 +203,7 @@ ENABLE_TARGET_OPTSPACE:target-optspace
 %rename cpp_unique_options  nano_cpp_unique_options
 
 *cpp_unique_options:
--isystem =/include/newlib-nano %(nano_cpp_unique_options)
+-isystem %:getenv(GCC_EXEC_PREFIX ../../${CT_TARGET}/include/newlib-nano) %(nano_cpp_unique_options)
 
 *nano_libc:
 -lc_nano


stephanosio:
cc @galak since you pushed this change initially

cpackham (approved):

cpackham:
looks good to me but lets see if you get any response from the others

keith-packard:
The '=' applies when you configure gcc with a sysroot; the = (or the literal string `$SYSROOT`) is replaced with the current sysroot value. If I could figure out what sysroot actually meant to gcc, it might be a reasonable thing to use with crosstool-ng, but it seems like it has several conflicting meanings in different places. What I want it to mean is that $SYSROOT is where gcc is installed, so that you can use gcc-installation-relative paths to adjust the include search order. And sometimes it does mean that, and othertimes there are additional semantics mixed in which I don't quite understand...

stephanosio:
> The '=' applies when you configure gcc with a sysroot; the = (or the literal string `$SYSROOT`) is replaced with the current sysroot value.

Ok, that clears up what the original intention was. Thanks for clarifying.

> If I could figure out what sysroot actually meant to gcc, it might be a reasonable thing to use with crosstool-ng, but it seems like it has several conflicting meanings in different places.

Yes, it seems to be a mess in general. Another issue I had with the sysroot earlier was that relocating toolchains did not work with it (but this might have been fixed in recent GCC releases) ...

> it's just that most gcc packages are built without this option

Indeed, even the GNU Arm Embedded, where this was taken from, ironically builds without sysroot, so they have the same issue.

> Btw, the use of = or $SYSROOT is documented in the GCC man page

I was looking at the "Spec Files" page for the answer 🤦‍♂️

PR #1753, opened 2022-05-31
gcc: Fix compilation error in core C gcc compiler older than v5

gemaq:
When core gcc older than v5 is compiled it shows the error message:

Build failed in step 'Installing core C gcc compiler'
      called in step '(top-level)'

Error happened in: CT_DoExecLog[scripts/functions@376]
      called from: do_gcc_core_backend[scripts/build/cc/gcc.sh@627]
      called from: do_cc_core[scripts/build/cc/gcc.sh@210]
      called from: main[scripts/crosstool-NG.sh@697]

configure: error: in
`.../build/build-cc-gcc-core/fixincludes':
configure: error: C compiler cannot create executables

This patch disable `all-build-libcpp' target when core gcc
older than v5 is configured.

Signed-off-by: Guillermo E. Martinez <guillermo.e.martinez@oracle.com>

### DIFF
--- a/scripts/build/cc/gcc.sh
+++ b/scripts/build/cc/gcc.sh
@@ -600,6 +600,12 @@ do_gcc_core_backend() {
         --enable-languages="${lang_list}"              \
         "${extra_user_config[@]}"
 
+    gcc_core_build_libcpp=all-build-libcpp
+    # disable target all-build-libcpp in gcc older verions
+    if [ "${CT_GCC_older_than_5}" = "y" ]; then
+        gcc_core_build_libcpp=""
+    fi
+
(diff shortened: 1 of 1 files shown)


gemaq:
I think that it was a regression introduce in db6f703f52e33a

cpackham:
> I think that it was a regression introduce in [db6f703](https://github.com/crosstool-ng/crosstool-ng/commit/db6f703f52e33a5791c5c2728fa1e3a330a08e98)

Yes I think so too. I even appeared to know it only went as far back as GCC 6 but obviously forgot that we do still support configurations with ancient GCC versions.

cpackham (approved):

PR #1754, opened 2022-06-01
Restore Oracle products to crosstool-ng using gcc-oracle

gemaq:
This commit restore the support for the following products
in Oracle Linux 7.9:

Binutils 2.27-44.base.0.400
GCC      4.8.5-44.0.5
glibc    2.17-317.0.3
UEK5/u4  4.14.35-2025.400.8

Sample configurations were updated to use such versions,
and be compiled in host machine using recent GCC versions.

Signed-off-by: Guillermo E. Martinez <guillermo.e.martinez@oracle.com>


stilor:
1. This brings back the versions of the GCC/ISL/CLooG that have been previously obsoleted. It is not enough to just put the versions back into packages/; their dependencies removed by cc6b7fad46f5cb3d84cf87ae47c5f723803a67a3 need to be reinstated as well.
2. The package description for this version of GCC checks out a branch, but the patch needed to build it is tied to a particular git commit hash. If a patch is needed that cannot be picked up upstream for some reason (repository inactive, etc) then pin that version to that specific git revision, as done for elf2flt, for example.

Although @cpackham is doing most of the work now on this project so he gets to make the final decision, I think we shouldn't bring back the obsoleted versions. The best way here, I think, is either a fork with OL support, or if there is a sufficient community interest in this, a branch (possibly created before aa524934f5fcfedb91b03d2fec7a8dd4dd080aea retired the obsolete versions used by these configurations).

cpackham:
I'm not completely anti bringing some of this stuff back but I'd really like to know more about why.

GCC 4.8.5 was marked as obsolete in 2017 (I actually suspect it was obsolete prior to this but 5935d586 was the commit that added the `obsolete="yes"` line). In the 5 years since no-one piped up to complain. To be fair the Oracle interest started relatively recently in 470ce219 so we really should have highlighted back then that 4.8.5 was already obsolete.

What is the motivation here? Even if you want to support using ct-ng on an old-ish Oracle Linux release why does it have to be with such old components?

If ultimately we do bring some of these things back I'd like to have a clear EoL plan for them so we can drop the old versions.

stilor:
To expand on @cpackham 's comment: we have such "compatibility" configurations for various CentOS/Ubuntu releases, but these configurations are pinned to specific kernel/glibc versions with a recent compiler/binutils.

gemaq:
> I'm not completely anti bringing some of this stuff back but I'd really like to know more about why.
happy to hear this :-).

> GCC 4.8.5 was marked as obsolete in 2017 (I actually suspect it was obsolete prior to this but [5935d58](https://github.com/crosstool-ng/crosstool-ng/commit/5935d586e140b35398f45772d5749924dea7da57) was the commit that added the `obsolete="yes"` line). In the 5 years since no-one piped up to complain. To be fair the Oracle interest started relatively recently in [470ce21](https://github.com/crosstool-ng/crosstool-ng/commit/470ce219af2ee28f32caaacc6c568b2bc6d1cc50) so we really should have highlighted back then that 4.8.5 was already obsolete.
> 
> What is the motivation here? Even if you want to support using ct-ng on an old-ish Oracle Linux release why does it have to be with such old components?
The main motivation to build with those "obsolete" packages versions is that Oracle Linux 7.9 repositories support just these version:
https://yum.oracle.com/repo/OracleLinux/OL7/latest/x86_64/index.html
So, Oracle customer wants to cross build application to OL7.
 
> If ultimately we do bring some of these things back I'd like to have a clear EoL plan for them so we can drop the old versions.
ok, understood.
Thanks for your comments

cpackham:
I'm still a little confused by Oracle Linux and some of the restrictions imposed. It appears that you're only able to download source RPMs yet somehow crosstool-ng is OK to download as a tarball?

One clarifying doc I did find was https://www.oracle.com/a/ocom/docs/elsp-lifetime-069338.pdf which says that the EoS date for Oracle Linux 7 is July 2024 so that's a useful thing to record as the date for when these would be removed (assuming we actually do reinstate them).

gemaq:
> I'm still a little confused by Oracle Linux and some of the restrictions imposed. It appears that you're only able to download source RPMs yet somehow crosstool-ng is OK to download as a tarball?

right, OL source packages are in RPM format but toolchain packages (gcc, binutils-gdb and glibc) in which crosstool-ng can rely for OL7 are in https://github.com/oracle repositories, those are up to date by Oracle package maintainers 

> One clarifying doc I did find was https://www.oracle.com/a/ocom/docs/elsp-lifetime-069338.pdf which says that the EoS date for Oracle Linux 7 is July 2024 so that's a useful thing to record as the date for when these would be removed (assuming we actually do reinstate them).

ok.

cpackham:
> those are up to date by Oracle package maintainers

Heh, up to date but at the same time exceptionally out of date :P.

Anyway I'm not really here to criticize Oracle's or their customers policies. I can certainly see an argument for being able to build and develop an application locally and then cross-compile with the same gcc/glibc version once (or fewer times at least).

In terms of getting this PR across the line. I'd like to see the package versions that you are adding back marked as obsolete with a comment in the version.desc about the EoS date for OL7. I've done all the removals I intend to for now and hopefully I'd see the comment and remember post the next release.

Unless @stilor (or anyone else) is strongly opposed to the updated PR I'd be happy to merge it.

gemaq:
>

gemaq:
>

gemaq:
> 
> > those are up to date by Oracle package maintainers
> 
> Heh, up to date but at the same time exceptionally out of date :P.

yes , I know but it is still in use.
 
> Anyway I'm not really here to criticize Oracle's or their customers policies. I can certainly see an argument for being able to build and develop an application locally and then cross-compile with the same gcc/glibc version once (or fewer times at least).
> 
> In terms of getting this PR across the line. I'd like to see the package versions that you are adding back marked as obsolete with a comment in the version.desc about the EoS date for OL7. I've done all the removals I intend to for now and hopefully I'd see the comment and remember post the next release.

Ok, consider it done, thanks!
 
> Unless @stilor (or anyone else) is strongly opposed to the updated PR I'd be happy to merge it.

great!

gemaq:
@cpackham, IMHO i also detected a non-easy way to update branches/patches, because for instance if we have a branch configured in version.desc  (e.g repository_branch='oracle/gcc/ol7-u9')  and suppose this branch needs some patches to be applied, so they must be located in: ${devel_vcs}-${unique_id} (bundled_patch_dir), or (bundled_common_patch_dir) directories, but when is updated with a new commit the directory is not longer valid to looks for patches because the hash was changed, so would be nice to have a bundled_patch_dir_{version,branch_name} or something like this pointed to this branch in order to don't update directories in CT-NG. What do you think? .. thanks for your comments!

cpackham:
> @cpackham, IMHO i also detected a non-easy way to update branches/patches, because for instance if we have a branch configured in version.desc (e.g repository_branch='oracle/gcc/ol7-u9') and suppose this branch needs some patches to be applied, so they must be located in: develvcs−{unique_id} (bundled_patch_dir), or (bundled_common_patch_dir) directories, but when is updated with a new commit the directory is not longer valid to looks for patches because the hash was changed, so would be nice to have a bundled_patch_dir_{version,branch_name} or something like this pointed to this branch in order to don't update directories in CT-NG. What do you think? .. thanks for your comments!

Does `bundled_common_patch_dir` not work for this case? Or is that now too generic because it'll apply to all gcc-oracle versions.

Adding patches on top of VCS sources is not really an intended use-case (at least IMHO, that all pre-dates my involvement). Once a VCS is involved then surely the place to put patches is in the VCS. It does work and might be useful short-term while preparing patches to send upstream but I don't really see it as a thing that should be relied on long-term.

gemaq:
> > @cpackham, IMHO i also detected a non-easy way to update branches/patches, because for instance if we have a branch configured in version.desc (e.g repository_branch='oracle/gcc/ol7-u9') and suppose this branch needs some patches to be applied, so they must be located in: develvcs−{unique_id} (bundled_patch_dir), or (bundled_common_patch_dir) directories, but when is updated with a new commit the directory is not longer valid to looks for patches because the hash was changed, so would be nice to have a bundled_patch_dir_{version,branch_name} or something like this pointed to this branch in order to don't update directories in CT-NG. What do you think? .. thanks for your comments!
> 
> Does `bundled_common_patch_dir` not work for this case? Or is that now too generic because it'll apply to all gcc-oracle versions.

Too generic, because now we have two distros OL7 and OL8

> 
> Adding patches on top of VCS sources is not really an intended use-case (at least IMHO, that all pre-dates my involvement). Once a VCS is involved then surely the place to put patches is in the VCS. It does work and might be useful short-term while preparing patches to send upstream but I don't really see it as a thing that should be relied on long-term.

Oh, Ok, in that case I will use git-xxxx directory to put "temporary" patches

gemaq:
Done. Please let me know your comments, thanks in advance!

packages/binutils-oracle/2.27/version.desc
```diff
@@ -0,0 +1,13 @@
...
+#
+#    ARM64: 'oracle/binutils/ol7-u9-aarch64'
+#    ARM32: 'oracle/binutils/ol7-u9-arm'
+#
+# branch name is adjusted in:
+#   samples/{aarch64,arm}*ol7u9-*
+#
+# using CT_BINUTILS_ORACLE_DEVEL_BRANCH
+#
+# This package will be "OBSOLETE" at EoS date for OL7: Jul 2014.
```
cpackham, line 12: I was also meaning this and the other packages should have `obsolete='yes'` to require people to select `CT_OBSOLETE` to get these older versions. The comment is mainly for the maintainers so we don't unintentionally remove these again until after the EoS date.

cpackham (commented):
Looks pretty good. One minor comment below

packages/binutils-oracle/2.27/version.desc
gemaq: > 2024

ohh no.. i'm a fool .. sorry for this.

gemaq: > I was also meaning this and the other packages should have `obsolete='yes'` to require people to select `CT_OBSOLETE` to get these older versions. The comment is mainly for the maintainers so we don't unintentionally remove these again until after the EoS date.

Ooh, ok understood, there is a new push now with `obsolete='yes'`  in version.desc
Thanks @cpackham !

cpackham (approved):

PR #1762, opened 2022-06-14
bootstrap: parallelly generating packages

chent11:
This can improve the performance when running bootstrap.


cpackham:
I like the idea but unfortunately this mangles the generated `maintainer/package-versions` file

chent11:
@cpackham Thanks for the review.

It's weird that the `maintainer/package-versions` file didn't mess up after several tests. But I know it's unreliable, so I made some changes to make the file generation more reliable.

PR #1768, opened 2022-06-27
Fix mixed indent in bootstrap

bhundven:
Tabs or spaces... I believe the policy on ct-ng is spaces for shell
scripts. So update bootstrap to stick to 4 space indent.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# vi: ts=4:sw=4:et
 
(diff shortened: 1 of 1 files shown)


bootstrap
```diff
@@ -823,7 +824,7 @@ msg "*** Gathering the list of data files to install"
             continue
         fi
         echo " \\"
-        echo -n "	${f}"
+        echo -n "    ${f}"
```
cpackham, line 827: I think this one should still be a tab since we're outputting a Makefile

bhundven:
Also need to create a new PR for this one.

PR #1774, opened 2022-06-27
Update configure ac

bhundven:
Some updates to configure.ac since 2.70

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -300,7 +300,6 @@ CTNG_SET_KCONFIG_OPTION([sha512sum])
 
 # FIXME why checking if not using the result? We don't provide replacemant malloc/alloca/...
 AC_C_INLINE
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -93,7 +93,7 @@ AC_PROG_CC
 AM_PROG_CC_C_O
 AC_PROG_CXX
 AC_PROG_CPP
-AC_PROG_LEX(yywrap)
+AC_PROG_LEX([yywrap])
```
cpackham, line 96: This probably could have been squashed into the original commit. No harm done though and probably not worth another round of CI builds just to do the squash.

cpackham (approved):

configure.ac
bhundven: Resolved

PR #1775, opened 2022-06-27
Fix mixed indent in bootstrap

bhundven:
Tabs or spaces... I believe the policy on ct-ng is spaces for shell
scripts. So update bootstrap to stick to 4 space indent.

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# vi: ts=4:sw=4:et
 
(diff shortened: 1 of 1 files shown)


bhundven:
Fixed the echo at the end to do `echo -en "\t${f}"`, since the vim modeline at the top now forces tabs to be 4 spaces, and it makes it a bit more "literal".

cpackham (approved):

PR #1792, opened 2022-07-26
CI: Install meson and ninja when building

cpackham:
Add meson and ninja-build to the prerequisites so we actually build
picolibc.


bhundven:
There are two separate checks for ninja and meson.
Maybe make those decisions in configure.ac?

bhundven (approved):

bhundven:
> There are two separate checks for ninja and meson. Maybe make those decisions in configure.ac?

I digress on this comment. This is not related.
💯 Approve.

bhundven (approved):

PR #1797, opened 2022-08-02
Updates

cpackham:
Update glibc and linux


packages/glibc/2.36/version.desc
```diff
@@ -0,0 +1 @@
+# Released 3 Feb 2022
```
jiegec, line 1: This should be `1 Aug 2022`.

cpackham: Good catch. Will fix

PR #1818, opened 2022-09-11
avr,pru: Disable CT_CC_GCC_ENABLE_DEFAULT_PIE

dinuxbg:
PIE is not supported by PRU and AVR backends for GCC.

This fixes LD errors when trying to link user programs with a crosstool-NG toolchain:

  /home/dinux/x-tools/avr/lib/gcc/avr/12.2.0/../../../../avr/bin/ld: -pie not supported

Signed-off-by: Dimitar Dimitrov <dimitar@dinux.eu>

### DIFF
--- a/config/cc/gcc.in
+++ b/config/cc/gcc.in
@@ -201,7 +201,7 @@ config CC_GCC_ENABLE_DEFAULT_PIE
     bool
     prompt "Enable Position Independent Executable as default"
     default y
-    depends on GCC_6_or_later && ! WINDOWS && ! ARCH_MIPS
+    depends on GCC_6_or_later && ! WINDOWS && ! ARCH_MIPS && ! ARCH_AVR && ! ARCH_PRU
     help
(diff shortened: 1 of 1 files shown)


samples/avr/crosstool.config
```diff
@@ -1,4 +1,5 @@
 CT_CONFIG_VERSION="4"
 CT_ARCH_AVR=y
+# CT_CC_GCC_ENABLE_DEFAULT_PIE is not set
```
cpackham, line 3: I think it would be better to add `&& ! ARCH_AVR && ! ARCH_PRU` to the depends in  [config/cc/gcc.in](https://github.com/crosstool-ng/crosstool-ng/blob/master/config/cc/gcc.in#L200). That way any custom AVR configs won't need updating

dinuxbg: Thanks. Fixed and updated the pull request.

PR #1820, opened 2022-09-12
Build libopcodes for target

driesdeschout:
Add an option to build and install binutil's libopcodes into the toolchain.

Setting the default to 'y' like in the libbfd and libiberty options, not sure if that is ok.

### DIFF
--- a/config/binutils/binutils.in
+++ b/config/binutils/binutils.in
@@ -184,6 +184,11 @@ config BINUTILS_FOR_TARGET_BFD
     prompt "libbfd"
     default y
 
+config BINUTILS_FOR_TARGET_OPCODES
+    bool
+    prompt "libopcodes"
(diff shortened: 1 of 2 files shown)


cpackham:
Please include a commit message ([this](https://cbea.ms/git-commit/) is a good blog post on writing good commit messages) and add your signed-off-by line.

As for default y. I'd normally say no unless there is a compelling reason most crosstool-ng users might want this.

driesdeschout:
I needed libopcodes when building the Linux perf tool for target: when build with the libbfd feature, it assumes libopcodes to be available, without checking. In that sense, the default 'y' matching libbfd's default 'y' makes sense. But, that is just one point of view.

cpackham:
Yep that sounds like a good enough justification.

PR #1852, opened 2022-11-18
packages: Fix broken zlib mirror URL

jnohlgard:
The [fossils directory](https://zlib.net/fossils/) contains all versions of zlib, including the current version.


cpackham:
unfortunately the fossils directory only has the .tar.gz files not the full complement of other artifacts we could probably live without the .xz file but the signature is kind of important

jnohlgard:
I see, let's close this.

Issue #1850, opened 2022-11-12
Issues with include paths when building picolibc configuration for a Windows MinGW compiler


multiplemonomials:
Crosstool-NG version 1.25
Host OS: Ubuntu 20.04 (WSL2)

Hi all!  I was able to get a picolibc toolchain building fine for an Ubuntu host, but encountered some issues with generated include paths when trying to build the same toolchain as a canadian build for a Windows host.  Unfortunately, due to company policy, I can't share the exact files I used, but I can tell you what steps will reproduce the error.

1. Create a configuration with picolibc enabled: `ct-ng arm-picolibc-eabi`
2. Install the standard Ubuntu mingw toolchain with `sudo apt-get install mingw-w64`
3. Use `ct-ng menuconfig` to change the toolchain type to "canadian" and the host system tuple to "x86_64-w64-mingw32"
4. Edit .config to change the zlib version and work around #1832 
5. Edit /usr/local/share/crosstool-ng/scripts/build/companion_libs/050-zlib.sh and remove the following lines.  I'm not exactly sure why, but unless I did this, zlib did not link for mingw32:
    ```
    SHAREDLIB= \
    IMPLIB= \
    ```
6. Edit /usr/local/share/crosstool-ng/scripts/build/companion_libs/100-gmp.sh to manually apply the patch for #1328
7. Run `ct-ng build`.

By doing this, I get the following build error during the final step:
```
[INFO ]  =================================================================
[INFO ]  Installing libstdc++ picolibc
[EXTRA]    Configuring libstdc++ for picolibc
[EXTRA]    Building libstdc++ picolibc library
[ERROR]    /home/jamie/crosstool/.build/HOST-x86_64-w64-mingw32/arm-picolibc-eabi/src/gcc/libgcc/../gcc/tsystem.h:87:10: fatal error: stdi.h: No such file or directory
[ERROR]    /home/jamie/crosstool/.build/HOST-x86_64-w64-mingw32/arm-picolibc-eabi/src/gcc/libgcc/../gcc/tsystem.h:87:10: fatal error: stdi.h: No such file or directory
[ERROR]    make[3]: *** [Makefile:501: _cmpdi2.o] Error 1
[ERROR]    make[3]: *** Waiting for unfinished jobs....
[...]

PR #1862, opened 2022-12-01
scripts/gcc: Add extra include dir with -idirafter

jnohlgard:
Fixes Canadian cross builds failing with missing include file `stdio.h` when building libstdc++ for a companion libc with system libc set to `LIBC_NONE`.


jnohlgard:
Using a config based on samples/arm-picolibc-eabi but configured for Canadian cross compile fails without this patch.

<details><summary><tt>x86_64-pc-linux-gnu,arm-picolibc-eabi.config</tt> configuration to verify with</summary>

```sh
CT_CONFIG_VERSION="4"
CT_ARCH_ARM=y
CT_MULTILIB=y
CT_TARGET_CFLAGS="-ftls-model=local-exec"
CT_TARGET_VENDOR="picolibc"
CT_CANADIAN=y
CT_HOST="x86_64-pc-linux-gnu"
CT_LIBC_NONE=y
CT_CC_GCC_MULTILIB_LIST="rmprofile"
CT_CC_GCC_CONFIG_TLS=y
CT_CC_GCC_LTO_ZSTD=m
CT_CC_LANG_CXX=y
CT_COMP_LIBS_PICOLIBC=y
```

</details>

The errors I get when using the above config on current master without this patch look like this:

`/work/picolibc-toolchain/build-pr-test/.build/HOST-x86_64-pc-linux-gnu/arm-picolibc-eabi/src/gcc/libgcc/../gcc/tsystem.h:87:10: fatal error: stdio.h: No such file or directory`

Not sure exactly what is different between a canadian and regular cross build in the gcc configuration, but the logs show that the configure script on a regular cross finds the target compiler as xgcc with some extra -isystem flags while the Canadian cross logs show that the configure script only uses the target compiler without any automatic extra include flags, so instead we add them manually using CFLAGS_FOR_TARGET.

jnohlgard:
Added the Canadian cross test config to the samples directory.

jnohlgard:
Fixes #1850

cpackham (changes requested):
Could you put a bit in the commit message about why this change makes sense? I don't know if `-z` is any more or less portable than `= ""` (I did check that it is in posix) so I could take or leave this change. It does appear that `-z` is used more throughout crosstool-ng so that may be justification enough.

jnohlgard:
Updated the commit message in the first commit. The change for the if statements felt too small to warrant opening a separate PR so they got to piggyback on this one.

cpackham (approved):

PR #1863, opened 2022-12-01
picolibc: Place C++ headers in a subdir of the libc headers

jnohlgard:
This changes the picolibc libstdc++ configuration to place the C++ headers under the same base include path that picolibc uses for the C headers.

### DIFF
--- a/scripts/build/companion_libs/340-picolibc.sh
+++ b/scripts/build/companion_libs/340-picolibc.sh
@@ -41,6 +41,7 @@ do_cc_libstdcxx_picolibc()
         final_opts+=( "build_step=libstdcxx" )
         final_opts+=( "extra_config+=('--enable-stdio=stdio_pure')" )
         final_opts+=( "extra_config+=('--disable-wchar_t')" )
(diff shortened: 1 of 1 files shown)


cpackham (changes requested):
Could you put a bit of info in the commit message. The PR only exists on github so we want to get useful info in the commit message which will go with the repo wherever it is hosted in the future.

cpackham:
@keith-packard any opinion on this?

jnohlgard:
@cpackham that makes sense. I have pushed an updated commit with a bit more explanation in the commit message.

keith-packard:
> @keith-packard any opinion on this?

Seems like this will require changes to the picolibc.specs file so that the right C++ headers will be used? Probably better than assuming they don't depend on the underlying C library though?

cpackham:
@keith-packard between #1901 and #1908 is this still needed?

jnohlgard:
Let's close this for now. I will not have the time to rework this to actually do what I originally intended.

PR #1895, opened 2023-02-17
binutils: build static gold without -all-static

jokva:
Most of the binutils programs are built with gcc and linked with libtool, but gold (and dwp) are exceptions being C++ programs. autotools tries to link with $(CXXLD), which usually maps to g++, and not libtool like for the C programs.

binutils.sh passes LDFLAGS="$ldflags -all-static" to make, which assumes the linking program is always libtool, but this fails for gold. By checking if we're currently building binutils with gold we can pass the g++ flag --static to the C++ linker (through --with-gold-ldflags=... to configure), and build gold in a separate make invocation without the overriden LDFLAGS.

### DIFF
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -176,6 +176,12 @@ do_binutils_backend() {
 
     [ "${CT_TOOLCHAIN_ENABLE_NLS}" != "y" ] && extra_config+=("--disable-nls")
 
+    if [ "${CT_STATIC_TOOLCHAIN}" = "y" ] && [ "${CT_BINUTILS_LINKER_GOLD}" = "y" ]; then
+        # gold links with CXXLINK/g++, not libtool, and does not understand
+        # -all-static
+        extra_config+=("--with-gold-ldflags=--static")
+    fi
+
     # Disable usage of glob for higher compatibility.
     # Not strictly needed for anything but GDB anyways.
     export ac_cv_func_glob=no
@@ -209,6 +215,9 @@ do_binutils_backend() {
     fi
 
     CT_DoLog EXTRA "Building binutils"
+    if [ "${CT_BINUTILS_LINKER_GOLD}" = "y" ]; then
+        CT_DoExecLog ALL make -C gold ${CT_JOBSFLAGS}
+    fi
     CT_DoExecLog ALL make "${extra_make_flags[@]}" ${CT_JOBSFLAGS}
 
     CT_DoLog EXTRA "Installing binutils"


jokva:
I didn't find another obvious way to do this, and dwp and gold would not link statically since they link with g++ and not libtool. I didn't find a way to configure static gold already. What do you think?

cpackham:
Could you post a .config that shows the problem. I had a quick try and the resulting cross-ld was statically linked

cpackham:
Ah we stop you from linking gold statically. So how did you get it to fail?

scripts/build/binutils/binutils.sh
```diff
@@ -176,6 +176,12 @@ do_binutils_backend() {
 
     [ "${CT_TOOLCHAIN_ENABLE_NLS}" != "y" ] && extra_config+=("--disable-nls")
 
+    if [ "${CT_STATIC_TOOLCHAIN}" = "y" ] && [ "${CT_BINUTILS_LINKER_GOLD}" = "y" ]; then
+        # gold links with CXXLINK/g++, not libtool, and does not understand
+        # -all-static
+        extra_config+=("--with-gold-ldflags=--static")
+    fi
+
```
cpackham, line 184: I think if we just change the `-all-static` below to `-static` it'll work for both libtool and g++. But the bigger question is if we change things to allow gold to be statically linked and used for linking glibc. There's some old code that stops this combination but I'm not sure any of the justification is still relevant.

jokva:
> Ah we stop you from linking gold statically. So how did you get it to fail?
That's correct. The regular cross-ld is statically linked and perfectly fine.

I tried two ways of building gold; one with --enable-gold given to binutils, the other to specify ld, gold in CT_BINUTILS_LINKERS_LIST. Both fail at the same stage:

```
[INFO ]  Installing binutils for host
[EXTRA]    Configuring binutils
[EXTRA]    Prepare binutils for static build
[EXTRA]    Building binutils
[ERROR]    g++: error: unrecognized command-line option '-all-static'; did you mean '--static'?
[ERROR]    make[6]: *** [Makefile:1014: dwp] Error 1
[ERROR]    make[5]: *** [Makefile:1158: all-recursive] Error 1
[ERROR]    make[4]: *** [Makefile:898: all] Error 2
[ERROR]    make[3]: *** [Makefile:6842: all-gold] Error 2
[ERROR]    make[2]: *** [Makefile:1000: all] Error 2
```
[fails-on-configure-binutils---enable-gold.config.txt](https://github.com/crosstool-ng/crosstool-ng/files/10831289/fails-on-configure-binutils---enable-gold.config.txt)

[fails-on-binutils-linker-list.config.txt](https://github.com/crosstool-ng/crosstool-ng/files/10831290/fails-on-binutils-linker-list.config.txt)

Under some circumstances the build finishes (I can try to dig up a .config later, if interesting) and ld.gold is produced, but linked dynamically.

scripts/build/binutils/binutils.sh
jokva: That could very well be - I have mostly tried the combination glibc 2.27 and binutils 2.34 and cannot recall any problems.

jokva:
Also note that I modified the config by hand, as the kconfig for gold disables it if the toolchain is static.

scripts/build/binutils/binutils.sh
jokva: Oh, I might have misunderstood this. I also build bfd and keep it the default, ending up with two linkers. There are issues like this https://sourceware.org/bugzilla/show_bug.cgi?id=14995 but I don't know if it still applies.

cpackham:
I think I've got something that can sanely build a static ld.gold. I'll post it after the last volley of PRs gets merged.

jokva:
Splendid news!

cpackham:
#1927 is my attempt at dealing with gold.

cpackham:
Obsoleted by #1927

PR #1901, opened 2023-02-23
Allowing picolibc to be used as the 'system' C library

keith-packard:
The current picolibc mechanism adds picolibc as a 'companion' library so that users can build it in addition to another C library. Using picolibc with the resulting toolchain requires the inclusion of `--specs=picolibc.specs` in all of the compiler command lines. 

This series creates an option for picolibc to be the default C library while still leaving the option for using it as a companion library available. The picolibc configuration options are moved to the C-library section and made available when configuring the companion library. Other GCC options required for picolibc are integrated into the gcc scripts. This work builds upon patches started by @jnohlgard , whose name continues to appear on the main patches. I modified those in minor ways, moving the picolibc configuration and adjusting some dependency mechanisms.

To make this work, changes to gcc (which have been submitted upstream) are included that add picolibc support.

I've tested the resulting toolchains on arm and risc-v, verifying that both C and C++ programs build and run correctly.


jnohlgard:
Thanks for continuing my work on this, @keith-packard!

keith-packard:
Thanks for the most excellent patch to work from! I'd been working on the gcc bits upstream and you made me realize that ct-ng could actually take those as patches while they get reviewed.

cpackham:
Build failures were due to the macos runners running out of disk space. We've probably been pushing them a bit too hard so lets leave the dust to settle before kicking this off again.

keith-packard:
Yeah, I noticed and (foolishly) attempted to make them try again. Will let you drive.

cpackham:
@keith-packard are you able to access a macos system? Or maybe just one with CC=clang. I had assumed that the failures were just a run of bad luck but now I wonder if there is something going haywire for picolibc under lto

keith-packard:
> @keith-packard are you able to access a macos system? Or maybe just one with CC=clang. I had assumed that the failures were just a run of bad luck but now I wonder if there is something going haywire for picolibc under lto

Oof. "how could that have any effect". I do have a modest macos system here that I can experiment with, but probably won't get to that today.

keith-packard:
building; had to figure out how to create a case-sensitive spot for ~/x-tools, and then how to install half of a "real" operating system to get the build going.

cpackham:
I was thinking about this a bit more. Initially I couldn't fathom picolibc taking up more space than a full-blown glibc or musl. But there may be a difference in that picolibc supports a lot more of the multilib permutations. So it could simply be that picolibc is genuinely hitting a limit on the free tier macos runners. If that's the case we might just add picolibc to the list of exclusions for macos.

keith-packard:
We could reduce the amount of multilib variance too. Let me see how a macos build goes here; if that 'just works', then at least we'll know it's a simple space problem. Reducing the multilib matrix to a couple of useful variants, or even building it for a single variant might at least keep the bits from failing in the future?

keith-packard:
ok, the macos build of arm-picolibc-eabi appears to be happy. It consumed a bit over 18GB of disk space --  19114064k is used on the case-sensitive partition I created just for this. That includes the crosstool-ng directory, build directory and x-tools install directory. Peak usage was a bit higher, probably around  19362540k. Feel free to adjust the CI as you prefer.

PR #1904, opened 2023-02-25
Updates - binutils and linux

cpackham:


cpackham:
Builds fail due to https://github.com/uclinux-dev/elf2flt/issues/26

PR #1908, opened 2023-02-28
packages/gcc: Add gcc patch to allow getenv default values

keith-packard:
This patch, submitted upstream but not (yet?) accepted, adds a third parameter to the specs file 'getenv' function that provides a value for when the environment variable is not set, instead of having gcc fail.

This seemed like the safest way to provide a mechanism for getting the installed location of the toolchain from inside a specs file as, when not installed in the built-in location, gcc already sets the GCC_EXEC_PREFIX environment variable to a well defined location within that directory hierarchy, but when installed in the location specified at compile time, gcc does not. Providing a default value that matches the compile-time location then allows the specs file to compute paths relative to the current GCC installation location, whereever it is installed.

### DIFF
--- a/packages/gcc/12.2.0/0005-driver-Extend-getenv-function-to-allow-default-value.patch
+++ b/packages/gcc/12.2.0/0005-driver-Extend-getenv-function-to-allow-default-value.patch
@@ -1,0 +1,102 @@
+From fff469f89d865ef3c15efe8e6b0511ea4d48603d Mon Sep 17 00:00:00 2001
+From: Keith Packard <keithp@keithp.com>
+Date: Fri, 26 Aug 2022 14:30:03 -0700
+Subject: [PATCH 6/6] driver: Extend 'getenv' function to allow default value
+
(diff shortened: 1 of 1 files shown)


bhundven:
Yea, that's crazy... back in August. Maybe bump the thread? I know they hate it when you do that. I'm going to approve this, but I'm going to open an issue to make some new "experimental" patch path that is behind `CT_EXPERIMENTAL`, so that patches that are experimental can be enabled with a sub option of experimental.

bhundven (approved):

abrodkin:
For the record, here's the patch of @keith-packard in the GCC mailing list https://gcc.gnu.org/pipermail/gcc-patches/2022-August/600452.html. @keith-packard is there a bug in GCC's Bugzilla associated with this patch? Usually it helps to get changes integrated.

keith-packard:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109627

PR #1910, opened 2023-03-06
Support targeting macOS aarch64 in older GCCs

sfackler:
Ported from https://github.com/richfelker/musl-cross-make/pull/129


cpackham:
I'd like to see a bit of a blurb in the commit message as well as a Signed-off-by: line (I really must see if there is a github action to check that so I don't need to keep asking). I'm a bit ambivalent to the change itself, I'd personally try to steer people towards newer compilers rather than getting older compilers to support new platforms but it's not really going to affect anything to carry these patches.

sfackler:
Updated!

I'd love to use modern GCCs but unfortunately have to build binaries that can dynamically link to the target environment's libstdc++ :(

cpackham (approved):

PR #1917, opened 2023-03-10
Experimental patches

bhundven:


cpackham (approved):

cpackham:
I did wonder if patches like this should live under contrib people could add them to their local directory if they choose. But then they'd probably get no exposure.

PR #1931, opened 2023-03-25
config/arch/x86: enable libsanitizer

jdknight:
GCC's libsanitizer module does provide a subset of support for x86 targets (e.g. libasan). GCC's `libsanitizer/configure.tgt` file only outlines flags for optional sanitizers.

### DIFF
--- a/config/arch/x86.in
+++ b/config/arch/x86.in
@@ -9,7 +9,7 @@
 ## select ARCH_SUPPORTS_WITH_CPU
 ## select ARCH_SUPPORTS_WITH_TUNE
 ## select ARCH_SUPPORTS_WITH_32_64
-## select ARCH_SUPPORTS_LIBSANITIZER if ARCH_64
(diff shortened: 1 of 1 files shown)


bhundven (commented):
I'm not against this change. Just a comment.
Some configurations have issues with this enabled: https://github.com/crosstool-ng/crosstool-ng/commit/f5e0d33ab46086c88ce800bb80be547c046a0efa

bhundven:
I want to let the CI run to see if there are any problems. Otherwise, I'm ok with this change.

cpackham:
Yeah I'm not sure where I got the impression it wasn't supported for x86_64. It might be because the `x86_64-*-linux` one is at the top of the case whereas the ` x86_64-*-freebsd`/` x86_64-*-darwin` is at the bottom. Regardless the CI run as passed so I think we're all good. Will merge shortly

PR #1933, opened 2023-04-03
Picolibc 1.8.1

keith-packard:
Update picolibc to version 1.8.1, adding new configuration for default linker script values and a sample riscv32-picolibc-elf setup.


cpackham (approved):
Is it worth adding riscv32-picolibc-elf to the CI builds?

cpackham:
Just a passing thought, How many older versions of picolibc should we keep supporting? I'm starting to think about the next ct-ng release (hopefully shortly after the GCC 13 release) so now might be a good time to think about marking some versions as obsolete so we can remove them post release.

keith-packard:
> Just a passing thought, How many older versions of picolibc should we keep supporting? I'm starting to think about the next ct-ng release (hopefully shortly after the GCC 13 release) so now might be a good time to think about marking some versions as obsolete so we can remove them post release.

I'm not sure there's any real reason to support more than a couple of versions; it's a C library, after all, and kinda by definition doesn't have any incompatible changes across releases. For people interested in reproducible builds, they're going to need to keep the whole toolchain frozen, presumably by using an older version of ct-ng.

keith-packard:
> Is it worth adding riscv32-picolibc-elf to the CI builds?

That's a good question. If you run the picolibc test suite under qemu, you do get fairly complete end-to-end testing of the system, which can catch compiler regressions that seem more likely to occur on a target changing as fast as risc-v. I'm not sure ct-ng CI does that?

PR #1936, opened 2023-04-19
zstd: add Zstandard v1.5.5

MOHAMED19OS:


cpackham:
Code change is OK but I'd like to see a commit message with a sign-off.

Weird thing is the email notification I got did list a bit of a changelog which may have been part of the PR or the commit for this but there's no sign of it through the github web interface so I'm not sure where it came from or who added it.

cpackham:
c116c9a2cd is a recent example of how I'd like commits to look.

PR #1952, opened 2023-05-10
newlib: Add missing linker options for nano.specs

kolerov:
Old options `%(newlib_nano_link)` for the linker must be passed otherwise linking may fail. E.g., in case of multilib configurations a correct emulation mode may be not passed.


cpackham (approved):
Looks OK. CI failure seems to be an an issue on githubs side.

PR #1953, opened 2023-05-11
Add ncurses 6.4

bhundven:


cpackham (changes requested):
Something's gone weird here. The checksums are for ncurses 6.2

cpackham:
Consider using `./maintainer/manage-packages.sh` for package bumps. It takes care of a lot of the fiddly bits. Only trick is you have to add the new `package/foo/VERSION` before running `./bootstrap` for it to pick up that there is a new version.

cpackham (approved):

PR #1956, opened 2023-05-13
strace: Add strace 6.3

cpackham:
https://lists.strace.io/pipermail/strace-devel/2023-May/011317.html

Add strace 6.3. Remove 6.2.


cpackham:
Seems to be hitting https://github.com/strace/strace/issues/253

PR #1964, opened 2023-05-20
Picolibc prune

cpackham:
Marks some verisons of picolibc as obsolete an drops 1.7.9


cpackham:
@keith-packard does this look OK to you?

keith-packard:
> @keith-packard does this look OK to you?

yup, seems fine. Most picolibc users are keeping pretty close to the latest release.

PR #1968, opened 2023-05-27
gdb: Static cross GDB is linked statically with std c++ library

maxmorozov:
This change makes linking cross GDBstatically with std c++ library if statis build is selected (because there is no separate option for std c++ lib for cross GDB). Also fixes usage of not existing variable CT_GDB_NATIVE_STATIC_LIBSTDC, by replacing it with CT_GDB_NATIVE_STATIC_LIBSTDCXX


cpackham (changes requested):
Change looks fine but could I please see a proper commit message with a sign-off.

PR #1989, opened 2023-07-12
Revert "Update requirements: autoconf==2.71 and automake>=1.16"

tpetazzoni:
This reverts commit 658f5e574f4025655da318fb0b688a394b6ff43f, which was made without any justification, and prevents using crosstool-ng even in decently recent distributions like Ubuntu 20.04 (which has autoconf 2.69).

We have tested that ./bootstrap && ./configure --enable-local works fine on Ubuntu 18.04, which uses autoconf 2.69 and automake 1.15, so the existing requirements were enough.

### DIFF
--- a/config/configure.in.in
+++ b/config/configure.in.in
@@ -36,14 +36,14 @@ config CONFIGURE_has_libtool_2_4_or_newer
 config CONFIGURE_has_libtoolize_2_4_or_newer
     @KCONFIG_libtoolize_2_4_or_newer@
 
-config CONFIGURE_has_autoconf_2_71_or_newer
-    @KCONFIG_autoconf_2_71_or_newer@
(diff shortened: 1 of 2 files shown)


cpackham:
I believe this was done to improve python detection on some newer systems (arch and macos I believe). Still breaking support for ubuntu 20.04 is definitely a good justification for a revert. @bhundven any comment?

tpetazzoni:
Improve python detection? In what sense? Newer autoconf has improved macros for python detection? Probably this should have been explained in the commit log then?

nolange:
Python detection is a weird thing in itself. I made a debian package and it will use the path /usr/bin/python3.9 instead of just /usr/bin/python3, meaning it will break after every python3 update.

I would understand if there is some setup needed for headers/libs, but its just the plain exeutable that you can and should find via `PATH` ?

Issue #1884, opened 2023-02-01
Build for RPi Zero, glibc 2.31 freezes in multilib step


Lithimlin:
I'm currently trying to build a toolchain so I can compile for a Pi Zero.
The Zero has the following:
- OS and kernel: `Raspbian 32bit, Linux 5.15.76+ armv6l`
- glibc version: `ldd (Debian GLIBC 2.31-13+rpt2+rpi1+deb11u5) 2.31`

I have specified the version of glibc in the config and ever since when I try to build the chain, the process freezes in the multilib step without any indications as to why this is happening.
Before pulling down the glibc version I could build it just fine, but the compiled code didn't run on the Pi Zero due to the version mismatch.
I'm not sure if I'm missing further settings and would appreciate some help.

```
[...]
[INFO ]  =================================================================
[INFO ]  Installing kernel headers
[EXTRA]    Installing kernel headers
[INFO ]  Installing kernel headers: done in 9.66s (at 05:21)
[INFO ]  =================================================================
[INFO ]  Installing core C gcc compiler
[EXTRA]    Configuring core C gcc compiler
[EXTRA]    Building gcc
[EXTRA]    Installing gcc
[EXTRA]    Housekeeping for core gcc compiler
[EXTRA]       '' --> lib (gcc)   lib (os)
[INFO ]  Installing core C gcc compiler: done in 474.95s (at 13:16)
[INFO ]  =================================================================
[INFO ]  Installing C library
[INFO ]    =================================================================
[INFO ]    Building for multilib 1/1: ''
[EXTRA]      Configuring C library
[EXTRA]      Building C library
[15:20] -
```

- Crosstool-NG version: `1.25.0.124_0ce5b27`
- Host OS and kernel: `Arch Linux x86_64, 6.1.8-arch1-1`
- Attach `.config`: [config.txt](https://github.com/crosstool-ng/crosstool-ng/files/10559747/config.txt)
- Attach `build.log`: [build.zip](https://github.com/crosstool-ng/crosstool-ng/files/10559754/build.zip)

cpackham:
When you changed the glibc version did you start from fresh? There could be some left over build artifact from the earlier build that's interfering.

You can select `CT_LOG_ALL` to get (a lot) more output during the build, that'd help distinguish a genuine freeze from things just taking a while.

[...]

Issue #1975, opened 2023-06-06
"ct-ng build.1" hangs, cpu 100%


antekone:
I'm trying to run the `build.1` command, but crosstool-NG hangs during build of libc, and stays on 100% CPU utilization. The same happens with `build`.

ps auxf:

```
buildbot   86188  0.0  0.0  16424  4892 ?        Ss   Jun05   0:17 tmux
buildbot   86189  0.0  0.0  11228  5504 pts/1    Ss   Jun05   0:00  \_ -bash
buildbot  333645  0.0  0.0  14132  3968 pts/1    S+   10:26   0:00  |   \_ /usr/bin/make -rf /home/buildbot/dev/thirdparty/.bin/bin/ct-ng build.1
buildbot  333653  0.0  0.0  14128  4096 pts/1    S+   10:26   0:00  |       \_ /usr/bin/make -rf /home/buildbot/dev/thirdparty/.bin/bin/ct-ng build CT_JOBS=1
buildbot  333661  0.0  0.0  13352  7296 pts/1    S+   10:26   0:01  |           \_ /usr/bin/bash /home/buildbot/dev/thirdparty/.bin/share/crosstool-ng/scripts/crosstool-NG.sh
buildbot  491732  0.0  0.0  13352  5604 pts/1    S+   11:25   0:00  |               \_ /usr/bin/bash /home/buildbot/dev/thirdparty/.bin/share/crosstool-ng/scripts/crosstool-NG.sh
buildbot  493826  0.0  0.0  13352  5480 pts/1    S+   11:25   0:00  |                   \_ /usr/bin/bash /home/buildbot/dev/thirdparty/.bin/share/crosstool-ng/scripts/crosstool-NG.sh
buildbot  493830  0.0  0.0   8272  3200 pts/1    S+   11:25   0:00  |                       \_ make -j1 -l CXX= default_cflags= BUILD_CFLAGS=-O2 -g -I/home/buildbot/dev/thirdparty/crosstool-ng/.build/x86_64-centos7-linux-gnu/buildtools/include   BUILD_CPPFLAGS= BUILD_LDFLAGS=-L/home/buildbot/dev/thirdparty/crosstool-ng/.build/x86_64-centos7-linux-gnu/buildtools/lib   all
[...]

Issue #1932, opened 2023-03-29
Build Not completed in Arch Linux


MOHAMED19OS:
-  crosstool-NG version: the lest in GitHub
-  host OS and its version: 
        OS: Arch Linux x86_64
         Kernel: 6.2.8
         DE: Plasma 5.27.3
- any test for build stop

`[INFO ]  Performing some trivial sanity checks
[WARN ]  Number of open files 1024 may not be sufficient to build the toolchain; increasing to 2048
[INFO ]  Build started 20230329.050451
[INFO ]  Building environment variables
[WARN ]  Directory '/home/mohamed/src' does not exist.
[WARN ]  Will not save downloaded tarballs to local storage.
[EXTRA]  Preparing working directories
[EXTRA]  Installing user-supplied crosstool-NG configuration
[EXTRA]  =================================================================
[EXTRA]  Dumping internal crosstool-NG configuration
[EXTRA]    Building a toolchain for:
[EXTRA]      build  = x86_64-pc-linux-gnu
[EXTRA]      host   = x86_64-pc-linux-gnu
[EXTRA]      target = armv7-vuduo4k-linux-gnueabihf
[EXTRA]  Dumping internal crosstool-NG configuration: done in 0.07s (at 00:02)
[INFO ]  =================================================================
[INFO ]  Retrieving needed toolchain components' tarballs
[EXTRA]    Retrieving 'linux-6.1.13'
[EXTRA]    Verifying SHA512 checksum for 'linux-6.1.13.tar.xz'
[EXTRA]    Retrieving 'zlib-1.2.13'
[EXTRA]    Verifying SHA512 checksum for 'zlib-1.2.13.tar.xz'
[EXTRA]    Retrieving 'zstd-1.5.2'
[EXTRA]    Verifying SHA512 checksum for 'zstd-1.5.2.tar.gz'
[EXTRA]    Retrieving 'gmp-6.2.1'
[EXTRA]    Verifying SHA512 checksum for 'gmp-6.2.1.tar.xz'
[EXTRA]    Retrieving 'mpfr-4.1.0'
[EXTRA]    Verifying SHA512 checksum for 'mpfr-4.1.0.tar.xz'
[EXTRA]    Retrieving 'isl-0.25'
[EXTRA]    Verifying SHA512 checksum for 'isl-0.25.tar.xz'
[EXTRA]    Retrieving 'mpc-1.2.1'
[...]

PR #1990, opened 2023-07-15
provide a workaround for a glibc and GNU make incompatibility

### LINKED ISSUE #1946: Deal with old glibc + new GNU make incompatiblities
Users have been reporting the build hanging. Finally between #1944 and #1932 it's been identified that older glibc (<=2.31) has a problem with newer GNU Make (>=4.4). The workaround of having ct-ng build and use GNU Make 4.3 in the case seems to work. 

I expect that the problem will become more frequent as GNU Make 4.4 is rolled out in more distros and users update their systems. It would be good to codify the workaround such that if GNU Make 4.4 or newer is detected at configure time ct-ng automagically selects COMP_TOOLS_MAKE and GNU Make 4.3 when attempting to build one of these older glibcs.

cpackham:
Versions of GNU make newer than 4.4 trigger a hang in versions of glibc older than 2.31. As distros update to the newer GNU make version we'll start seeing more and more reports of this hang. Fortunately we already carry GNU make as a comp tool so all we need to do is select the right version when needed.

Fixes #1946, #1932, #1975


cpackham:
Fixes  #1884

PR #1992, opened 2023-07-18
newlib: add 4.2.0.20211231

nolange:
Remove upstreamed patches.


cpackham:
Given that we have newlib 4.3.0 is this really needed? I guess the same question applies to the other versions we have. We probably should at least mark <= 4.1.0 as obsolete

cpackham (approved):

nolange:
> Given that we have newlib 4.3.0 is this really needed? I guess the same question applies to the other versions we have. We probably should at least mark <= 4.1.0 as obsolete

Its been on my backlog for a while. its not recent but I sometimes go to ct-ng to build a fitting replacement for an older toolchain (or older libc on another architecture for test). Its a big feature for me that you can match compiler/software versions.
4.2.0 seems to be a more boring and  "safer" upgrade over 4.1.0 compared to 4.3.0 aswell

PR #1997, opened 2023-07-28
Fix missing quote mark

KirillSmirnov:
The missing quotes affect bootstrap routine:
./bootstrap: eval: line 646: unexpected EOF while looking for matching `'' ./bootstrap: eval: line 647: syntax error: unexpected end of file

For some reason bootstrap script ignores these errors and terminates successfully.


cpackham:
Thanks for the quick fix. Sorry for not noticing myself.

Issue #2001, opened 2023-08-03
GDB fails to build on Canadian Cross  


c-grant:
Attempting to build m68k-unknown-elf which I stated with from the sample.
I added GDB support.
Sample builds fine without gdb (Canadian).

crosstool-NG version: 1.25.0.209_465207b (pulled from git)
OS & version: Ubuntu 22.04.1 LTS

Files:
[build.log](https://github.com/crosstool-ng/crosstool-ng/files/12246119/build.log)
[m68k.config.txt](https://github.com/crosstool-ng/crosstool-ng/files/12246122/m68k.config.txt)

Compilers

[EXTRA] Dumping internal crosstool-NG configuration
[EXTRA] Building a toolchain for:
[EXTRA] build = x86_64-pc-linux-gnu
[EXTRA] host = x86_64-w64-mingw32
[EXTRA] target = m68k-unknown-elf

Seems to be a linker issue missing a library.

Tail of log
[ALL  ]    /usr/bin/x86_64-w64-mingw32-ld: ../readline/readline/libreadline.a(terminal.o):/home/vboxuser/CT-NG/m68k/.build/HOST-x86_64-w64-mingw32/m68k-unknown-elf/src/gdb/readline/readline/terminal.c:791: more undefined references to `__imp_tputs' follow

[ERROR]    collect2: error: ld returned 1 exit status

[ERROR]    make[3]: *** [Makefile:2149: gdb.exe] Error 1

[ALL  ]    make[3]: Leaving directory '/home/vboxuser/CT-NG/m68k/.build/HOST-x86_64-w64-mingw32/m68k-unknown-elf/build/build-gdb-cross/gdb'

KirillSmirnov:
Duplicate of #1982 

ncurses version <= 6.2 should be fine (assuming already applied PR #2000). ncurses >= 6.3 needs extra care if being built as static on mingw64.

cpackham:
looks like this might be fixed. Can anyone confirm?

KirillSmirnov:
Works for me on the latest git master.

cpackham:
Closing based on the above comment

Issue #1991, opened 2023-07-15
1.26 Release planning


cpackham:
Blocking issues: ~#1989~, ~#1990~
Nice to have: #1569, ~#2002~

- [x] Tag 1.26.0-rc1
- - [x] Upload 1.26.0-rc1 tarball to github
- - [x] Upload 1.26.0-rc1 tarball to crosstool-ng.org?
- [x] Run `build-all`
- [x] Tag 1.26.0-rc2 (if needed)
- - [x] Upload 1.26.0-rc2 tarball to github
- - [x] Upload 1.26.0-rc2 tarball to crosstool-ng.org?
- [x] Tag 1.26.0
- - [x] Upload 1.26.0 tarball to github
- - [x] Upload 1.26.0 tarball to crosstool-ng.org
- [x] Release announcement
- - [x] crossgcc ML 
- - [x] crosstool-ng website

nolange:
This one seems rather serious to me: #1973

Given that newlib 4.3.0 fails to build with old GCC versions when those patches are missing

modbw:
I have build all samples on Ubuntu 20.04. 9 of 118 samples failed.  These are:

```
arm-multilib-linux-uclibcgnueabi
i686-w64-mingw32_nios2-spico-elf
moxie-unknown-moxiebox
powerpc-unknown-linux-uclibc_m68k-unknown-uclinux-uclibc
x86_64-multilib-linux-uclibc_moxie-unknown-moxiebox
x86_64-multilib-linux-uclibc_powerpc-unknown-elf
x86_64-unknown-linux-gnu_arm-picolibc-eabi
x86_64-w64-mingw32_arm-cortexa9_neon-linux-gnueabihf
xtensa-fsf-elf
```
It took about 9 hours on my build server to run the test using [this](https://gist.github.com/modbw/e6454472642dc3c949967bb00446eb34) script which runs multiple builds in parallel.

[arm-multilib-linux-uclibcgnueabi.build.log.gz](https://github.com/crosstool-ng/crosstool-ng/files/12247429/arm-multilib-linux-uclibcgnueabi.build.log.gz)
[i686-w64-mingw32_nios2-spico-elf.build.log.gz](https://github.com/crosstool-ng/crosstool-ng/files/12247430/i686-w64-mingw32_nios2-spico-elf.build.log.gz)
[...]

Issue #1982, opened 2023-06-22
Sample configurations `i686-w64-mingw32,nios2-spico-elf` and `x86_64-w64-mingw32,arm-cortexa9_neon-linux-gnueabihf` broken


jbglaw:
Hi!

Based on the 36ad0b17a732aaffe4701d5d8d410d6e3e3abba9 commit, I see build issues for the i686-w64-mingw32,nios2-spico-elf 
 and `x86_64-w64-mingw32,arm-cortexa9_neon-linux-gnueabihf` sample configurations while linking GDB:

```
[...]

PR #2004, opened 2023-08-13
gdb: fix build with recent ncurses

KirillSmirnov:
For mingw host a special macro NCURSES_STATIC must be defined.


KirillSmirnov:
Fixes #2001, #1982 and a couple of samples in #1991

cpackham (approved):

PR #2007, opened 2023-08-15
allow plugins and LTO with static toolchains

chrisnc:
I was able to build a static toolchain that supports LTO by disabling these restrictions. Based on the original commits, it seems that this was a temporary limitation of the toolchain build and separately of Arch. I tested this with an arm-none-eabi target on Debian Bookworm.

https://github.com/crosstool-ng/crosstool-ng/commit/0841e2f820de7a4dca1
https://github.com/crosstool-ng/crosstool-ng/commit/45512b003d04b5a89c5

### DIFF
--- a/config/binutils/binutils.in
+++ b/config/binutils/binutils.in
@@ -128,7 +128,6 @@ config BINUTILS_LINKER_DEFAULT
 config BINUTILS_PLUGINS
     bool
     prompt "Enable support for plugins"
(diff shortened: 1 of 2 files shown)


cpackham:
Some of this discussion belongs in the commit message. Also going to need a proper Signed-off-by line.

chrisnc:
Updated commit message with more detail and the Signed-off-by.

PR #2008, opened 2023-08-16
fix static gdb

chrisnc:
I noticed that even when the static toolchain and static gdb options are enabled, `gdb` is the only binary in the output that is dynamically linked, which is just due to libtool vs. linking directly. The approach here is borrowed from `scripts/build/binutils/binutils.sh`.

### DIFF
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -255,6 +255,7 @@ do_gdb_backend()
 {
     local host prefix destdir cflags ldflags static buildtype subdir includedir
     local -a extra_config
+    local -a extra_make_flags
 
     for arg in "$@"; do
         case "$arg" in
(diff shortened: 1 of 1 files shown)


scripts/build/debug/300-gdb.sh
```diff
@@ -285,7 +286,6 @@ do_gdb_backend()
     fi
 
     if [ "${static}" = "y" ]; then
-        cflags+=" -static"
         ldflags+=" -static"
```
cpackham, line 289: Does changing this to `-all-static` work?

cpackham:
Commit message needs a proper Signed-off-by line.

scripts/build/debug/300-gdb.sh
chrisnc: I tried that at first, but this causes the `gdb` `./configure` step to fail, because it expects the toolchain to be able to build an executable without `libtool`, and in that case `-all-static` is not recognized.

chrisnc:
Thanks for the review! Rebased and added the Signed-off-by as requested.

PR #2013, opened 2023-08-20
Do not process 'packages' folder twice

KirillSmirnov:
The folder 'packages' is processed in bootstrap, so there is no need to process it again in Makefile.
This fixes a regression introduced in eb62ec3fbe3982f5f16561675fd0820d4313a0b4

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,7 +8,7 @@ SUBDIRS		= kconfig
 bin_SCRIPTS		= ct-ng
 CLEANFILES		= ct-ng bash-completion/ct-ng docs/ct-ng.1
 EXTRA_DIST		= bootstrap ct-ng.in bash-completion/ct-ng.in \
-				  docs/ct-ng.1.in packages maintainer testing
+				  docs/ct-ng.1.in maintainer testing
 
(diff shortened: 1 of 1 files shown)


cpackham:
Hmm. I don't think this'll work. We need packages in EXTRA_DIST so that it
gets included in the tarball.

KirillSmirnov:
The packages folder indeed gets included in the tarball, and nothing is missed there, I swear :-). The only difference is that newlib-nano patches are shipped as plain files/directories, not as symlinks.
BTW, the config folder and some others are processed the similar way: in bootstrap only.

The commit d14adeb1d96f784d2f2784ef4a7d03e2d3fe7e1e added packages in EXTRA_DIST. But I cannot figure out the motivation, and the commit message is unrelated. Maybe @stilor give us a hint?

cpackham:
> The packages folder indeed gets included in the tarball, and nothing is missed there, I swear :-). The only difference is that newlib-nano patches are shipped as plain files/directories, not as symlinks. BTW, the config folder and some others are processed the similar way: in bootstrap only.
> 

Ah, yes I see. OK lets get this in then

PR #2015, opened 2023-08-21
glibc: Add option for building libcrypt

cpackham:
As of glibc-2.38 libcrypt is not built by default. Add an option to allow building libcrypt support into glibc.


cpackham:
@milkylainen is this what you had in mind

PR #2016, opened 2023-08-23
gcc/13.2.0: Add patch with fixes for m68k soft float code

keith-packard:
Here's a patch I've sent upstream which fixes a range of floating point bugs in libgcc.

### DIFF
--- a/packages/gcc/13.2.0/0012-libgcc-m68k-Fixes-for-soft-float.patch
+++ b/packages/gcc/13.2.0/0012-libgcc-m68k-Fixes-for-soft-float.patch
@@ -1,0 +1,348 @@
+From 3393fcce1e6b18e580b712c0855bba13fb622531 Mon Sep 17 00:00:00 2001
+From: Keith Packard <keithp@keithp.com>
+Date: Tue, 22 Aug 2023 12:28:53 -0700
+Subject: [PATCH] libgcc/m68k: Fixes for soft float
+
+Check for non-zero denorm in __adddf3. Need to check both the upper and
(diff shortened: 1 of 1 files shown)


cpackham (approved):
Thanks @keith-packard looks great.

keith-packard:
> Thanks @keith-packard looks great.

I'm just finishing up testing these in picolibc CI -- I'm hoping the current run will succeed (picolibc CI runs are taking hours at this point, we have plans to improve that).

There are still pretty significant gaps in the 80-bit float code, but I stopped when picolibc CI stopped breaking ...

keith-packard:
oh, is there some reason the m68k-unknown-elf sample config doesn't enable multilib support? Seems kinda necessary for any embedded uses I can imagine.

cpackham:
> oh, is there some reason the m68k-unknown-elf sample config doesn't enable multilib support? Seems kinda necessary for any embedded uses I can imagine.

Not that I'm aware of. That sample is fairly old and multilib has always been considered a bit fragile. I don't have any objections to enabling it (after the 1.26.0 release).

cpackham:
Just noticed our CI builds don't cover m68k. I'll do one manually as well

keith-packard:
> Just noticed our CI builds don't cover m68k. I'll do one manually as well

If you want to run the picolibc test suite with the resulting toolchain (which I'm doing happily), you'll need a patched version of qemu as well. Speaking of which, I wonder if crosstool-ng should offer to include qemu as a tool in the toolchain? I know I need a matching qemu for every compiler ...

https://github.com/keith-packard/qemu/tree/m68k-mul2

PR #2023, opened 2023-08-31
glibc: Fix SVE detection on aarch64

cpackham:
glibc-2.38 requires SVE in order to support the vector math library for AArch64.


modbw:
I am not sure this is the right way to do this. RPI4 doesn't come with SVE implemented in the CPU.

cpackham:
Yeah the alternative is to disable mathvec which apparently will cause an
abi break. Or pin these configuration to glibc-2.37

PR #2024, opened 2023-09-01
picolibc: Bump to 1.8.3

lrusak:
This bumps the picolibc version to the latest release

ref: https://github.com/picolibc/picolibc/releases/tag/1.8.3


cpackham (commented):
Normally we'd mark 1.8.1 as obsolete and only remove it after a release. Not sure that's really a policy we should follow super strictly, picolibc is probably a bit niche and no-one will really mind that 1.8.1 is removed without warning.

PR #2027, opened 2023-09-06
uClibc-ng: issues with __pthread_key_create

cpackham:
A toolchain uclibc-ng-1.0.43, binutils-2.40 and gcc-13.2.0 hits the following error when building:

  ld.bfd: isl_test2.o: non-canonical reference to canonical protected function `__pthread_key_create' in x86_64-multilib-linux-uclibc/sysroot/lib64/libc.so.1
  ld.bfd: failed to set dynamic section sizes: bad value

Removing the protected attribute from __pthread_key_create seems to resolve the issue.


cpackham:
I'm really not sure if this is the right fix (or even a sensible one) but it does let the `x86_64-multilib-linux-uclibc,powerpc-unknown-elf` build complete.

QBos07:
doesn't change the binary output but is definitly something that shoudn't be like this

PR #2037, opened 2023-09-19
Add more caching to the CI

QBos07:
more specificaly to the tarballs download. The function CT_Fetch now
touches the already existing files to be comparable to the not used ones
that can araise when a package is updated.
This comparsion is needed because if it would not exist the tarball
would grow in size due to not used but still cached packages.
This would take time but is definitly something to worry about.

Signed-off-by: Quentin Boswank <qubos@outlook.de>

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -61,6 +61,19 @@ jobs:
       - name: "extract ct-ng"
         run: |
           tar -xf ct-ng.tar
+      - name: cache tarballs
+        id: cache
+        uses: actions/cache@v3
+        with:
+          path: src.tar
+          key: src.tar-${{ hashFiles('.local/share/crosstool-ng/packages') }}-${{ hashFiles('.local/share/crosstool-ng/samples') }}
+          restore-keys: |
+            src.tar-${{ hashFiles('.local/share/crosstool-ng/packages') }}-
+            src.tar-
+      - name: extract tarballs
+        run: |
+          tar -xvf src.tar || true
+          touch stamp
       - name: "prereq Linux"
         if: ${{ runner.os == 'Linux' }}
         run: |
@@ -82,12 +95,8 @@ jobs:
                 ct-ng olddefconfig; \
                 ct-ng source; \
           done
+          find src -type f -not -newer stamp -delete -print
           tar -cvf src.tar src
-      - name: "upload sources"
-        uses: actions/upload-artifact@v3
-        with:
-          name: src.tar
-          path: src.tar
 
   toolchains:
     needs: [crosstool, tarballs]
--- a/.github/workflows/build-toolchains.yml
+++ b/.github/workflows/build-toolchains.yml
@@ -55,10 +55,12 @@ jobs:
         run: |
           tar -xf ct-ng.tar
       - name: download tarballs
-        uses: actions/download-artifact@v3
+        uses: actions/cache/restore@v3
         with:
-          name: src.tar
+          path: src.tar
+          key: src.tar-${{ hashFiles('.local/share/crosstool-ng/packages') }}-${{ hashFiles('.local/share/crosstool-ng/samples') }}
       - name: extract tarballs
+        continue-on-error: true
         run: |
           tar -xvf src.tar
       - name: prereq Linux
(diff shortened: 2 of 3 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -61,6 +61,17 @@ jobs:
       - name: "extract ct-ng"
         run: |
           tar -xf ct-ng.tar
+      - name: download tarballs
+        uses: actions/download-artifact@v3
+        continue-on-error: true
+        with:
+          name: src.tar
```
cpackham, line 68: Will this work? The thing that creates src.tar hasn't run yet so I don't think it's smart enough to use the one from the last build.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -82,6 +93,7 @@ jobs:
                 ct-ng olddefconfig; \
                 ct-ng source; \
           done
+          find ./.local/src -type f -not -newer stamp -delete -print
```
cpackham, line 96: Which means this won't do anything because everything will be newer than `stamp`

cpackham (commented):
I was thinking of something more along the lines of https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows which would let us re-use the tarballs between builds.

.github/workflows/continuous-integration-workflow.yml
QBos07: You seem to be right. I wanted to test this because artifact last for 90 days und caches only for 7.

QBos07:
@cpackham shoud now work

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -61,6 +61,21 @@ jobs:
...
+          key: src.tar-${{ hashFiles('.local/share/crosstool-ng/packages') }}-${{ hashFiles('.local/share/crosstool-ng/samples') }}
+          restore-keys: |
+            src.tar-${{ hashFiles('.local/share/crosstool-ng/packages') }}-
+            src.tar-
+      - name: extract tarballs
+        if: ${{ steps.cache.outputs.cache-hit == 'true' }}
+        continue-on-error: true
+        run: |
+          tar -xvf src.tar
+          touch stamp
```
cpackham, line 78: Is this part still needed now. Won't the hashing on `packages`+`samples` be enough to ensure the cache gets dropped when we add a new version of a package?

QBos07: Yes because it can restore via restore-keys (backup-keys)

QBos07:
made out the last bugs and fixed them

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -82,6 +95,7 @@ jobs:
                 ct-ng olddefconfig; \
                 ct-ng source; \
           done
+          find src -type f -not -newer stamp -delete -print
           tar -cvf src.tar src
```
cpackham, line 99: Do we have to do anything to tell the cache to be updated with what is now in src.tar?

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -82,6 +95,7 @@ jobs:
...
       - name: "upload sources"
```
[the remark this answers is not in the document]
QBos07: changed it to the version you probably want

.github/workflows/continuous-integration-workflow.yml
QBos07: to clarify ^: there are `actions/cache/restore`, `actions/cache/save` and `action/cache` with combines the fomer 2

cpackham (approved):
Looks good now thanks. 

One thing we could follow up with later is getting rid of the src.tar thing. It was convenient when uploading/downloading an artifact but now the cache does all of that so there's no need to create a tarball of all the tarballs.

QBos07:
> One thing we could follow up with later is getting rid of the src.tar thing. It was convenient when uploading/downloading an artifact but now the cache does all of that so there's no need to create a tarball of all the tarballs.

I think we can improve further by fusing `./bootstrap` into `Makefile.am` and then making an out-of-src build and caching that.

PR #2043, opened 2023-09-24
Updates

cpackham:
Updates for mpc, linux and strace. Removal of obsolete versions


packages/make/4.2.1/version.desc
```diff
@@ -1,2 +0,0 @@
-obsolete='yes'
-archive_formats='.tar.bz2 .tar.gz'
```
egtvedt, line 2: I have run into issues on some open source project still not 100% compatible with make >= 4.3, perhaps we want to hold on for one more release?

cpackham: Sure. I'll drop this from the list.

PR #2046, opened 2023-10-09
glibc: Bring in fix for CVE-2023-4911

cpackham:
Bring in the upstream fix for the tunables CVE. The patches are taken from glibc's git repository. These needed to drop the changes to the NEWS file which would conflict due to the changes not being brought in.


cpackham:
I'm really not sure about this. CVE-2023-4911 has hit the press so it has had more publicity but I'm not bringing in the changes for CVE-2023-4527, CVE-2023-4806 or CVE-2023-5156 (or the dozen or so other changes that didn't get a CVE number). So perhaps ct-ng should be taking the stance that we're only supporting the upstream tarball plus patches to get things working on odd hardware. Rather than giving the impression that we're doing security fixes.

It was suggested on IRC that perhaps we should be providing an option to point at the glibc release branches which are intended to be a rolling release. I haven't thought about what that would look like.

QBos07:
I think upstream should backport these fixes. But I like the idea of a rolling release option

cpackham:
Yeah I'm going to self-reject this.

It's actually not too hard to point at the `release/$version/master` branch of the glibc repo. I had a quick try yesterday. Anyone concerned about these CVEs should be doing that.

It might be worth writing up a little guide on doing that.

QBos07:
That is a good idea

Issue #2084, opened 2024-01-11
'Installing native gdb': GMP is missing or unusable


ryanmwe:
crosstool-NG version: 1.26.0
Host OS:  Arch Linux x86_64 6.6.10-arch1-1
Config Log:
[config.log](https://github.com/crosstool-ng/crosstool-ng/files/13898471/config.log)
Build Log (had to compress due to file size):
[build.tar.gz](https://github.com/crosstool-ng/crosstool-ng/files/13899405/build.tar.gz)

abrodkin:
Indeed, gets reproduced with v1.26 release and the following `defconfig`:
```
CT_CONFIG_VERSION="4"
# CT_PREFIX_DIR_RO is not set
CT_ARCH_ARM=y
CT_ARCH_CPU="cortex-a8"
CT_ARCH_FPU="neon"
CT_ARCH_FLOAT_HW=y
CT_TARGET_VENDOR="cortex_a8"
CT_KERNEL_LINUX=y
CT_LINUX_V_5_2=y
CT_BINUTILS_LINKER_LD_GOLD=y
CT_BINUTILS_GOLD_THREADS=y
CT_BINUTILS_LD_WRAPPER=y
CT_BINUTILS_PLUGINS=y
CT_CC_LANG_CXX=y
CT_DEBUG_DUMA=y
CT_DEBUG_GDB=y
# CT_GDB_CROSS_PYTHON is not set
CT_GDB_NATIVE=y
CT_DEBUG_LTRACE=y
CT_DEBUG_STRACE=y
```

abrodkin:
Duplicate of https://github.com/crosstool-ng/crosstool-ng/issues/1656.
The problem happens because on the native GDB configuration for some reason host's `libgmp.so` is used, thus test fails, and we get build error.
```
configure:21844: arm-cortex_a8-linux-gnueabihf-gcc -o conftest -mcpu=cortex-a8 -mfpu=neon -mhard-float       -Wl,-EL conftest.c -lncurses -lm  /usr/lib/libgmp.so >&5
~/x-tools/arm-cortex_a8-linux-gnueabihf/bin/../lib/gcc/arm-cortex_a8-linux-gnueabihf/13.2.0/../../../../arm-cortex_a8-linux-gnueabihf/bin/ld.bfd: /usr/lib/libgmp.so: error a
dding symbols: file in wrong format
collect2: error: ld returned 1 exit status

configure:21854: result: no
[...]

Issue #1656, opened 2022-01-22
Build failed in step 'Installing native gdb'


kanakshilledar:
### Unable to proceed stuck at 'Installing native gdb'
- CrossTool-NG version:
 ```
./ct-ng version
This is crosstool-NG version 1.24.0.504_6737cfa
```
- Host OS details:
   - Endeavour OS (Arch Linux)
   - `uname -r` -> `5.16.1-arc
h1-g14-2`
   - `uname -m` -> `x86_64`
- Building for `arm-cortex_a8-linux-gnueabi` with 2 configuration changes
 - In **Paths and misc options** disabled **Render the toolchain read-only** (`CT_INSTALL_DIR_RO`)
 - In **Target options | Floating point** selected **hardware (FPU)** (`CT_ARCH_FLOAT_HW`)
 - Attached a `.config` as `config.txt` below
- During build I get the error in the step of Installing native gdb. I do have a `gdb` preinstalled on my system. The following error I get.
```
[INFO ]  Installing native gdb
[EXTRA]    Configuring native gdb
[EXTRA]    Building native gdb
[ERROR]    configure: error: GMP is missing or unusable
[ERROR]    make[3]: *** [Makefile:9752: configure-gdb] Error 1
[ERROR]    make[2]: *** [Makefile:903: all] Error 2
[ERROR]   
[ERROR]  >>
[ERROR]  >>  Build failed in step 'Installing native gdb'
[ERROR]  >>        called in step '(top-level)'
[ERROR]  >>
[ERROR]  >>  Error happened in: CT_DoExecLog[scripts/functions@376]
[ERROR]  >>        called from: do_gdb_backend[scripts/build/debug/300-gdb.sh@346]
[ERROR]  >>        called from: do_debug_gdb_build[scripts/build/debug/300-gdb.sh@168]
[ERROR]  >>        called from: do_debug[scripts/build/debug.sh@35]
[ERROR]  >>        called from: main[scripts/crosstool-NG.sh@697]
[ERROR]  >>
[ERROR]  >>  For more info on this error, look at the file: 'build.log'
[ERROR]  >>  There is a list of known issues, some with workarounds, in:
[ERROR]  >>      https://crosstool-ng.github.io/docs/known-issues/
[ERROR]  >>
[ERROR]  >>  If you feel this is a bug in crosstool-NG, report it at:
[ERROR]  >>      https://github.com/crosstool-ng/crosstool-ng/issues/
[ERROR]  >>
[...]

PR #2088, opened 2024-01-14
gdb: Explicitly point to target's libgmp for native build

abrodkin:
Starting from GDB 11.x, gmp is needed as a dependency to build full gdb. And by default build system of native GDB will try to link with libgmp of the build host. And to make sure that doesn't happen we need to specify location of the target's sysroot so that library search starts from there. Which we do in that change.

Fixes [1] & [1].

[1] https://github.com/crosstool-ng/crosstool-ng/issues/2084
[2] https://github.com/crosstool-ng/crosstool-ng/issues/1656

### DIFF
--- a/scripts/build/debug/300-gdb.sh
+++ b/scripts/build/debug/300-gdb.sh
@@ -122,6 +122,14 @@ do_debug_gdb_build()
 
         native_extra_config+=("--program-prefix=")
 
+        # Starting from GDB 11.x, gmp is needed as a dependency to build full
+        # gdb. And if target GMP gets built, explicitly point to installed library,
+        # as otherwise host library might be attempted to be used for target binary
+        # linkage.
+        if [ "${CT_GMP_TARGET}" = "y" ]; then
+            native_extra_config+=("--with-libgmp-prefix=${CT_SYSROOT_DIR}")
+        fi
+
         # gdbserver gets enabled by default with gdb
         # since gdbserver was promoted to top-level
         if [ "${CT_GDB_GDBSERVER_TOPLEVEL}" = "y" ]; then


scripts/build/debug/300-gdb.sh
```diff
@@ -122,6 +122,14 @@ do_debug_gdb_build()
...
         native_extra_config+=("--program-prefix=")
 
+        # Starting from GDB 11.x, gmp is needed as a dependency to build full
+        # gdb. And if target GMP gets built, explicitly point to installed library,
+        # as otherwise host library might be attempted to be used for target binary
+        # linkage.
+        if [ "${CT_GMP_TARGET}" = "y" ]; then
+            native_extra_config+=("--with-libgmp-prefix=${CT_SYSROOT_DIR}")
+        fi
+
```
cpackham, line 132: I wonder if this should be handled like zstd below i.e. `--with-libgmp=${complibs}`. It's slightly different because zstd is optional but GMP is required.

QBos07: IIRC it's `--with-libgmp-prefix`

abrodkin: > I wonder if this should be handled like zstd below i.e. `--with-libgmp=${complibs}`. It's slightly different because zstd is optional but GMP is required.

Well, that's how it's done in Buildroot, see https://git.buildroot.net/buildroot/tree/package/gdb/gdb.mk#n161.
Do you want me to try something else?

abrodkin:
@cpackham do you want me to still do something here?

cpackham:
I want to take another look. It appears that this doesn't solve the
original problem. But I think something like it would, just maybe we need
to point it at a different directory.

ryanmwe:
@cpackham this actually did fix the issue for me in the end.

PR #2108, opened 2024-02-08
Remove Oracle UEK linux dependencies.

cupermir:
This patch removes any dependency to the Oracle UEK Linux sources since it can be easily replaced by a standard kernel explicitly pointing to the exact kernel version, as the toolchain building only requires the kernel headers.


cupermir:
I do not believe those failures are relative to the changes presented in the PR.
Can someone please verify what happened with the CI.

Thanks!

PR #2117, opened 2024-02-28
Check for bash version 4 or higher

### LINKED ISSUE #2115: Configure script does not enforce bash version
- crosstool-NG version: 1.26.0
- Host OS: macOS (I know this is unsupported but this issue is not macOS specific)

Hi, I was getting an error when building like "bad substitution".  Tracked it down to this line: https://github.com/crosstool-ng/crosstool-ng/blob/master/scripts/build/companion_tools.sh#L8.  

As it turns out, this line is using a feature [added](https://github.com/bminor/bash/blob/f3b6bd19457e260b65d11f2712ec3da56cef463f/CHANGES#L5256) in Bash 4.0, and I had bash 3.2.  That's fine, it's actually pretty ridiculous that OS X doesn't include bash 4.0 or newer by default, and it's easily fixable via `brew install bash`.  

However, it would have saved me some time if the [check in the configure script ](https://github.com/crosstool-ng/crosstool-ng/blob/master/configure.ac#L175) checked for Bash 4.0 and reported an error since I didn't have it.  It would be great if you could look into this!

QBos07:
We use bash 4 features in some scripts

Signed-off-by: Quentin Boswank <qubos@outlook.de>

Fixes #2115

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -171,12 +171,12 @@ AS_IF([$PATCH --no-backup-if-mismatch </dev/null >/dev/null 2>&1],
       AC_MSG_RESULT([yes]),
       AC_MSG_ERROR([$PATCH does not support --no-backup-if-mismatch]))
 
-# We need a bash that is >= 3.1
+# We need a bash that is >= 4.0
 CTNG_PROG_VERSION_REQ_STRICT([BASH_SHELL],
-    [GNU bash >= 3.1],
+    [GNU bash >= 4.0],
     [bash],
     [bash],
-    [^GNU bash, version (3\.[1-9]|4|5)])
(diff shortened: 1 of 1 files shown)


cpackham (approved):

configure.ac
```diff
@@ -173,10 +173,10 @@ AS_IF([$PATCH --no-backup-if-mismatch </dev/null >/dev/null 2>&1],
 
 # We need a bash that is >= 3.1
```
cpackham, line 174: Comment needs updating

cpackham (changes requested):

cpackham (approved):

PR #2121, opened 2024-03-05
Fix homebrew path

bhundven:
Homebrew changed it's default install path from `/usr/local` to `/opt/homebrew` a while back. Hardcoding the path is a bad idea, so instead use `$(brew --prefix)` to get the prefix of the path for tools.

Also fix some [shellcheck](https://github.com/koalaman/shellcheck) issues.


bhundven:
https://github.com/actions/runner-images/blob/main/images/macos/scripts/build/install-homebrew.sh

cpackham (approved):

cpackham (approved):
Snap! https://github.com/crosstool-ng/crosstool-ng/pull/2106. As your version has some shellcheck fixes lets go with that (I'll close my PR). 

We also seem to be running out of space on the canadian cross builds. Initially I thought this was on the actionsv4 due to nodejs growing but it just seems that we're generally too big for the free tier runners now.

PR #2133, opened 2024-03-21
Add strace-6.7 and strace-6.8

bhundven:
https://github.com/strace/strace/releases


cpackham:
We can probably just take strace-6.8. Strace releases are backwards compatible so we only need to keep the latest one (and the one that we have in the latest ct-ng release as per our obsolescence policy)

bhundven:
Ok, I'll update the PR

PR #2137, opened 2024-03-25
Make package reproducible on arch linux

andrin-geiger:
Currently this package is not fully reproducible on arch linux. See the [diffoscope](https://reproducible.archlinux.org/api/v0/builds/594872/diffoscope).
The issue is the recorded timestamp in the gzip header which can easily be ignored with the [no-name](https://wiki.debian.org/ReproducibleBuilds/TimestampsInGzipHeaders) flag for gzip.


Antiz96:
Friendly ping about this MR :)

I [backported the patch](https://gitlab.archlinux.org/archlinux/packaging/packages/crosstool-ng/-/commit/ff371cbb5f8990a3ecc70155a2cb68db5b18713e) on the Arch Linux side to make the package reproducible in the mean time.

PR #2143, opened 2024-04-18
mingw-w64: enable wildcard

mati865:
Wildcard is an opt-in (disabled by the default) feature that is used by many GNU tools like Binutils.

Signed-off-by: Mateusz Mikuła <mati865@gmail.com>

### DIFF
--- a/scripts/build/libc/mingw-w64.sh
+++ b/scripts/build/libc/mingw-w64.sh
@@ -208,6 +208,7 @@ mingw_w64_main()
         --prefix=${MINGW_INSTALL_PREFIX} \
         --build=${CT_BUILD} \
         --host=${CT_TARGET} \
+        --enable-wildcard \
         "${crt_opts[@]}"
 
     # mingw-w64-crt has a missing dependency occasionally breaking the


scripts/build/libc/mingw-w64.sh
```diff
@@ -208,6 +208,8 @@ mingw_w64_main()
         --prefix=${MINGW_INSTALL_PREFIX} \
         --build=${CT_BUILD} \
         --host=${CT_TARGET} \
+        --enable-wildcard \
+        --disable-dependency-tracking \
```
cpackham, line 212: I'm fine with the wildcard change but we should keep dependency-tracking enabled

cpackham (changes requested):

scripts/build/libc/mingw-w64.sh
mati865: I think crosstool-ng always cleans up build directory so there is no benefit in tracking the changes but dropped this change.

cpackham (approved):

PR #2147, opened 2024-05-10
Fix MacOS CI pre-requisites

nikkon-dev:
- add missing texinfo
- add missing libtool


cpackham:
I think maybe the failures are due to macos-latest moving from macos-13 to macos-14. Instead of playing whack-a-mole with issues I'll pin us back to macos-13

PR #2149, opened 2024-05-11
CI: pin macos runners to macos-12

cpackham:
macos-latest has recently (as of April 2024) started transitioning to macos-14 (previously it was macos-12). This seems to be missing things we rely on (e.g. makeinfo) and even when those are added throws other errors.  For now lets pin things to macos-12 until someone can spend some time looking at why we can't use macos-14.


cpackham:
https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/

PR #2156, opened 2024-05-18
musl: Mark not EXPERIMENTAL

cpackham:
Musl was marked experimental in commit 08d91d41 ("musl: config is broken for !EXPERIMENTAL"). Most of the reasoning for that change no longer applies and as it's been about 8 years it's time to let musl loose on the world. Drop the `depends on EXPERIMENTAL` and update the sample configs for aarch64 and x86_64.


cpackham:
I did find that powerpc64 didn't build cleanly. https://musl.libc.org/doc/1.1.24/manual.html says it should be supported

PR #2157, opened 2024-05-22
gcc: Add GCC 13.3.0

cpackham:
https://gcc.gnu.org/pipermail/gcc-announce/2024/000181.html

Add the new version dropping patches that have been applied upstream.


packages/gcc/13.3.0/0012-libgcc-m68k-Fixes-for-soft-float.patch
```diff
@@ -39,7 +37,7 @@ index fe41edf26aa..eaa7858493c 100644
  #define EXP(fp)		(((fp) >> 23L) & 0xFF)
  #define MANT(fp)	(((fp) & 0x7FFFFFL) | HIDDEN)
  #define PACK(s,e,m)	((s) | ((e) << 23L) | (m))
-@@ -262,6 +263,9 @@ __extendsfdf2 (float a1)
+@@ -262,6 +263,9 @@
```
egtvedt, line 40: Is the patches refreshed correctly? You have lost some meta information about function changed here.

I really like if you could keep it, as it helps when moving patches to future releases.

cpackham: Yes and no. Some of the patches are plain patches others are exported from git. When updating I use the `maintainers/manage-packages.sh` script which pushes them all into quilt to check that they all still apply. The result is then exported. The quilt patch format is a bit different to git so some metadata is preserved and some isn't. I'll see if I can be a bit more selective about the final changes.

egtvedt: I've used that script in the past, and it has kept that specific metadata.

I use quilt refresh locally in other projects with args `-p ab -u --no-timestamps --no-index --strip-trailing-whitespace`

I'm fine with whatever you end up with, just wanted to make sure we didn't miss information if the automated tools would include them.

PR #2158, opened 2024-05-23
config: add dependency on cmake

egtvedt:
We want to add mold linker support to crosstool-ng, which will require cmake while building. Hence add a dependency to cmake.

Relevant for https://github.com/crosstool-ng/crosstool-ng/pull/2150

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -153,6 +153,10 @@ AC_CHECK_PROGS([curl], [curl])
 CTNG_SET_KCONFIG_OPTION([curl])
 AC_SUBST([curl])
 
+AC_CHECK_PROGS([cmake], [cmake])
+CTNG_SET_KCONFIG_OPTION([cmake])
+AC_SUBST([cmake])
+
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -153,6 +153,10 @@ AC_CHECK_PROGS([curl], [curl])
 CTNG_SET_KCONFIG_OPTION([curl])
 AC_SUBST([curl])
 
+AC_CHECK_PROGS([cmake], [cmake])
+CTNG_SET_KCONFIG_OPTION([cmake])
+AC_SUBST([cmake])
+
```
cpackham, line 159: I think instead of this we should do something more like https://github.com/crosstool-ng/crosstool-ng/commit/4442dc87dd5a7c237e465fd873dc40928749610a

Same applies for meson and ninja but this wasn't possible at the time those were added as dependencies.

egtvedt:
Should implement directly in https://github.com/crosstool-ng/crosstool-ng/pull/2150

PR #2164, opened 2024-06-02
Fix typo retrun→return in 0001-fix-configure-for-gcc14.patch

tttapa:
Fixes a typo introduced in d97a1ecdf0058251c33f6e98ee30d0ca9c93d6ab.


cpackham:
good catch thanks.

PR #2165, opened 2024-06-03
scripts/functions: Update sourceware mirrors

cpackham:
Add https mirror for sourceware.org. Use https instead of http for kernel.org and gnu.org.


QBos07:
Maybe we should put these over ftp since that is unencrypted

cpackham:
We definitely want https since that gives some assurance that the data
isn't changed in flight. I almost removed the ftp mirrors as they're being
phased out by most website operators.

PR #2168, opened 2024-06-05
Optimize zstd we build

QBos07:
- multithreading enabled
- release mode


cpackham:
Sorry hit the wrong button when merging. You might get a couple of odd notifications

PR #2176, opened 2024-06-23
gcc: Add GCC 12.4.0

temap:
https://gcc.gnu.org/pipermail/gcc-announce/2024/000182.html

Add the new version dropping patches that have been applied upstream.


cpackham:
I think you've dropped a few too many patches.

cpackham:
While I was checking which patches were still needed I ended up just doing the update in #2177

temap:
Thanks @cpackham . Did you reapply the patches manually, or did you use the maintainer's script?

cpackham:
I applied them to gcc 12.3 and rebased the branch against 12.4 and ran `git format-patch` which gave me those 11 remaining patches (which are the ones I expected). I then used a combination of `maintainer/manage-packages.sh` and manual edits to clean them up a little bit more.

PR #2196, opened 2024-08-26
prefer default env: CT_LOCAL_TARBALLS_DIR

valord577:
refs https://github.com/crosstool-ng/crosstool-ng/pull/2196#issuecomment-2310830179

### DIFF
--- a/config/global/paths.in
+++ b/config/global/paths.in
@@ -5,7 +5,7 @@ comment "Paths"
 config LOCAL_TARBALLS_DIR
     string
     prompt "Local tarballs directory"
-    default "${HOME}/src"
+    default "${CT_LOCAL_TARBALLS_DIR:-${HOME}/src}"
     help
       If you have previously downloaded the tarballs, enter the PATH where
       you stored them here.


valord577:
@cpackham Could you please take a time for code review?

QBos07:
Great idea, but you are setting the variable to itself wich may not be obvious but Kconfig adds the `CT_` prefix. If you really need your env to be taken do it with this:
```sh
env >> .config
ct-ng build
```

It is  just `source .config` internally

valord577:
> Great idea, but you are setting the variable to itself wich may not be obvious but Kconfig adds the `CT_` prefix. If you really need your env to be taken do it with this:
> 
> ```shell
> env >> .config
> ct-ng build
> ```
> 
> It is just `source .config` internally

@QBos07 

`CT_LOCAL_TARBALLS_DIR` is documented at https://crosstool-ng.github.io/docs/configuration/, but it is not really implemented.

image

QBos07:
I think you understand the docs wrong. These aren't your normal environment variables (none on that site are), these are the configuration key names. You can look them up by pressing `?` on a specific configuration entry in `ct-ng menuconfig`. If you want to set config options you have to append them to `.config` or directly change the already existing entry in there. Everything documented at https://crosstool-ng.github.io/docs/configuration/ is only to be used from the inside of crosstool-ng!

Here is an slightly convoluted example of how to change the config at runtime: https://github.com/QBos07/CP-dockercross/blob/main/Dockerfile#L33. You can just write to the `.config` there directly

valord577:
> I think you understand the docs wrong. 

@QBos07 Yeah, you are right. I does understand the docs wrong. Thank you very much for the detailed explanation. I will edit the `.config` file, then use it.

PR #2211, opened 2024-09-30
zlib: Update to zlib-1.3.1

WXbet:
add zlib 1.3.1 and remove version 1.3


cpackham:
Looks good to me. Could you also add another commit to this series to remove the zlib-1.3 release (we haven't released an official ct-ng version with zlib-1.3 so no need to worry about marking it as obsolete)

WXbet:
> Could you also add another commit to this series to remove the zlib-1.3 release

Done

PR #2219, opened 2024-10-18
Drop ol7u9 support

cpackham:
Oracle Linux 7 was due for end of support in July 2024. The configurations don't currently build because we don't have an old enough version of ISL in crosstool-ng. Drop the ol7u9 samples and the binutils-oracle/2.27 and glibc-oracle/2.17 packages which the ol7u9 configs were the only users of.


cpackham:
@cupermir, @gemaq you've had changes keeping OL7 around. I'm not against keeping it past the EOS date but the easiest thing for me to do is drop the currently broken configs.

cupermir:
Hi @cpackham,

Please allow me a few days to check if it makes sense to update libisl and extend the support for ol7.

Thanks

cupermir:
Please do not remove ol7-u9 from ct-ng.
I think the proper solution would be to use the libisl and libcloog that is already present in [GCC branch for ol7-u9](https://github.com/oracle/gcc/tree/oracle/gcc/ol7-u9).
Please allow me some time to work on the required code changes to do that.

cpackham:
Another option that might work would be to add isl 0.14

PR #2223, opened 2024-10-25
Add support for gcc --enable-multilib-space option

keith-packard:
Set CT_MULTILIB_SPACE to pass --enable-multilib-space during GCC build. When set, also build
dependent libraries in multilib mode, even if CT_MULTILIB is not set.


keith-packard:
sorry for the noise; we're messing with this in Zephyr, but I want to wait and see what upstream gcc does with this option before submitting a crosstool-ng PR.

PR #2227, opened 2024-11-14
mold: tell cmake which compilers to use

novas0x2a:
The mold build was using gcc in the PATH, instead of the shiny gcc we just built (the way the other companion libs/tools do)


egtvedt (approved):
Works for me, nice upgrade.

PR #2229, opened 2024-11-17
samples: Disable GRAPHITE for ol7 configs

cpackham:
Graphite loop optimizations require the isl and cloog libraries. ct-ng doesn't have an old enough version of these for the ol7 version of GCC. Disabling GRAPHITE support allows the toolchains to build.


cpackham:
@cupermir this is an alternative to #2219 it keeps the ol7u9 toolchains albeit with a slight loss of functionality.

cpackham:
I've gone with #2260 for now. That'll get us to a release. After that I'm going to merge (and updated version of) #2219 to master.

PR #2242, opened 2024-12-22
Enable install of .exe binaries in dtc package

toriningen:
Fixes https://github.com/crosstool-ng/crosstool-ng/issues/2241


QBos07:
Maybe consider upstreaming this

packages/dtc/1.4.7/0001-enable-install-exe.patch
```diff
@@ -0,0 +1,31 @@
...
++++ b/Makefile
+@@ -199,7 +199,13 @@ endif
+ install-bin: all $(SCRIPTS)
+ 	@$(VECHO) INSTALL-BIN
+ 	$(INSTALL) -d $(DESTDIR)$(BINDIR)
+-	$(INSTALL_PROGRAM) $(BIN) $(DESTDIR)$(BINDIR)
++	for bin in $(BIN); do \
++	  if [ -f "$$bin" ]; then \
++	    $(INSTALL_PROGRAM) $$bin $(DESTDIR)$(BINDIR); \
++	  elif [ -f "$$bin.exe" ]; then \
```
cpackham, line 22: The way this is usually done would involve having a variable (e.g. `EXEEXT`) that is defined to nothing on Linux/BSD/etc and defined to `.exe` on msys/windows. Regardless see what you can land upstream first then ct-ng can carry a backport.

toriningen: > The way this is usually done would involve having a variable (e.g. `EXEEXT`) that is defined to nothing on Linux/BSD/etc and defined to `.exe` on msys/windows. Regardless see what you can land upstream first then ct-ng can carry a backport.

That's the route I wanted to go, but the challenge is that technically this is happening on Linux, we are cross-compiling for Windows, so uname won't match the extension.

I suppose it's configure's job to detect EXEEXT, but I can in theory do this in Makefile myself with something like this?

```
EXEEXT := $(shell echo | "$(CC)" -x c - '-###' -o foobar 2>&1 | \
	    sed -En '/-o foobar[^ ]*/s/.*-o foobar([^ ]*).*?/\1/p' | \
	    head -1)
```

PR #2246, opened 2024-12-27
lm32: Initial LatticeMicro32 support

FlyGoat:
This target is in GCC/binutils for a while.

It's baremetal only without upstream Linux support.


cpackham:
Looks like this might need a rebase and push -f to get rid of that unnecessary merge. Other than that looks good.

FlyGoat:
> Looks like this might need a rebase and push -f to get rid of that unnecessary merge. Other than that looks good.

Thanks for reviewing, done.

cpackham (approved):

PR #2253, opened 2025-01-05
c6x: pin to GCC 13

cpackham:
The custom uclibc has compile errors with GCC 14. Ignoring that even a bare metal config fails to build (assembler errors in libgcc) with GCC
14. For now pin the sample config to GCC 13 since that seems to work.


cpackham:
Upstream bug report https://gcc.gnu.org/bugzilla/show_bug.cgi?id=118299

cpackham:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114910

PR #2279, opened 2025-02-03
Add GMP 6.3.0

dkm:
GMP 6.2.1 may fail some autoconf test when using GCC master (to-be-15) branch.


dkm:
Haven't forward ported the x18 patch found in 6.2.1, not really sure how to test (ping @galak ?)

FYI, all nightly cross gcc builds are failing because of this (see https://github.com/compiler-explorer/compiler-explorer/issues/7360)

cpackham:
Looks like the x18 changes have been incorporated upstream from the ChangeLog "mpn/arm64: Avoid the x18 register since it is reserved on Darwin."

dkm:
> Looks like the x18 changes have been incorporated upstream from the ChangeLog "mpn/arm64: Avoid the x18 register since it is reserved on Darwin."

Thanks for checking!

dkm:
Also, have only added 2 types of archives and left zst alone as I couldn't find its corresponding gpg signature.

dkm:
Something's off, the builds are still failing in our docker images, but I got it working locally. I'll need to better investigate, but maybe you can wait before merging this... At least, it's lowest prio until I understand what's happening.

cpackham:
Hmm. I was planning on doing the 1.27.0 release this week. I've kicked off
a build-all locally with this update but maybe it's best to hold off and do
the gmp bump after the release.

dkm:
I'll let you decide... It should not break anything.
For some reason, when building with ct-ng, the GMP configure fails (even for 6.3.0) when I have a fresh build of GCC in `$PATH`, but I can't reproduce the error when re-executing the configure command, there must be something else that makes it fails, but I don't see it.

QBos07:
Did you have the same environment as the failed configure? There are a lot of variables that get passed thought the environment

dkm:
> Did you have the same environment as the failed configure? There are a lot of variables that get passed thought the environment

I guess not, but I don't really know how to easily debug this. My goto solution for this is to `set -x` the configure script and read the huge log. But maybe there's something easier with ct-ng. I believe it may be possible to have ct-ng start a shell in the correct env?

cpackham:
There's a few options under the "Debug crosstool-NG". You might have some luck with "Interactive shell on failed commands" although I've never actually used it personally.

The build-all actually seems to be going fairly well so I'm inclined to merge this before the release. Sounds like there may be some more problems lurking for GCC-15 but we'll cross that bridge when we come to it.

dkm:
Yes, it's probably gcc-15. I've been able to reproduce, and it's caused by a conftest C code that does not compile anymore:
```
--- gmp-6.3.0/configure 2023-07-29 15:42:22.000000000 +0200
+++ gmp/configure       2025-02-03 22:00:33.567817886 +0100
@@ -6568,7 +6566,7 @@
 
 #if defined (__GNUC__) && ! defined (__cplusplus)
 typedef unsigned long long t1;typedef t1*t2;
-void g(){}
+void g(int, const t1*, t1, t1*, const t1*, int){}
 void h(){}
 static __inline__ t1 e(t2 rp,t2 up,int n,t1 v0)
 {t1 c,x,r;int i;if(v0){c=1;for(i=1;i<n;i++){x=up[i];r=x+1;rp[i]=r;}}return c;}
```
This fixes the GMP build, but I'm afraid to get more errors down the road :sweat_smile:

cpackham:
Yeah I've just got through (what I hope are) the last issues like that for GCC 14. configure checks that fail outright are OK to deal with. The nefarious ones are where the check fails but then the functionality is back-filled from gnulib so you end up with some other error way down the track.

dkm:
FYI, this GMP issue with gcc 15 is fixed in gmp now: https://gmplib.org/list-archives/gmp-bugs/2025-February/005560.html
Meanwhile, I've applied a patch to fix our builds: https://github.com/compiler-explorer/gcc-cross-builder/blob/main/build/patches/crosstool-ng/0001-gcc-15-fix.patch

cpackham:
#2285 is an adapted version (+configure -ChangeLog) of the upstream change

PR #2305, opened 2025-03-08
zstd: Bring in upstream fix for MSYS

cpackham:


QBos07:
Nice to see that my work is appreciated

cpackham:
Very much so :). I'm surprised we haven't hit this as a bug report yet.

PR #2314, opened 2025-04-01
add samples for mips64-ol9u5

peter-shoes:
Hello! Here is an additional sample for OL9.5 configured for mips64.


peter-shoes:
Hello, it looks like all checks have passed -- just pinging this :-)

cpackham:
Running a `ct-ng build-all` I get a failure with this config

cpackham:
[build.log.gz](https://github.com/user-attachments/files/19919929/build.log.gz)

peter-shoes:
Hey Chris, yes it looks like this is the error we were seeing on ol8 machines where building GCC with PIE ended up being a problem. I sent an update to the patch that removes the --enable-host-pie flag by default. It looks like that change was merged into master.

PR #2320, opened 2025-04-05
Pie fix ol8

peter-shoes:
Hello, the previous version of this sample would build successfully for ol9 (and Ubuntu LTS), but not ol8. This allows for building on ol8, as well as the other platforms.


cpackham:
Could you rebase this to get rid of the extra merge.

PR #2332, opened 2025-04-16
musl/loongarch64: Add hwcap header

wojiushixiaobai:


cpackham:
Can we get a bit more of a description in the commit message. Is this an upstream patch being back ported?

wojiushixiaobai:
> Can we get a bit more of a description in the commit message. Is this an upstream patch being back ported?

Yes, this is an upstream patch being back ported.

cpackham:
> > Can we get a bit more of a description in the commit message. Is this an upstream patch being back ported?
> 
> Sorry, this is not an upstream patch being back ported. maybe I need to add it to the upstream first.

Yes my preference is for patches to go upstream first. I'm happy to carry things as back ports. This seems to have been accepted upstream just hasn't shown up on the public git repo.

cpackham:
Looks like this has landed https://git.musl-libc.org/cgit/musl/commit/?id=b0dc340b3865a9b980e3b9a6938fdec6fe10fba2

PR #2334, opened 2025-04-21
gcc: Allow addition of -fno-exceptions when building libstdc++

keith-packard:
Newlib-nano has a configuration variable which controls additional flags used when building libstdc++, such as `-fno-exceptions`.  Add two new variables to allow this to be added to any libstdc++ build as well as to the picolibc-specific build when picolibc is added as a companion library: `CC_GCC_LIBSTDCXX_TARGET_CXXFLAGS` and `LIBC_PICOLIBC_GCC_LIBSTDCXX_TARGET_CXXFLAGS`


QBos07:
Isn't `CT_CC_GCC_ENABLE_CXX_FLAGS` meant to do the same already? What's the difference here exactly

PR #2344, opened 2025-04-25
gcc: Add 15.1.0

egtvedt:
https://gcc.gnu.org/pipermail/gcc-announce/2025/000185.html

See upstream changes at https://gcc.gnu.org/gcc-15/changes.html

Add the new version clean up the patches. Remove GCC 14 patches applied to GCC 15.


cpackham:
beat me to it :)

cpackham:
Could you add a milestone entry for this to `packages/gcc/package.desc`. I already added an entry to `config/arch/nios2.in` to handle the removal.

cpackham:
I think the first batch of build failures are due to some ./configure tests failing to pick up the right functions and using bad fall-backs.

egtvedt:
> Could you add a milestone entry for this to `packages/gcc/package.desc`. I already added an entry to `config/arch/nios2.in` to handle the removal.

Ah, didn't know, included now.

tttapa:
@cpackham Could you comment on why the massive ncurses patches are needed? Close to half a million (!) LOC of new patches seems excessive for a compiler upgrade.

Could (some of) these patches be avoided by explicitly using `-std=gnu17` when building ncurses? https://gcc.gnu.org/gcc-15/porting_to.html

cpackham:
@tttapa yes it's not just support for GCC15 it's basically ncurses updates since 6.5. I did briefly think about trying to reduce the update to just the fixes for the `bool` issues we were seeing but I figure there are other fixes in there we probably want too.

I don't think it's quite as bad as it seems. Some of it seems bigger because of the way the patches have been exported so there is a lot more context than a the `diff -u3` you'd get out of git. Fedora 42 points at the full source snapshot https://invisible-island.net/archives/ncurses/current/ncurses-6.5-20250426.tgz but that's not part of the GNU mirrors so it's a toss up between carrying those patches or taking away access to the source via the GNU mirrors.

cpackham:
The overall ncurses changes are `673 files changed, 38646 insertions(+), 26450 deletions(-)`  a good majority of that  in the docs

cpackham:
Ran a build-all over the weekend and the only failures were
```
  - i686-w64-mingw32,nios2-spico-elf
  - mips64-ol9u5-linux-gnu
  - powerpc-unknown-linux-uclibc,m68k-unknown-uclinux-uclibc
  - x86_64-multilib-linux-uclibc,powerpc-unknown-elf
```

I think the mips64-ol9u5-linux-gnu is pre-existing. I'll create issues for them and attach the build.log

PR #2353, opened 2025-05-05
Bump libexpat and kernel versions

zagura:


packages/expat/package.desc
```diff
@@ -4,3 +4,4 @@ bootstrap='./buildconf.sh && make -C doc all'
 mirrors='http://downloads.sourceforge.net/project/expat/expat/${CT_EXPAT_VERSION} https://github.com/libexpat/libexpat/releases/download/R_${CT_EXPAT_VERSION//./_}'
 archive_formats='.tar.xz .tar.lz .tar.bz2 .tar.gz'
 relevantpattern='*.*|.'
+milestones='2.7'
```
cpackham, line 7: The milestone isn't strictly necessary unless we have something that requires a specific version. No harm to have it though. So not worth re-spinning unless there's something else to change.

zagura: I wasn't sure about that, put it due to version number change in filename: https://github.com/libexpat/libexpat/blob/R_2_7_1/expat/Changes#L168-L170
If it doesn't require milestones, I would remove it.

PR #2354, opened 2025-05-06
packages:mold: add version 2.39.0

egtvedt:
Add newly released mold version 2.39.0 from upstream https://github.com/rui314/mold

2.39.0
New Features
- [ARM32] Support for 32-bit big-endian ARM has been added. Although running ARM32 in big-endian mode is very rare, the processor does technically support both little- and big-endian modes, and we now support both.

- There are two variants of big-endian mode for ARM32: BE32 and BE8. BE32 is now obsolete and uses big-endian format for both instructions and data. In BE8, instructions are always in little-endian (i.e., the same as little-endian ARM32), while only the data is in big-endian. mold supports only BE8 output.

Bug Fixes and Compatibility Improvements
- Fixed a spurious --no-allow-shlib-undefined error.

- [ARM][PPC] Fixed a regression introduced in 2.38.0 that mold could crash when linking a large program.

- Previously, --default-symver didn't set versions to symbols if the symbols were marked as global: in a version script. Now, --default-symver correctly version all symbols with the soname of the output file.

- [RISC-V] Fixed an issue where mold reported an error on R_RISCV_32 when the target was 64-bit RISC-V.

- [RISC-V] Fixed an issue where a call to an weak undefined symbol within the same shared library was mistakenly turned into an infinite loop. Now, such calls are promoted to a function call through the PLT entry.

- Fixed an issue that mold falls into an infinite loop in a rare occasion when computing an address of the program header.


egtvedt:
Bumped to bugfix release 2.39.1

PR #2355, opened 2025-05-11
Build companions on gcc15

zagura:
Use -std=c17 for companion libs, which won't build with C23 standard, default for gcc15

### DIFF
--- a/scripts/build/companion_libs/100-gmp.sh
+++ b/scripts/build/companion_libs/100-gmp.sh
@@ -122,6 +122,8 @@ do_gmp_backend() {
         extra_config+=("--with-pic")
     fi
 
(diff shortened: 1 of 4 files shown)


scripts/build/companion_libs/100-gmp.sh
```diff
@@ -122,6 +122,10 @@ do_gmp_backend() {
         extra_config+=("--with-pic")
     fi
 
+    if [ "${CT_GCC_15_or_later}" = "y" ]; then
+        cflags="${cflags} -std=c17"
```
egtvedt, line 126: Should we do `-std=gnu17` instead? Or will that cause compatibility problems for non-Linux builds.

PR #2388, opened 2025-06-15
Updates on zstd

zagura:
* https://zstd.net won't connect. http://zstd.net redirects to project site (github pages from facebook) without files, on sourceforge there is mirror with matching files.
* Add zstd 1.5.7


cpackham (commented):
If you want this to auto close the issue you'll need

```
Fixes #2387
```

On the line just before your sign-off. I'm a little curious though as the github download URL should have been used first.

zagura:
Without original build.log I can only guess that first link was temporarily failing. For testing if second option works, I've null-routed github's IP.

Issue #2163, opened 2024-06-02
ct-ng build expects exactly Python 3.11


trobanga:
I was trying to build aarch64-rpi4-linux-gnu on Arch Linux using ct-ng 1.26.0 but it failed during building gcc because it was expecting Python 3.11 but I have 3.12. 

Fixed it by installing 3.11 but I think it should work with any Python >= 3.x

cpackham:
Have you got an example of the failure. Yes we should absolutely support python versions other than 3.11.

cpackham:
I have seem something that might be what you're reporting. It's not that it expects python 3.11. It's that when it goes looking for the python libraries it fails to find it and repeatedly uses the wrong check for every possible version. I would have expected in your case the configure step to be repeatedly doing `checking for main in -l/usr/bin/python3.12... (cached) no`.
I'm looking at a fix for this but it may or may not be the problem you're actually reporting.

PR #2400, opened 2025-06-17
m4/ctng_python.m4: clear cached variables

KirillSmirnov:
Prevent cached variables interfering with checks made in separate loop iterations.

This patch fixes python detection in case a user has installed multiple python versions.

### DIFF
--- a/m4/ctng_python.m4
+++ b/m4/ctng_python.m4
@@ -63,6 +63,11 @@ AC_DEFUN([CTNG_PYTHON],
   for python in python3.13 python3.12 python3.11 python3.10 python3.9 python3.8 python3.7 dnl
 python3.6 python3.5 python3.4 python3.3 python3.2 python3.1 python3.0 python2.7 dnl
 python2.6 python2.5 python2.4 python2.3 python2.2 python2.1 python; do
+    AS_UNSET(ac_cv_path_PYTHON_BIN)
+    AS_UNSET(PYTHON_BIN)
+    AS_UNSET(ac_cv_prog_PYTHON_BIN_)
+    AS_UNSET(PYTHON_BIN_)
+
     AC_PATH_PROGS(PYTHON_BIN, [$python])
     AC_CHECK_PROGS(PYTHON_BIN_, [$python])
     ctng_python_bin=$PYTHON_BIN


KirillSmirnov:
Fixes the issue for my host configuration, might fix #2163 as well

m4/ctng_python.m4
```diff
@@ -63,6 +63,11 @@ AC_DEFUN([CTNG_PYTHON],
   for python in python3.13 python3.12 python3.11 python3.10 python3.9 python3.8 python3.7 dnl
 python3.6 python3.5 python3.4 python3.3 python3.2 python3.1 python3.0 python2.7 dnl
 python2.6 python2.5 python2.4 python2.3 python2.2 python2.1 python; do
+    AS_UNSET(ac_cv_path_PYTHON_BIN)
+    AS_UNSET(PYTHON_BIN)
+    AS_UNSET(ac_cv_prog_PYTHON_BIN_)
+    AS_UNSET(PYTHON_BIN_)
+
```
cpackham, line 70: Hmm I'm a bit dubious about this. Seems to invalidate the purpose of having a cache. I think the right thing is probably to do what I hinted at in https://github.com/crosstool-ng/crosstool-ng/commit/40c7e23be89bfb47fbae28e1db46f524bf5b8ed0 and use the upstream AX_PYTHON and come up with a better way of getting the path.

cpackham:
I think #2163 was fixed by https://github.com/crosstool-ng/crosstool-ng/commit/40c7e23be89bfb47fbae28e1db46f524bf5b8ed0 but I'm not sure

cpackham:
Yeah I think there is a problem if you have multiple pythons installed. The
official docker images are like that. They have the python from whatever
Debian they're based on then they have the version you're targeting in
/usr/local.

The checks get confused by the base version (if it's newer) then fail to
find the libraries because they aren't there for that version (which is the
only version checked because of the caching).

All the same problems exist in the upstream AX_PYTHON so it's not the magic
fix I was hoping for. I wonder if we can replace it with something based
around python3-config. May not help as I think gdb uses AX_PYTHON so even
if we can detect the right python gdb may not.

QBos07:
Also if possible this should be a menu config time check

PR #2402, opened 2025-06-20
binutils: Disable gprofng

ismaell:
It contains unportable code that fails to build against implementations
other than glibc.

### DIFF
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -167,6 +167,9 @@ do_binutils_backend() {
         extra_config+=("--disable-multilib")
     fi
 
+    # Disable gprofng, it contains unportable code
+    extra_config+=("--disable-gprofng")
+
     # Disable gdb when building from the binutils-gdb repository.
     extra_config+=("--disable-sim")
     extra_config+=("--disable-gdb")


cpackham:
Could you explain a bit more about what problem this is solving? Is this a
problem for a non-glibc (e.g. alpine) build machine?

In terms of implementation this should be controlled by a menuconfig option.

ismaell:
On 20/Jun/2025 13:14, Chris Packham wrote:
> cpackham left a comment (crosstool-ng/crosstool-ng#2402)
> 
> Could you explain a bit more about what problem this is solving? Is this a
> problem for a non-glibc (e.g. alpine) build machine?
> 
> In terms of implementation this should be controlled by a menuconfig option.

Yes, particularly musl, but possibly others.

The error is as follows:
[INFO ]  Installing binutils for host
[EXTRA]    Configuring binutils
[EXTRA]    Building binutils
[ERROR]    /home/ismael/toolchain-test/.build/x86_64-unknown-linux-musl/src/binutils/gprofng/src/Experiment.cc:234:12: error: 'fopen64' was not declared in this scope; did you mean 'open64'?
[ERROR]    /home/ismael/toolchain-test/.build/x86_64-unknown-linux-musl/src/binutils/gprofng/src/Experiment.cc:244:9: error: 'fseeko64' was not declared in this scope; did you mean 'fseeko'?
[ERROR]    /home/ismael/toolchain-test/.build/x86_64-unknown-linux-musl/src/binutils/gprofng/src/Experiment.cc:285:16: error: 'ftello64' was not declared in this scope; did you mean 'ftello'?
[ERROR]    make[6]: *** [Makefile:844: Experiment.lo] Error 1
[ERROR]    make[5]: *** [Makefile:573: all] Error 2
[ERROR]    make[4]: *** [Makefile:472: all-recursive] Error 1
[ERROR]    make[3]: *** [Makefile:404: all] Error 2
[ERROR]    make[2]: *** [Makefile:7318: all-gprofng] Error 2
[ERROR]    make[1]: *** [Makefile:1028: all] Error 2
[

It's a compliance issue from binutils, it needs to define
_LARGEFILE64_SOURCE if it wants to use those identifiers.

I was just trying to make it build but maybe I should add a
disabled-by-default knob for gprofng?

Later I'll look into fixing binutils.

cpackham (changes requested):
As discussed should be wrapped up behind a config option, or maybe even dealt with with and upstream change.

PR #2403, opened 2025-06-26
m4: refine m4's ssp lib requirement for mingw32

Kerogi:
Do not link m4 with libssp if it is explicitly disabled in gcc

### DIFF
--- a/scripts/build/companion_tools/100-m4.sh
+++ b/scripts/build/companion_tools/100-m4.sh
@@ -57,9 +57,15 @@ do_m4_backend()
             ldflags="${ldflags} -lrt"
             ;;
         *-mingw32)
-            # m4 is built with stack smashing protection enabled which
-            # is not part of mingw-w64 c library in v7.0.0 and later.
-            libs="${libs} -lssp"
+            case "${CT_CC_GCC_LIBSSP}" in
+                "" | "n") ;;
+                *)
+                    # m4 is built with stack smashing protection enabled which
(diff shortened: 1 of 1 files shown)


scripts/build/companion_tools/100-m4.sh
```diff
@@ -57,9 +57,15 @@ do_m4_backend()
...
-            libs="${libs} -lssp"
+            case "${CT_CC_GCC_LIBSSP}" in
+                "" | "n") ;;
+                *)
+                    # m4 is built with stack smashing protection enabled which
+                    # is not part of mingw-w64 c library in v7.0.0 and later.
+                    libs="${libs} -lssp"
+                    ;;
+
+            esac
```
cpackham, line 68: I don't think this will quite work the way you need it to. `CT_CC_GCC_LIBSSP` affects the host configuration (i.e. the eventual target of your toolchain) but not the build. So this may do what you need for `do_companion_tools_m4_for_host()` but it may not do what you need for `do_companion_tools_m4_for_build()`.

I think we might see things start failing if you unset `CT_CC_GCC_LIBSSP` and build on a mingw32 machine with `CT_COMP_TOOLS_M4=y`. I don't have access to such a machine so I can't confirm it for myself.

Kerogi: Maybe it's worth introducing a separate option to control libs for m4 for both build/host?

PR #2410, opened 2025-07-14
mingw-w64: add v13.0.0

dbermond:
https://sourceforge.net/p/mingw-w64/mailman/message/59192665/

Notable changes:
 
  - Added import libraries for msvcr40d.dll, msvcrtd.dll, msvcr70d.dll, msvcr71d.dll, msvcr80d.dll, msvcr100d.dll, msvcr110d.dll.
  - Better CRT API consistency between UCRT and MSVCRT, making many APIs available to earlier versions of the MSVCR* runtime as well.
  - CRTDLL stat API fixes.
  - Updated Hyper-V headers.
  - Synchronized with Wine headers (from Wine 10.9).
  - Many other new win32 APIs.
  - Fix _atexit call differing between DLLs and EXEs.
  - Basic support for ARM64EC targets (arm64ec-w64-mingw32).
  - Make it possible to build winpthreads with MSVC and clang-cl.
  - Many small fixes for gendef, genidl and genpeimg.


cpackham:
It's helpful to include a bit of a blurb in the commit message about what the new version is adding. Even if it's just a link to the upstream release announcement.

dbermond:
Done. Thank you for the reply.

Please let me know if something else needs to be changed.

PR #2429, opened 2025-08-10
Packages update q2 2025

zagura:
* update config using ct-ng updatetools.
* bump linux kernel and packages.
* run ./maintainer/manage-packages.sh -D -d and unify chksum files
* mark non-LTS linux kernel 4.x, 5.x before 5.4 obsolete


packages/RELEASES
```diff
@@ -0,0 +1,24 @@
+automake 1.17: https://lists.gnu.org/archive/html/autotools-announce/2024-07/msg00000.html
```
cpackham, line 1: Did you intend to include this? I'm not actually opposed to including something that links to various upstream release announcements (I tend to include this in the commit message) but this list seems a bit ad-hoc and incomplete.

zagura: Not intended include - local notes for the source of releases.

QBos07:
Seems like my config.sub patch for SuperH is not applied yet: https://mail.gnu.org/archive/html/config-patches/2025-07/msg00021.html

cpackham:
I kicked the automake update out because it seemed to break `m68k-unknown-elf` and `x86_64-w64-mingw32,x86_64-pc-linux-gnu` when I ran a build-all locally. I did retry those manually with the automake update on top of the re-jigged series and they worked. Probably worth revisiting. It'd be good to get the automake update in if anyone has bandwidth to look at it. I'm starting to think about cutting a release soon.

PR #2144, opened 2024-04-19
binutils,gdb: fix enabling zstd compression

sobczyk:

PR #2466, opened 2025-12-19
scripts: fix zstd companion library usage

Lapshin:
Proper fix that was introduced in https://github.com/crosstool-ng/crosstool-ng/pull/2144

- added proper AR for zstd build to fix macos builds
- zstd was never installed properly because of the absence of a pkg-config file
- binutils option `--with-zstd=` does not process paths (see https://github.com/bminor/binutils-gdb/blob/200164d46732f6a2fb9d81aa9650c006db4148e6/config/zstd.m4)

### DIFF
--- a/packages/musl/package.desc
+++ b/packages/musl/package.desc
@@ -1,4 +1,4 @@
 repository='git git://git.musl-libc.org/musl'
-mirrors='https://www.musl-libc.org/releases'
+mirrors='https://www.musl-libc.org/releases https://sources.buildroot.net/musl/'
 archive_formats='.tar.gz'
 signature_format='packed/.asc'
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -33,6 +33,7 @@ do_binutils_for_build() {
     binutils_opts+=( "prefix=${CT_BUILDTOOLS_PREFIX_DIR}" )
     binutils_opts+=( "cflags=${CT_CFLAGS_FOR_BUILD}" )
     binutils_opts+=( "ldflags=${CT_LDFLAGS_FOR_BUILD}" )
+    binutils_opts+=( "complibs=${CT_BUILDTOOLS_PREFIX_DIR}" )
 
     do_binutils_backend "${binutils_opts[@]}"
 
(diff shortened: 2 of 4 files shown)


cpackham:
@Lapshin looks like there was a build failure related to this
[powerpc64-unknown-linux-musl.ubuntu-22.04.log.zip](https://github.com/user-attachments/files/24919958/powerpc64-unknown-linux-musl.ubuntu-22.04.log.zip)

cpackham:
Latest failure looks like a legit problem (powerpc-unknown-linux-gnu)

```
[ERROR]    configure: error: --with-zstd was given, but pkgconfig/libzstd.pc is not found
[ERROR]    checking whether the compiler supports the __inline keyword... make[2]: *** [Makefile:3055: configure-bfd] Error 1
```

cpackham:
I think I know what's going on. Long story short it's been broken since it was first introduced in either 711e7bc11 or ee5a4026c3. It was just broken in such a way that it turned zstd support off in binutils and gdb.

The gdb-native part is doubly broken because of the `}}` and because we don't have a recipe for building zstd for the target. I'll see if I can sort out building zstd for the target. Failing that I'll try and disable it for the gdb-native portion.

smarvonohr:
Somehow I'm having problems with this change. Depending on the location from where I build my toolchain it either works, or I get the same configure error mentioned above. Everything works when I build from `/workdir`, but when building from `/home/jenkins/toolchain` it does not find the libzstd.pc for some reason (I'm still degugging). Maybe I'm hitting some path/argument length limit?

cpackham:
Hmm is it possible there is another libzstd.pc somewhere? Could you attach
a build.log from both the good and bad case.

smarvonohr:
Just tested again, it looks like the difference is not the path, but there were some leftover files from previous tests with rc1 in the `/workdir`. I have now started with fresh clean folders and it fails now in both cases. See the attached build log and config.
[config.txt](https://github.com/user-attachments/files/30930731/config.txt)
[build.log.tar.gz](https://github.com/user-attachments/files/30930718/build.log.tar.gz)

smarvonohr:
Nevermind, I was just missing pkg-config in my container:
`configure: WARNING: pkg-config is not installed`

QBos07:
Related error https://github.com/QBos07/CP-dockercross/actions/runs/31534935270/job/93923701987

cpackham:
@QBos07 I think that's likely the fact that pkg-config is missing from the build environment. I'm tempted to drop the explicit `--with-zstd` and leave it as `auto`. That may make things a little non-deterministic. GCC looks as though it'll happily pick up zstd without pkg-config (it does seem to carry the zstd.m4 that uses pkg-config but seems to have other open-coded ways of finding libzstd). But binutils and gdb both require pkg-config.

QBos07:
I added `pkgconf` and it worked. The error message is not very good. Maybe we should add a build tool check for pkg-config?

PR #2469, opened 2025-12-29
libexpat comment it

lygstate:


cpackham:
Code change looks OK but the commit message is missing a sign-off and could be a bit more descriptive

QBos07:
That looks better

PR #2472, opened 2026-01-09
Add experimental support for Windows ARM64 PE/COFF generation

ahmedarif193:
This PR updates the toolchain to enable the generation of PE/COFF binaries for Windows ARM64:

- Successfully generated valid binaries for Windows ARM64.
- The current toolchain was able to compile ReactOS for ARM64 and successfully perform handoff to the kernel.
- While comprehensive smoke tests are still required, the PR in its current state provides functional experimental support for this target.
- This PR allow building crosstool-ng in Tahoe 26 too (X86 targets included)


ahmedarif193:
@QBos07 I did please review again

QBos07:
LGTM
@cpackham has final say

cpackham (commented):
I think this will fix a bug I was seeing when attempting to use the macos-15 runners.

PR #2483, opened 2026-01-31
Disable fail-fast for CI build

QBos07:
This makes other failures surface that would not be seen before due to cancelation because of another failure.


QBos07:
Depends, if there are multiple issue it needs multiple reruns to fix them all.

Problem is that we can't distinguish between a configuration specific failure or a broader problem.

cpackham:
I'm going to close this one. I kind of agree with the sentiment and if we had our own action runners I'd apply it but as we're using the free tier failing fast is the better option.

PR #2503, opened 2026-05-10
Fix aarch64-mingw builds with posix threading

BtbN:
This makes gcc aware of the posix threading model, as it is on x86.
Without this, a build with posix threads/winpthread will eventually fail with linker errors about missing pthread symbols.

### DIFF
--- a/packages/gcc/15.2.0/0027-aarch64-mingw-fix-support-for-posix-threading.patch
+++ b/packages/gcc/15.2.0/0027-aarch64-mingw-fix-support-for-posix-threading.patch
@@ -1,0 +1,65 @@
+commit 51b8f38d0859a1c1a068f4710b2876b6827c3634
+Author: Timo Rothenpieler <timo@rothenpieler.org>
+Date:   Fri May 15 21:55:38 2026 +0200
+
+    aarch64: mingw: fix support for posix threading
+
+    Currently without this --with-threads=pthread fails.
+
+    gcc/ChangeLog:
+
+            * config/i386/mingw-pthread.h:
+            rename to generic config/mingw/mingw-pthread.h
+            * config.gcc [aarch64-*-mingw*]:
+            Fix support for posix threading on aarch64 mingw targets.
+
+    Signed-off-by: Timo Rothenpieler <timo@rothenpieler.org>
+
+diff --git a/gcc/config.gcc b/gcc/config.gcc
+index 087aaa7be37..eb8f48f4443 100644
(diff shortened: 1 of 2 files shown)


packages/gcc/15.2.0/0027-Fix-pthread-library-linking-on-aarch64-mingw-builds.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git a/gcc/config.gcc b/gcc/config.gcc
+index 8fe99616f82..a9d2da5994c 100644
+--- a/gcc/config.gcc
++++ b/gcc/config.gcc
+@@ -1325,6 +1325,9 @@ aarch64-*-mingw*)
+ 	tm_file="${tm_file} aarch64/aarch64-abi-ms.h"
+ 	tm_file="${tm_file} aarch64/aarch64-coff.h"
+ 	tm_file="${tm_file} aarch64/cygming.h"
++	if test x$enable_threads = xposix ; then
++		tm_file="${tm_file} i386/mingw-pthread.h"
```
cpackham: I'd like to hear from upstream but this doesn't seem right to me. We're in the middle of an aarch64 block and all of a sudden we're pulling in an i386 file?

cpackham (changes requested):

packages/gcc/15.2.0/0027-Fix-pthread-library-linking-on-aarch64-mingw-builds.patch
BtbN: All it does is set a singular flag that's consumed by the mingw32 header directly following it, to switch the threading model to posix.
For a bigger patch the file should probably just be moved to a generic location.
Though looking at the woarm64 gcc repo, they just have a load of i386 stuff in their aarch64-mingw section as well:
https://github.com/Windows-on-ARM-Experiments/gcc-woarm64/blob/woarm64/libgcc/config.host#L467

cpackham:
I think ultimately someone just needs to send an upstream patch to move the
header. I'm a lot more comfortable carrying backports of upstream changes.

BtbN:
I've sent the same patch upstream, but got zero reaction so far.
The thing is, without this patch, a build with CT_THREADS=posix is flat out broken due to numerous linker errors not finding pthread symbols.
While it works perfectly fine with this.

BtbN:
Patch has now been merged upstream.

cpackham (approved):

PR #2504, opened 2026-05-11
Fix i686-w64-mingw32 build. Disable cdtor decloning by default for PE/COFF

nikkon-dev:


cpackham:
Have you sent this upstream?

nikkon-dev:
> Have you sent this upstream?

I have not. I'm not sure this is how the issue should be fixed upstream. Maybe improving libstdc++ to properly use/expose all ctors/dtors would be a better approach than disabling dedup in -Os, which would increase the binary size.

So far, this oneliner unblocks the i686-w64-mingw32 target if anyone wants it (I'm not sure how relevant 32bit builds are in the modern world).

QBos07:
I would create a libstdcxx issue and link to here and upload the gcc patch aswell. If the tell you to fuck off then go gcc.

cpackham:
@nikkon-dev did you get anywhere with reporting this as an upstream issue?

PR #2516, opened 2026-07-09
debug: Add option to keep only the last-saved step's state

### LINKED ISSUE #2436: Keep only last-executed step's state
We currently have `CT_DEBUG_CT_SAVE_STEPS` available to us in order to save progress and thus potentially restore progress from different points.
I have a slightly different use-case for this -- building toolchains on publicly-hosted GitHub actions: [GHA runs](https://github.com/tzarc/qmk_toolchains/actions/workflows/create_qmk_toolchains.yaml)

I'd like to request an additional option -- only saving the final step so it can be resumed from the end. Intermediate steps should be removed.

I realise there's currently CI on crosstool-ng to build toolchains however the sample configurations seem to be completing far quicker than the ones we build for [QMK](https://github.com/qmk/qmk_firmware) -- yours all complete in <3hrs or so, whereas the bare-metal ARM builds would regularly exceed the 6hr termination for GHA jobs.

As a workaround, I've had to split up our builds into 4 separate ct-ng runs leveraging the saving of steps -- inclusive of things like clearing up disk space on the runners as the amount of space necessary for all intermediate steps when building/saving is prohibitive.

Would it be possible to get another option for ct-ng to perform removal of state during execution for steps apart from the last-completed one?

tzarc:
Add `CT_DEBUG_CT_SAVE_STEPS_LAST_ONLY`, which prunes earlier saved states as the build progresses, keeping only the most recently saved step. This bounds the disk space used by saved states to roughly one snapshot, which is useful on space-constrained CI runners (e.g. GitHub Actions).

Pruning happens after the new state is fully written, so an interrupted save never leaves the build without a valid state to restart from. The build can still be restarted, but only from the last-saved step.

Closes #2436.

Successful build on GitHub Actions can be found here: https://github.com/tzarc/qmk_toolchains/actions/runs/28990567525

### DIFF
--- a/config/global/ct-behave.in
+++ b/config/global/ct-behave.in
@@ -149,6 +149,18 @@ config DEBUG_CT_SAVE_STEPS_COMPRESS
     default "zstd"  if DEBUG_CT_SAVE_STEPS_COMPRESS_ZSTD
     default ""      if DEBUG_CT_SAVE_STEPS_COMPRESS_NONE
 
+config DEBUG_CT_SAVE_STEPS_LAST_ONLY
+    bool
+    prompt "Keep only the last-saved step's state"
+    depends on DEBUG_CT_SAVE_STEPS
+    help
+      Remove earlier saved states as the build progresses, keeping only
+      the most recently saved step. This bounds the disk space used by
+      saved states to roughly one snapshot, which is useful on space-
+      constrained CI runners.
+
+      You can still restart the build, but only from the last-saved step.
+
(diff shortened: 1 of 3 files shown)


tzarc:
Seemingly unrelated CI failure, unable to retrieve the newlib-nano tarball.

QBos07:
wouldnt a rebase be better?

tzarc:
Done.

QBos07 (approved):

PR #2524, opened 2026-07-29
binutils, gdb: fix stray brace that disables zstd support

mheimlich:
The zstd guard is written with a stray `}` in two build scripts:

```sh
if [ "${CT_COMP_LIBS_ZSTD}}" = "y" ]; then
```

`"${CT_COMP_LIBS_ZSTD}}"` expands to e.g. `y}`, so the test is never true and the tool is always configured `--without-zstd`, even when `CT_COMP_LIBS_ZSTD=y` builds the zstd companion library. As a result the affected tools are built without zstd support (no `--compress-debug-sections=zstd`, and no zstd-compressed LTO objects). Removing the extra brace restores the intended `--with-zstd=${complibs}`.

Fixed in both places:

- `scripts/build/binutils/binutils.sh` (binutils `ld`/`as`)
- `scripts/build/debug/300-gdb.sh` (native `gdb`)

### DIFF
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -181,7 +181,7 @@ do_binutils_backend() {
 
     [ "${CT_TOOLCHAIN_ENABLE_NLS}" != "y" ] && extra_config+=("--disable-nls")
 
-    if [ "${CT_COMP_LIBS_ZSTD}}" = "y" ]; then
+    if [ "${CT_COMP_LIBS_ZSTD}" = "y" ]; then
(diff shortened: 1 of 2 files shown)


mheimlich:
Duplicate of #2466, which fixes the same `${CT_COMP_LIBS_ZSTD}}` typo in both `scripts/build/binutils/binutils.sh` and `scripts/build/debug/300-gdb.sh` (and additionally corrects the zstd companion library usage in `070-zstd.sh`). Closing in favor of that PR — thanks @Lapshin!

PR #2529, opened 2026-08-14
binutils, gdb: Add tristate option for zstd support

cpackham:
A plain --with-zstd translates to --with-zstd=yes which causes configure to fail if libzstd is not present (or if pkg-config is missing). Using --with-zstd=auto means it will try to find libzstd but if it is not found the build will continue without zstd support. It should be noted the GCC uses a slightly different method for locating libzstd so it is possible for GCC to be built with zstd support and binutils/gdb to be built without.


QBos07:
I am not really a fan of doing this implicitly and having inconsistencies between gcc and binutils.

Other options could be an explicit pkg-config check or manipulation of internal autoconf variables.

cpackham:
Hmm it's a little tough to unpick. I can add
```
config ZSTD_has_pkg_config
    def_bool $(success,which pkg-config)
```
but then un-picking how zstd is selected by gcc is a bit tricky.

I think the discrepancy is OK. For GCC zstd is using for compressing lto information. For binutils/gdb it's used for compressing debug sections.

The alternative is do to nothing and tell people to install pkg-config when things fail

QBos07:
There exists ZSTD_CFLAGS and ZSTD_LIBS as overrides to pkgconfig: https://github.com/gnutools/binutils-gdb/blob/master/bfd/configure#L15426

cpackham:
I'm erring on the side of doing nothing. For most systems pkg-config will be installed it's pretty ubiquitous. Our binutils and gdb recipes should probably grow tristate options for zstd support the same way gcc does for LTO.

PR #2533, opened 2026-08-23
Set PKG_CONFIG_PATH for binutils configure-gas

tttapa:
Set PKG_CONFIG_PATH for binutils configure-gas.  
Otherwise, "Building binutils" fails with `configure: error: --with-zstd was given, but pkgconfig/libzstd.pc is not found`.

### DIFF
--- a/scripts/build/binutils/binutils.sh
+++ b/scripts/build/binutils/binutils.sh
@@ -237,7 +237,9 @@ do_binutils_backend() {
                 ;;
         esac
     fi
-    CT_DoExecLog ALL make "${extra_make_flags[@]}" ${CT_JOBSFLAGS}
+    CT_DoExecLog ALL                                            \
+    PKG_CONFIG_PATH="${complibs}/lib/pkgconfig"                 \
+    make "${extra_make_flags[@]}" ${CT_JOBSFLAGS}
 
     CT_DoLog EXTRA "Installing binutils"
     CT_DoExecLog ALL make install


cpackham:
@tttapa do you have an example of a config that triggers a failure? I know we have had a series of issues related to this but I'm kind of surprised gas is affected (the zstd support is only used by ld as far as I'm aware).

tttapa:
<details>

<summary>Here's the defconfig I used (with 2e5d0b813867c4bca820592ff01775aba1367e0f)</summary>

```
CT_CONFIG_VERSION="4"
CT_OBSOLETE=y
# CT_LOG_PROGRESS_BAR is not set
CT_ARCH_X86=y
CT_ARCH_64=y
CT_TOOLCHAIN_PKGVERSION="tttapa/toolchains"
CT_TOOLCHAIN_BUGURL="https://github.com/tttapa/toolchains"
CT_TARGET_VENDOR="centos7"
CT_KERNEL_LINUX=y
CT_LINUX_V_3_10=y
# CT_KERNEL_LINUX_INSTALL_CHECK is not set
CT_BINUTILS_V_2_47=y
CT_BINUTILS_PLUGINS=y
CT_GLIBC_V_2_17=y
CT_GLIBC_EXTRA_CFLAGS="-Wno-implicit-int"
CT_CC_GCC_EXTRA_CONFIG_ARRAY="--enable-multiarch --enable-libstdcxx-backtrace=yes --with-pic"
# CT_CC_GCC_ENABLE_DEFAULT_PIE is not set
CT_CC_GCC_LIBGOMP=y
CT_CC_GCC_LIBQUADMATH=y
CT_CC_GCC_LIBSANITIZER=y
CT_CC_GCC_BUILD_ID=y
CT_CC_LANG_CXX=y
CT_CC_LANG_FORTRAN=y
CT_DEBUG_GDB=y
# CT_GDB_CROSS is not set
CT_GDB_NATIVE_BUILD_IPA_LIB=y
CT_GDB_NATIVE_STATIC_LIBSTDCXX=y
CT_GMP_EXTRA_CFLAGS="-std=gnu11"
CT_NCURSES_EXTRA_CFLAGS="-std=gnu11"
```

</details>

It fails with:

```
[ALL  ]    checking for x86_64-build_pc-linux-gnu-pkg-config... no
[ALL  ]    checking for pkg-config... no
[ALL  ]    checking for libzstd >= 1.4.0... no
[ERROR]    configure: error: --with-zstd was given, but pkgconfig/libzstd.pc is not found
[ALL  ]    yes
[ERROR]    checking for ffs... make[2]: *** [Makefile:5474: configure-gas] Error 1
[ERROR]    make[2]: *** Waiting for unfinished jobs....
```

But this is a parallel build, and there are other errors further down, so the output may be a bit misleading about where exactly the error occurred. (Sorry, I didn't find the time yet to further narrow it down. I just noticed that the `make` step also calls configure, which fails if the pkg-config path is not provided.)

Here's the full log: [build.log.gz](https://github.com/user-attachments/files/31372634/build.log.gz)

QBos07:
do you have pkg-config installed? the error message is a bit misleading here. it should be more detailed in the accompaning config.log

tttapa:
Thanks, I believe you're right: I wrongly concluded that the change above fixed the issue, but it was installing pkg-config while debugging that did it 🤦‍♂️

